PR #138, opened 2012-09-05
Added support for recording and displaying new/repeat episodes

sofakng:
This PR adds support for recording "New Episodes" versus "Repeat Episodes".

The EPG has a new column ("Repeat") which displays if a program is a New Episode or a Repeat.  (i.e. was this program ever aired or is it the first time it's broadcast?)  Similarly, there is a new filter to show only new or all episodes.

AutoRec has been modified to also allow recording of only new episodes (or all episodes).

(This PR was created because I didn't like recording every episode of a particular program and then later going back and deleting the repeat/rerun episodes manually.  This also saves you time because you don't need to know what timeslot that new airings are broadcast since you just create rules such as:  "Record MythBusters - New Episodes Only")

NOTE:  This has only been tested with xmltv EPG source.  It shouldn't break other EPG sources but I'm not sure if this supply the neccessary is_repeat flag.

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -74,6 +74,10 @@ typedef enum {
   DVR_PRIO_UNIMPORTANT,
 } dvr_prio_t;
 
+typedef enum {
+  DVR_REPEATS_ALLEPISODES,
+  DVR_REPEATS_NEWONLY
+} dvr_repeats_t;
 
 LIST_HEAD(dvr_rec_stream_list, dvr_rec_stream);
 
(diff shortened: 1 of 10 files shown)


yaqwsx:
Are there any updates on this capability?

perexg:
The new 4.1 code has 'Broadcast type' field in autorec which does exactly this thing. If episode is new / repeated - it should be visible when you click on the event details in EPG grid, but I agree that this might be better to show it directly in the grid using an icon...

ljalves:
Shouldn't this PR be closed?
I believe it no longer makes sense since the implementation of the "Related broadcasts" and "Alternative showings" buttons?

PR #218, opened 2013-01-09
Episode number extraction from title/description

skazi0:
Most of polish channels don't have episode numbers in dvb metadata. Proposed workaround extracts episode numbers based on basic keyword prefixes. When broadcast title or description contains configured keyword, following sequence of digits is used as episode number.
autoepnum2
The keywords are configured per channel.
autoepnum
It seems that at least for polish channels, one keyword covers most of shows which have any indication of episode number (i.e. channels use consistent format for episode-in-description).

Possible further extension: use regexps to make the extraction more flexible and/or allow season/episode number extraction where available.

### DIFF
--- a/src/epggrab/module/eit.c
+++ b/src/epggrab/module/eit.c
@@ -494,20 +494,21 @@ static int _eit_desc_crid
  * Parse description and extract episode number.
  */
 static int _eit_extract_enum
-  ( epg_episode_t *episode, lang_str_t *text, epggrab_module_t *mod )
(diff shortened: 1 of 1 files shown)


skazi0:
@adamsutton was there any movement around what you described in your comment (https://github.com/tvheadend/tvheadend/pull/218#issuecomment-12064448)? Is there any script-based EPG processing solution available in tvheadend? If not, I might want to get back to this subject.
If possible please add any links to old proposed solutions/PRs here for reference.

piotrasd:
+1

PR #274, opened 2013-05-12
Add transcoding via filter scripts to http streaming

BtbN:
This adds a new parameter to the http streaming server.
Adding ?transcoder=test makes it pipe all http output through a "filter", which can do basicaly anything with it.

It searches for the trascoders in hts_settings_get_root()/transcoders
It also passess all GET parameters in hc_req_args as environment variables, prefixed with GET_, so the transcoder can get options via url parameters.

This also modifies the writing to the socket a bit, so it checks if it can write to the http stream, because it can no longer rely on write() errors, because it no longer directly writes to the tcp socket. I use select and poll for this.

There is still a problem somewhere, because in my tests a transcoding vlc throws a lot of warnings about broken input data. Connecting with vlc via http, which should bring the same data to vlc, works perfectly. But this only seems to happen with higher bandwith streams, so i guess it's some buffer overflowing.

But in general, this already works fine.

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -227,7 +227,7 @@ page_static_file(http_connection_t *hc, const char *remain, void *opaque)
 }
 
 static pid_t
-http_stream_transcode(http_connection_t *hc, muxer_t *mux)
+http_stream_transcode(http_connection_t *hc, muxer_t *mux, char *mime, int *mime_size)
 {
   pid_t pid = 0;
   int fd_pipe[2];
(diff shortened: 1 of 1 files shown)


Hedda:
You might want to consider borrowing/stealing the UPnP / DNLA client profiles XML files from the latest Plex Media Server and use as a base to create custom MIME types for transcoding and/or remuxing, see:

https://github.com/Hedda/Plex-DNLA-client-profiles/tree/master

https://github.com/Hedda/Plex-DNLA-client-profiles/archive/master.zip

PS Hope it is OK to have copies of these XML files on GitHub as Plex.tv don't have an source code repository available because rest of the software is closed source.

perexg:
This functionality was implemented like MPEG-TS Spawn profile where users can configure the command which should be invoked for the original MPEG-TS stream.  v4.3-344-gc349525b2, more commits, the main is 5bbddc9e0143b5135be42714beb3201d9c2e0019 .

PR #369, opened 2014-04-08
Add SAT>IP support (remote network tuners)

perexg:
This code adds initial support for remote SAT>IP tuners (servers). I touched slightly other parts, too.

### DIFF
--- a/src/descrambler/capmt.c
+++ b/src/descrambler/capmt.c
@@ -1074,6 +1074,8 @@ capmt_service_start(service_t *s)
         tvhlog(LOG_DEBUG, "capmt",
           "New caid 0x%04X for service \"%s\"", c->caid, t->s_dvb_svcname);
 
+        mpegts_table_register_caid(((mpegts_service_t *)s)->s_dvb_mux, c->caid);
+
         /* add it to list */
         cce             = calloc(1, sizeof(capmt_caid_ecm_t));
         cce->cce_caid   = c->caid;
--- a/src/descrambler/cwc.c
+++ b/src/descrambler/cwc.c
@@ -2031,6 +2031,8 @@ cwc_service_start(service_t *t)
 
     if (ct) continue;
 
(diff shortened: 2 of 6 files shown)


perexg:
Little note for gssbox (and compatible boxes - IDL4K firmware) owners:

There is a new workaround in the latest tvh code which should fix ocassional RTSP 404 errors (bad tunes). I was able to hit this error when I set idle scan for 3 tuners. Basically, the firmware sometimes closed the RTSP connection with 404 error code after new SETUP because previous session was alive (even after TEARDOWN), so another - extra TEARDOWN command must be send to device with previous session and stream IDs to close the old (obsolete) streaming.

clandmeter:
@perexg there seems to be a newer firmware according to http://www.vdr-portal.de/board19-verschiedenes/board52-andere-hardware/p1227048-howto-idl4k-basierte-boxen-modifizieren-digibit-r1-dsi-400-idl400s/#post1227048
Have you tried it?

perexg:
@clandmeter : According this https://github.com/jollyjinx/idl4k/commit/6049416df7150308e59a320a5e8817519e4091e8 (yes, some parts of firmware are now available in sources), only some multicast streaming issues were fixed in V1.17.0.120 . But I'll try.

clandmeter:
I think i also had issues with my network, and i see a patch for the realtek nic. I will also give it a try.

perexg:
The 404 error is also in V1.17.0.120:

<pre>
[   INFO] subscription: 0141: "scan" subscribing to mux, weight: 1, adapter: "SAT>IP DVB-S Tuner #1 (192.168.100.236)", network: "DVB-S 1W", mux: "11843V"
[WARNING] satip: SAT>IP DVB-S Tuner #1 (192.168.100.236) - RTSP 404 ERROR (retrying)
[ INFO] satip: extra shutdown done for session D51C71EE
</pre>

PR #388, opened 2014-05-11
[Power] Implemented power-up/down functionality

schreibubi:
Implement all the required functionalities to shut-down the tvheadend server when it is idle and restart the tvheadend server before an upcoming recording.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,10 +117,12 @@ SRCS =  src/version.c \
 	src/rtsp.c \
 	src/fsmonitor.c \
 	src/cron.c \
+	src/power.c \
 	src/esfilter.c
 
 SRCS-${CONFIG_UPNP} += \
 	src/upnp.c
+
 
 SRCS += \
 	src/api.c \
@@ -274,7 +276,7 @@ endif
 # File bundles
 SRCS-${CONFIG_BUNDLE}     += bundle.c
 BUNDLES-yes               += docs/html docs/docresources src/webui/static
-BUNDLES-yes               += data/conf
+BUNDLES-yes               += data/conf bin
 BUNDLES-${CONFIG_DVBSCAN} += data/dvb-scan
 BUNDLES                    = $(BUNDLES-yes)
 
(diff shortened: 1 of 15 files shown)


fda77:
@perexg: Thanks! Did not noticed this before, im a beginner.
Btw, addin value to post-record script would be okay for my usage, because I want to hibernate and so the value does not change

fda77:
Uh, the script seems not to be for users. First some modules and tools where missing. Then some where in another place in my system (eg /usr/bin/dbus-launch) and then
```X11 initialization failed.```
.. but i've no X server installed :((

Hedda:
This might be dead now, but FYI, the terminology this feature is called "real-time clock alarm" when talking about the standard computer BIOS/UEFI functionality which is built-into most computers based on ACPI BIOS or UEFI and have a RTC (Real-Time Clock).

https://en.wikipedia.org/wiki/Real-time_clock_alarm

The feature is also often refered to as WoRTC (Wake on RTC) or WOT (Wake-On-Time), the latter abbreviation  WOT I believe was chosen because it follow a similar name-standard to WOL (Wake-On-LAN).

Btw, I think that this is a very nice DVR set-top box appliance-like function to have for a PVR software like Tvheadend, and it is also a feature which is available in MythTV, see: http://www.mythtv.org/wiki/ACPI_Wakeup

git-developer:
Just for the record and probably years too late:
> 
> 
> While I don't argue this can be useful, I may suggest that unless tvheadend is the only function that needs to wakeup/sleep the device this is not the right place to take this sleep/restore decision: it should be a centralized and tvheadend being only a client. If other software programs the RTC (or even if something is programed directly in the BIOS) it may well screw up the other wakeup

This idea was taken into account for the development of [git-developer/autosuspend](https://github.com/git-developer/autosuspend) and [languitar/autosuspend](https://github.com/languitar/autosuspend).

Flole998:
With #1845 it is now possible to use external software to get information about the next tvheadend event and when to wake up the system again. Adding the suspend/wake functionality to Tvheadend would be out of scope.

PR #526, opened 2014-10-28
tvhdhomerun: only run discovery once upon startup (solves #2418)

beralt:
Bug #2418 revealed that the current discovery code sends datagram broadcasts every 30 seconds. This behavior is pretty unwanted and unnecessary. This PR modifies the discovery code such that it is only run once when starting tvheadend.

An additional feature that would complete this is a discovery button in the web GUI. That is a bit beyond my current grasp of the internals of tvheadend though; maybe someone could help me with that part.

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
@@ -27,8 +27,7 @@
 #include <arpa/inet.h>
 #include <openssl/sha.h>
 
-static void tvhdhomerun_device_discovery_start( void );
-
+static void tvhdhomerun_device_discovery( void );
 
 static void
 tvhdhomerun_device_class_save ( idnode_t *in )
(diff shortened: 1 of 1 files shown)


pakar:
Just made a new build and noticed some issues and then and found this commit..

There is actually some thought behind why it behaved that way..
- Support hotplug of new devices. (but maybe better to have a manual scan for that)
- Some networkcards/switches may cause the network to take some time before becoming available. With STP it can take some time before the network becomes available.
- The discovery sometimes fails with some HDHomerun devices. I have one that fails randomly (maybe once every 15-20'th time) on a completely idle network. No dropped packages when it's streaming and a network dump shows that it just don't accept probe-requests from time to time (drops them or just don't  send replies)..... My most recent device do not display this behavior.

But i agree that it was a bit braindead (of me :) that i left it in that state without the possibility to disable the discover.

But anyhow, just leaving some comments regarding this so you know why it was there to start with since i will probably not have any time to do work on that myself. Reverted the commit locally so it's working for me..

PR #581, opened 2015-01-07
Issue 1625 - Option for Windows-compatible filenames (was: Filename: Umlaute and special characters)

dgolda:
See: https://tvheadend.org/issues/1625
When filename created by DVR contains both national characters (umlate or ąęłóśź) and special characters (like colon : ), file created by DVR is not accesible from Windows by Samba sharing.
It is possible to remove ALL unsafe characters, but it removes both special characters (it's OK) and national characters (it's not OK).

This pull request adds new option in DVR configuration, called "Replace Samba Unsafe Characters with '_'".
And when this option is checked, replaces unsafe characters / : \ < > | * ? ' " to _

### DIFF
--- a/src/dvr/dvr_config.c
+++ b/src/dvr/dvr_config.c
@@ -727,6 +727,13 @@ const idclass_t dvr_config_class = {
       .off      = offsetof(dvr_config_t, dvr_whitespace_in_title),
       .group    = 5,
     },
+    {
+      .type     = PT_BOOL,
+      .id       = "clean-samba-unsafe",
+      .name     = "Replace Samba Unsafe Characters with '_'",
+      .off      = offsetof(dvr_config_t, dvr_clean_samba_unsafe),
+      .group    = 5,
+    },
     {}
   },
 };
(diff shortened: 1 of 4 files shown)


src/dvr/dvr_config.c
```diff
@@ -727,6 +727,13 @@ const idclass_t dvr_config_class = {
       .off      = offsetof(dvr_config_t, dvr_whitespace_in_title),
       .group    = 5,
     },
+    {
+      .type     = PT_BOOL,
+      .id       = "clean-samba-unsafe",
+      .name     = "Replace Samba Unsafe Characters with '_'",
```
Jalle19: I would call this "Use Windows-compatible filenames". Normal people don't necessarily know what Samba means.

dgolda: Could be.

dgolda:
I've changed to caption and variables to suggested "Windows-compatible filenames" and also added description to help.

perexg:
Merged. Thanks.

PR #582, opened 2015-01-08
accept "%c" placeholder for channel icons to use unmangled channel

Jalle19:
I have all my channel icons on a web server named exactly like the channel, e.g. "Yle TV1 HD.png". With this patch I don't have to do any renaming when mapping channels.

@perexg does tvheadend have a helper for URL-encoding strings? I'm not completely sure that the current approach works since the URL isn't escaped (spaces should be converted to %20 etc.).

### DIFF
--- a/docs/html/config_misc.html
+++ b/docs/html/config_misc.html
@@ -83,9 +83,13 @@
     <dd>If both a picon and a channel-specific (e.g. channelname.jpg) icon are defined, use the picon.</dd>
  
     <dt>Channel icon path</dt>
-    <dd>Path to an icon for this channel. This can be named however you wish, as a local (file://) or remote (http://) image.<br>
-	Example: file:///tmp/icons/%C.png (where %C is the channel nameon local storage where
-	is TVHeadend) or http://example.com/example.png to set icon from web page.</dd>
+    <dd>Path to an icon for this channel. This can be named however you wish, as a local (file://) or remote (http://) image. 
+        The following placeholders are available:<br>
+	<ul>
+		<li>%C - the transliterated channel name in ASCII (safe characters, no spaces etc.)</li>
+		<li>%c - the channel name (URL encoded ASCII)</li>
+	</ul>
+	Example: file:///tmp/icons/%C.png or http://example.com/%c.png</dd>
   
     <dt>Picon path</dt>
     <dd>Path to a directory (folder) containing your picon collection. This can be named however 
--- a/src/channels.c
+++ b/src/channels.c
@@ -645,17 +645,14 @@ channel_get_icon ( channel_t *ch )
         (chname = channel_get_name(ch)) != NULL && chname[0]) {
       const char *chi, *send, *sname, *s;
       chi = strdup(chicon);
-      send = strstr(chi, "%C");
-      if (send == NULL) {
-        buf[0] = '\0';
-        sname = "";
-      } else {
-        *(char *)send = '\0';
-        send += 2;
+
+      /* Check for and replace placeholders */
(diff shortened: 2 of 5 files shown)


Jalle19:
@perexg I added proper (hopefully) URL encoding

docs/html/config_misc.html
```diff
@@ -83,9 +83,13 @@
...
  
     <dt>Channel icon path</dt>
-    <dd>Path to an icon for this channel. This can be named however you wish, as a local (file://) or remote (http://) image.<br>
-	Example: file:///tmp/icons/%C.png (where %C is the channel nameon local storage where
-	is TVHeadend) or http://example.com/example.png to set icon from web page.</dd>
+    <dd>Path to an icon for this channel. This can be named however you wish, as a local (file://) or remote (http://) image. 
+        The following placeholders are available:<br>
+	<ul>
+		<li>%C - the channel name with safe characters (no spaces etc.)</li>
+		<li>%c - the channel name (untouched)</li>
```
perexg: Mention utf8 encoding here ?

Jalle19: What about it?

Jalle19:
@perexg is this what you meant?

perexg:
I meant to explicitly tell the user that the encoded string is using the utf8 encoding (which internally tvheadend uses for all international/language strings).

ProfYaffle:
and while it's being debated, I'm not sure that mangled/unmangled carries the right meaning. 'Translated' or 'converted' would be better.

Jalle19:
@perexg in that case I read the code wrong, are you saying the resulting string is in UTF-8?

@ProfYaffle sure, there may be a better word. I believe "transliterated" is the correct one, but no one understands what that means.

ProfYaffle:
Well, 'transliteration' would certainly do, and makes sense to me - but yes, it'd probably be lost on most people... nearly as many as 'unmangled' would perhaps be.

And I'm not sure UTF-8 is the issue here, as space == ASCII 32 == UTF-8 0x20. I think the UTF-8 bit as it applies to non-ASCII characters (e.g. ö, ŵ) is necessary, but this is also about translating reserved or 'unsafe' characters, as per the URI specifications:

http://www.ietf.org/rfc/rfc1738.txt

Just trying to help, though, it's your call.

perexg:
The chname string is utf8 string. You only encode it in %c condition (for HTTP URL), right ? I meant only to tell the user in the help, that this URL encoded channel name willl use utf8 for international characters, because it's not obvious.

Jalle19:
@ProfYaffle updated the help text a bit, mind taking a look?
@perexg I changed it so that `%c` will also receive UTF-8 to ASCII conversion before the URL encoding takes place. This seems safer. The main idea behind this anyway is to preserve spaces.

src/channels.c
```diff
@@ -665,6 +662,22 @@ channel_get_icon ( channel_t *ch )
           s++;
         }
       }
+      else if((send = strstr(chi, "%c"))) {
+        char *aname = intlconv_utf8safestr(intlconv_charset_id("ASCII", 1, 1),
+                                     chname, strlen(chname) * 2);
+        sname = url_encode(aname);
```
perexg: Missing NULL check for aname. The NULL can be returned if iconv() function fails for a reason.

ProfYaffle:
Go for it - we can always alter it later if folks really don't understand/disagree (e.g. 'no spaces' <> 'ASCII' for obvious reasons). But it's fine.

perexg:
Merged. Thanks.

PR #584, opened 2015-01-12
hdhomerun: Upgrade libhdhomerun

tajson:
SiliconDust has pushed a new version of libhdhomerun and has therefore deleted the old version from their download servers, thus breaking TVHeadend build scripts.

Only tested in the sense that TVHeadend builds with this updated version.


perexg:
Merged. Thanks.

PR #586, opened 2015-01-14
Update timeshift_filemgr.c

jdierkse:
It seems that in currently the timeshift_unlimited_period boolean is completely ignored when it comes to taking action on removing files or reporting a full buffer. This small patch should fix that, so that when the timeshift_unlimited_period boolean is set to true, tvheadend no longer evaluates the period of the current timeshift.


perexg:
Merged. Thanks.

PR #588, opened 2015-01-23
hdhomerun: autodetect ATSC network type

laurimyllari:
Move hdhomerun device model query a little earlier and detect ATSC tuners from _atsc at the end of the model name if the device has no configuration yet.


perexg:
Merged. Thanks.

PR #589, opened 2015-01-24
Fix 519ec96

manio:
When there was a further subscribing requests, the queued greeting could
be queued not as the first position in the queue, leading to problems
when subscribing the channel.
Now requesting to flush the queue in capmt_notify_server() to be sure
that the greeting go out to OSCam as a first command.


perexg:
Merged. Thanks.

PR #590, opened 2015-01-24
build: Add -ldl for arch armhf

tajson:
Fixes building for Odroid-C1 on Debian Wheezy and Ubuntu Trusty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,8 @@ CFLAGS  += -Wno-unused-value -Wno-tautological-constant-out-of-range-compare
 CFLAGS  += -Wno-parentheses-equality -Wno-incompatible-pointer-types
 endif
 
+ARCH = $(shell $(CC) -dumpmachine | cut -d '-' -f 3)
+
 ifeq ($(CONFIG_LIBFFMPEG_STATIC),yes)
 CFLAGS  += -I${ROOTDIR}/libav_static/build/ffmpeg/include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -55,6 +55,8 @@ CFLAGS  += -Wno-unused-value -Wno-tautological-constant-out-of-range-compare
 CFLAGS  += -Wno-parentheses-equality -Wno-incompatible-pointer-types
 endif
 
+ARCH = $(shell $(CC) -dumpmachine | cut -d '-' -f 3)
```
perexg: ARCH is already defined in .config.mk ..

Makefile
```diff
@@ -67,6 +69,10 @@ ifeq ($(CONFIG_HDHOMERUN_STATIC),yes)
 CFLAGS  += -I${ROOTDIR}/libhdhomerun_static
 LDFLAGS += -L${ROOTDIR}/libhdhomerun_static/libhdhomerun \
            -Wl,-Bstatic -lhdhomerun -Wl,-Bdynamic
+
+ifneq (,$(filter gnueabihf,$(ARCH)))
+LDFLAGS += -ldl
```
perexg: The -ldl should be added on line 39.. This seems to be a duplicate.

tajson: Ok, reject this PL. It still won't compile if I don't add -ldl (complains about libx264).

PR #591, opened 2015-01-25
DVB-S2/T2 multistream support.

crazycat69:


crazycat69:
DVB-S2 Multiple Input Stream support.
DVB-T2 MultiPLP support.

perexg:
If PLS MODE and PLS CODE is dvb-s2 specific, it should be moved to the dvb_qpsk_config_t structure..

perexg:
Anyway, thanks for this patch. It looks good.

crazycat69:
But Pilot and Rolloff also defined as common fields of dvb_mux_conf struct ?

perexg:
Yes, hoping that these values can be reused for other modulations / signalling in future. You may keep them "global", if you think that's the right place. I don't have a strong objection for it.

perexg:
Merged. Thanks.

PR #592, opened 2015-01-28
tvhdhomerun: add modulation and symbolrate to channel tuning

beralt:
The current HDHomerun implementation uses "auto" during tuning. This requires users to set the tuning hints in the device configuration using the command-line (or GUI) provided by Sillicondust.

Users may experience issues with this approach when their network uses more than three different modulation strategies, since the device only allows for three hints.

This PR attempts to fix this for at least DVB-C networks, and only for 64 and 256 constellations. It does a fall-back to the auto behavior in other cases. I guess this should check against the supported demodulation types using the device feature call, but for now this works.

Note that in the Netherlands the two largest DVB-C providers are merging their networks causing a lot of issues with the old "auto" implementation.


perexg:
Merged.

PR #599, opened 2015-02-16
transcoding: use input time base for output

stbenz:
This fixes bitrate calculation when transcoding channels, which are not 25 fps. I have some channels which are 50 fps progressive and the 25 fps time base results in double the specified bitrate, as the time base here is used only for bitrate calculation in the codec, but the output video is still 50 fps.


perexg:
Merged. Thanks.

PR #600, opened 2015-02-22
No calls to stat() in dvr_inotify_add(). This may significantly slowdown tvh startup.

ksooo:
In my case, my recordings are stored on a NAS that automatically falls asleep if it is not accessed for a certain amount of time. Thus, if my NAS sleeps while tvh starts up, the stat() calls in dvr_inotify_add() awake my NAS which can take up to 15 secs and tvh startup is slowed down by (at least) this 15 secs.

OTOH the stat calls IMO can simply be omitted. I cannot see any drawbacks with adding an inotify_watch without doing stat before. inotify thread will detect later (asynchronously) that the file is not present.

### DIFF
--- a/src/dvr/dvr_inotify.c
+++ b/src/dvr/dvr_inotify.c
@@ -98,18 +98,13 @@ void dvr_inotify_add ( dvr_entry_t *de )
   if (_inot_fd < 0)
     return;
 
-  if (!de->de_filename || stat(de->de_filename, &st))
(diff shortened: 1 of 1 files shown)


ksooo:
I forgot to mention that during tvh startup dvr_inotiy_add is called for every finished recording. The watch will be set for the file containing the recording.

perexg:
Merged. Thanks.

PR #602, opened 2015-02-25
transcoding: fix encoder time_base again

stbenz:
Apparently my last push request still had an error in it. 

It was missing the `ticks_per_frame`, which is usually 2 when decoding h264 or mpeg2 but by default set to 1 in encoder context.

I also changed it to use `framerate` instead of `time_base` from decoder context to calculate `time_base` for encoder context, as reading `time_base` from decoder context is marked deprecated and will probably be removed in a future version of ffmpeg.


perexg:
Merged. Thanks.

PR #605, opened 2015-03-10
transcoding: fix compile error when using libav, fixes #2708

stbenz:
AVCodecContext.framerate and av_make_q() are only available since ffmpeg 2.5 and not available in libav


perexg:
Applied.

PR #606, opened 2015-03-10
Update transcoding.c

mkulicke:
The test of framerate.num and framerate.den fails on newer libavcodec builds (> 56.13.100) according to http://projects.vdr-developer.org/issues/2085


perexg:
Thanks, but it's duplicate to #605 .

PR #608, opened 2015-03-13
capmt: fix endless loop

manio:
When tvh receives less then 4 bytes from the socket for some reason,
it stuck in the endless loop in handle_single() calling:
capmt_msg_size()
capmt_analyze_cmd()
The capmt thread eats 100% CPU and cannot be normally terminated
because the capmt_msg_size() was constantly returning 0.

The commit fixes the problem.


perexg:
Merged. Thanks.

PR #609, opened 2015-03-14
epggrab: allow updating channel name/number/icon

bfennema:
Allow epg grabber to update channel name, number, and/or icon if "Update Channel name", "Update channel number", and/or "Update channel icon" are set in "EPG Grabber" -> "General Config".


perexg:
Merged. Thanks.

PR #616, opened 2015-03-24
Adding %ifarch statement to rpm/tvheadend.spec.in to address i386 and ar...

boblfoot:
...mv7hl build issues present in source.

### DIFF
--- a/rpm/tvheadend.spec.in
+++ b/rpm/tvheadend.spec.in
@@ -32,7 +32,11 @@ to Showtime, XBMC and various other clients.
 
 %build
 echo %{version} > %{_builddir}/%{buildsubdir}/rpm/version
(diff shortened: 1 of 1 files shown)


perexg:
Try with the latest - I added a possibility to disable libx264 in b941aedf93d95e4c01de65a4afb93cb6020e86de with --disable-libffmpeg_static_x264 argument for the configure script.

boblfoot:
While b941aed works for i386 rpm build it still fails for the armhfp build.  my temp patch to the spec file will resolve for now.  But I am wondering what might need to be done in the mean time.

boblfoot:
closing cause patch addresses i386 - will propose arm patch once test builds complete

PR #619, opened 2015-03-26
Patching rpm spec file so the arm architecture builds properly

boblfoot:
modified:   rpm/tvheadend.spec.in


perexg:
Merged. Thanks.

PR #623, opened 2015-04-09
avoid leaving dangling children by killing spawned processes recursively

Jalle19:
This is accomplished by making the spawned process the "group leader" and killing the group instead of just the PID. The group ID is the negation of the PID of the session leader.

This makes it possible to use `pipe://` scripts that spawn child processes, such as rtl_fm or ffmpeg. Without this patch the child processes may remain dangling after the parent has died.


perexg:
Merged.

PR #626, opened 2015-04-11
Some small fixed found by compiling with clang

UllrichKossow:
I found this by compiling with clang.
The fixes are compile clean, but not tested.


perexg:
Merged.

PR #627, opened 2015-04-14
transcoding: set AVFrame format, width and height, fixes #2763

stbenz:
ffmpeg 2.6 checks, if format, width and height is set in avcodec_encode_video2 and generates a warning, if they aren't set.


perexg:
Merged and upgraded to 2.6.2.

PR #628, opened 2015-04-15
Fix dvr_autorec_create_htsp to actually use "fulltext" parameter...

ksooo:
and not to always set fulltext to "1".


perexg:
Merged.

PR #631, opened 2015-04-21
another typo fix in dvb_mux_conf_str_dvbs()

manio:
I believe you was having in mind to fix it like me ;)


perexg:
Yes :-) But I was close.. Merged. Thanks.

PR #634, opened 2015-04-28
linuxdvb_ca: check for CAIDs on the ES level, fixes #2794

dmarion:
Current code is only taking CA descriptor from program level.
Some people reported that it is not working when CA descriptor exists only on the specific ES.
This changes this behaviour. 
Also It reuses results of existing pmt parser instead of parsing again...


perexg:
Merged. Thanks.

PR #635, opened 2015-04-29
Fix crash when trying to configure adapters that are already in use.

nickb937:
If another application has an adapter open, then tvheadend will enter a crash-loop on startup because there's a missing test against 'la' being NULL here.  The in-use adapter will be skipped, allowing other unused adapters to be configured.


perexg:
Merged. Thanks.

PR #638, opened 2015-05-02
Revert to earlier FRITZ!-workaround

WereCatf:
The method suggested by perexg didn't actually work now that I had time to test it properly, it's still not adding pid 21 where it should be. My method, as inelegant as it may be, at least adds the pid there at all times.

### DIFF
--- a/src/input/mpegts/satip/satip.c
+++ b/src/input/mpegts/satip/satip.c
@@ -384,7 +384,8 @@ satip_device_hack( satip_device_t *sd )
     sd->sd_pids_max    = 128;
     sd->sd_pids_len    = 2048;
     sd->sd_no_univ_lnb = 1;
-  } else if (strstr(sd->sd_info.modelname, "FRITZ!")) {
+  } else if (strstr(sd->sd_info.manufacturer, "AVM Berlin") &&
+             strstr(sd->sd_info.modelname, "FRITZ!")) {
     sd->sd_fullmux_ok  = 0;
     sd->sd_pids_deladd = 0;
     sd->sd_pids0       = 1;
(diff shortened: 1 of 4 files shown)


perexg:
OK. Reverted with small changes - d46b3c459cd7690847df1380a2cfbe625d174262 .

StofLE:
I have FRITZ!WLAN Repeater DVB-C wich is shown in Windows or iOS-SAT>IP App. What to do that Tvheadend (I compiled newest version) recognize it?

WereCatf:
StofLE: start tvheadend with the argument:
--satip_xml http://your.box.ip.address:49000/satipdesc.xml

That should bring it visible in tvheadend. If your box has multiple tuners in it you probably won't be able to use them simultaneously, though; with my box if one IP-address tries to use multiple tuners simultaneously it'll just stop responding to all the other streams other than the last one opened, ie. I can only use one tuner.

Wish I could point tvheadend to bind one tuner to one network interface and another tuner to another interface with another IP-address, then I could use multiple tuners.

StofLE:
@WereCatf and @perexg:

Thanks...it's working! :+1:

@WereCatf 
Yes, the problem with simultaneously use of multiple tuners (Repeater DVB-C = 2; FRITZ!Box 6490 = 4) to 1 destination IP seems to be a general problem. Hope AVM will fix it in future firmware. For now I'll only enable 1 tuner.

Stevo83:
The patch is working for me as well (using an AVM DVB-C repeater). I experienced two minor issues:
- When setting up, I selected pre-defined muxes for my region. Tvheadend started automatically scanning the muxes, but failed after scanning the first one. When I enabled them manually one after another, scans were successful.

- The second one might be related to the first: Just like StofLE channel switching (with the kodi frontend) does not work properly. I need to stop the running channel before tuning in to the next. 

Apart from this, setting up and using the device with tvheadend works beautiful. Thank you very much for your work!

PR #640, opened 2015-05-03
http: add interpretation of "X-Forwarded-For" header

stv0g:
This allows to use the builtin ACL when serving TVH behind a SSL reverse
proxy. Before the peer address of the HTTP socket was used.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -676,7 +676,13 @@ process_request(http_connection_t *hc, htsbuf_queue_t *spill)
   char authbuf[150];
 
   hc->hc_url_orig = tvh_strdupa(hc->hc_url);
+
+  v = http_arg_get(&hc->hc_args, "x-forwarded-for");
+  if (v)
+    tcp_get_sockaddr((struct sockaddr*)hc->hc_peer, v);
+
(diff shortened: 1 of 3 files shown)


InuSasha:
You should add check for allowed reverse proxies.
Now, any client can send the header, and avoiding the ACL.
This affects also a non-proxied environment.

perexg:
I agree. It looks like a big security hole.

stv0g:
Sorry for this really embarrassing mistake :-/
I’m feeling really stupid right now.

I will add an option to enable the interpretation manually.
Do you prefer this to be done with a global setting or one attached to the ACL entries?

Steffen

perexg:
@stv0g : I think it belongs to the global config, so the X-Forwarded-For will be used only when the source IP address is in the proxy whitelist defined in the global config.

nbittmann:
Just to bring this up again as I believe it's still an issue? Can we get support for the X Forward header at some point? I'm using a simple Apache mod_proxy setup and I'm very restricted in terms of acl because I only see my proxies address a feature like this would solve it.

perexg:
Added to latest (configurable).

PR #641, opened 2015-05-05
Tunercfg override

WereCatf:
Some devices, at least my FRITZ!Box Cable 6490, report wrong count of tuners in the SAT-IP description XML-file. I added a new option called "Override tuner control" (sd_tunercfg_override) that, if 1-32, overrides the number reported in the XML-file. Also added explanation for this in the web-GUI help.

Stay tuned for more FRITZ!-stuffs.

### DIFF
--- a/src/input/mpegts/satip/satip.c
+++ b/src/input/mpegts/satip/satip.c
@@ -224,6 +224,13 @@ const idclass_t satip_device_class =
       .off      = offsetof(satip_device_t, sd_bindaddr),
     },
     {
+      .type     = PT_BOOL,
+      .id       = "disablefritz",
+      .name     = "Disable FRITZ!-specific workarounds",
+      .opts     = PO_ADVANCED,
+      .off      = offsetof(satip_device_t, sd_no_fritz_workarounds),
+    },
+    {
       .type     = PT_STR,
       .id       = "addr",
       .name     = "IP Address",
@@ -398,7 +405,15 @@ satip_device_hack( satip_device_t *sd )
     sd->sd_pids_max    = 128;
     sd->sd_pids_len    = 2048;
     sd->sd_no_univ_lnb = 1;
+  } else if (strstr(sd->sd_info.manufacturer, "AVM Berlin") &&
+             strstr(sd->sd_info.modelname, "FRITZ!") &&
+             sd->sd_no_fritz_workarounds != 1) {
+    sd->sd_fullmux_ok  = 0;
+    sd->sd_pids_deladd = 0;
+    sd->sd_pids0       = 1;
+    sd->sd_pids21       = 1;
   }
+
 }
 
 static satip_device_t *
@@ -423,15 +438,6 @@ satip_device_create( satip_device_info_t *info )
   sd->sd_pids_deladd = 1;
   sd->sd_sig_scale   = 240;
   sd->sd_dbus_allow  = 1;
-
-  /* safe defaults for FRITZ!-devices */
-  if (strstr(info->manufacturer, "AVM Berlin") &&
-             strstr(info->modelname, "FRITZ!")) {
-    sd->sd_fullmux_ok  = 0;
-    sd->sd_pids_deladd = 0;
-    sd->sd_pids0       = 1;
-    sd->sd_pids21 = 1;
-  }
 
   if (!tvh_hardware_create0((tvh_hardware_t*)sd, &satip_device_class,
                             uuid.hex, conf)) {
--- a/src/input/mpegts/satip/satip.c
+++ b/src/input/mpegts/satip/satip.c
@@ -225,10 +225,10 @@ const idclass_t satip_device_class =
     },
     {
       .type     = PT_BOOL,
-      .id       = "disablefritz",
-      .name     = "Disable FRITZ!-specific workarounds",
+      .id       = "disableworkarounds",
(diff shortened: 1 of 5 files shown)


perexg:
Did you do a test with 4 simultanteous streams from different muxes ? I just want to know, if it's a firmware bug or a hardcoded restriction. Also someone reported that only one SAT>IP stream can be used for this device due to other bugs.

WereCatf:
I can play 4 simultaneous streams from different muxes just fine as long as the streams go to different IP-addresses, but if I try to play even just two simultaneous stream to the same IP one of them stops working as soon as the second starts.

I was planning on seeing if I could add a per-tuner address binding option in tvheadend; if the user has multiple NICs they could then use one NIC per tuner and solve that issue. I have no way of fixing things on the box's end.

src/input/mpegts/satip/satip.c
```diff
@@ -203,6 +203,20 @@ const idclass_t satip_device_class =
       .off      = offsetof(satip_device_t, sd_pilot_on),
     },
     {
+      .type     = PT_BOOL,
+      .id       = "fritzquirks",
+      .name     = "Enable FRITZ!Box-workarounds",
```
perexg: I don't like it so much. What about to rename this as 'pids21' which is more universal.

src/input/mpegts/satip/satip.c
```diff
@@ -508,6 +525,12 @@ satip_device_create( satip_device_info_t *info )
     }
   }
 
+  if (sd->sd_fritz_quirk == 1) {
```
perexg: Sorry, but having two override places for one hw is too much for me...

src/input/mpegts/satip/satip.c
```diff
@@ -384,12 +398,6 @@ satip_device_hack( satip_device_t *sd )
     sd->sd_pids_max    = 128;
     sd->sd_pids_len    = 2048;
     sd->sd_no_univ_lnb = 1;
-  } else if (strstr(sd->sd_info.manufacturer, "AVM Berlin") &&
```
perexg: Do not move this...

src/input/mpegts/satip/satip.c
```diff
@@ -493,6 +509,7 @@ satip_device_create( satip_device_info_t *info )
       m = atoi(argv[i] + 6);
       v2 = 2;
     }
+    if (sd->sd_tunercfg_override > 0 && sd->sd_tunercfg_override < 33) m = sd->sd_tunercfg_override;
```
perexg: Place assignment to the new line..

src/input/mpegts/satip/satip.c
WereCatf: I don't understand what you mean with this.

perexg: I meant - try to name the quirk using an unversal name and do only the "default" settings when the hw is detected. You're trying to override other settings when sd->sd_fritz_quirk is set by user which may not be wanted.

src/input/mpegts/satip/satip.c
WereCatf: I'm not sure what you mean here, either. If I leave those lines there tvheadend will always apply those settings and ignore user-set preferences. I wanted them optional just in case AVM ever releases a firmware update, and thus I moved the check earlier so that there are sane defaults.

WereCatf:
Well, I think I fixed all the things you mentioned.

src/input/mpegts/satip/satip.c
perexg: OK. I see your point (I thought that the hacks are applied before the settings is loaded, but it's after this operation). I prefer to change the code in this case (when the firmware is fixed). This settings is really for advanced users and I prefer to check the firmware capabilities manually. But for "unknown" firmware, all settings should be available. Perhaps, we can add an option 'do not apply any automatic hacks' rather..

WereCatf: Well, the way I set it up now is that by default it enables safe defaults for FRITZ!-devices, but the settings are saved in the config files and the user can turn them on and off at will. I do not think having to modify and recompile sources just to fix the settings if a new firmware lands is the right way to go.

perexg: Think about this in other way - user will override the default settings and it won't work with the current firmware. Again, if you like, add option to disable all known workarounds, so users can hack the settings without any auto-things.. But... keep these auto things in one place (hack function).

WereCatf: Okay, I moved the code back to hack function and added a new option called "Disable FRITZ!-specific workarounds" -- if it's not set the hacks are forced on, if it is set the hack is ignored and user's own settings will prevail.

src/input/mpegts/satip/satip.c
```diff
@@ -210,6 +224,13 @@ const idclass_t satip_device_class =
       .off      = offsetof(satip_device_t, sd_bindaddr),
     },
     {
+      .type     = PT_BOOL,
+      .id       = "disablefritz",
+      .name     = "Disable FRITZ!-specific workarounds",
```
perexg: Great, but could you change this as more universal 'No firmware-specific workarounds' which will apply for all hacks ?

WereCatf: I misunderstood you there. It's now universal.

perexg:
Merged. Thanks.

PR #643, opened 2015-05-06
Frontend workarounds

WereCatf:
If workarounds are enabled then turn "full PLAY cmd" - setting on for FRITZ-devices' tuners. Have to call satip_device_hack again after frontends have been created since the first call happens before that, that's why it's called twice.

### DIFF
--- a/src/input/mpegts/satip/satip.c
+++ b/src/input/mpegts/satip/satip.c
@@ -385,6 +385,7 @@ satip_device_calc_uuid( tvh_uuid_t *uuid, const char *satip_uuid )
 static void
 satip_device_hack( satip_device_t *sd )
 {
+  satip_frontend_t *lfe;
   if(sd->sd_disable_workarounds)
       return;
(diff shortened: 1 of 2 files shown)


src/input/mpegts/satip/satip.c
```diff
@@ -409,6 +410,8 @@ satip_device_hack( satip_device_t *sd )
     sd->sd_no_univ_lnb = 1;
   } else if (strstr(sd->sd_info.manufacturer, "AVM Berlin") &&
              strstr(sd->sd_info.modelname, "FRITZ!")) {
+    TAILQ_FOREACH(lfe, &sd->sd_frontends, sf_link)
+             lfe->sf_play2 = 1;
```
perexg: This setup should be in satip_frontend_hacks().

WereCatf:
Well, there, fixed.

perexg:
Merged. Thanks.

PR #645, opened 2015-05-13
Fix unsigned int 64 overflow in timeshift

Memphiz:
As the title says.

This fixes bug https://tvheadend.org/issues/2712

Regression introduced in:

https://github.com/tvheadend/tvheadend/commit/673549141d5d51988e1ee9bdc30ed5a7cc32bf8a

Calling atomic_add_u64(&timeshift_total_size, -tsf->size); while the second parameter is an uint64_t leads to wrapping and at the end our timeshift buffer is claimed full and no timeshift is possible anymore.

### DIFF
--- a/src/atomic.h
+++ b/src/atomic.h
@@ -52,6 +52,21 @@ atomic_add_u64(volatile uint64_t *ptr, uint64_t incr)
 }
 
 static inline uint64_t
+atomic_dec_u64(volatile uint64_t *ptr, uint64_t decr)
+{
+#if ENABLE_ATOMIC64
+  return __sync_fetch_and_sub(ptr, decr);
+#else
+  uint64_t ret;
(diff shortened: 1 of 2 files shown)


perexg:
Merged. Thanks.

Memphiz:
@perexg how do you merge without hitting the merge button and without having the merge commit in your history? *curious*

perexg:
Manual quick merge:

$ wget https://github.com/tvheadend/tvheadend/pull/645.patch
$ git am 645.patch

done...

Memphiz:
ahh ok - well sadly this approach removes the [prefix] from the commit message - so nothing for us in the kodi project :) - but thx for sharing...

perexg:
Use 'git am -k' to not remove [] text.. (I had to do it in this case, sure).

PR #646, opened 2015-05-16
Fix build in symlinked/mounted dirs

RustyDust:
Just a small fix to prevent "Wrong filename" errors when building from a symlinked directory where the target is on an NFS share.


perexg:
Thanks. Merged.

PR #647, opened 2015-05-18
WebUI static path definitions for icons/images/extjs resources

RustyDust:
Needed to add some web_static_content definitions to get the icons back in the web interface. There may be an easier solution but that's what I could come up with ;)

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -1419,6 +1419,10 @@ webui_init(int xspf)
   webui_static_content("/docs",          "docs/html");
   webui_static_content("/docresources",  "docs/docresources");
 
+  webui_static_content("/extjs/resources",     "src/webui/static/extjs/resources");
+  webui_static_content("/icons",               "src/webui/static/icons");
+  webui_static_content("/img",                 "src/webui/static/img");
+
   simpleui_start();
   extjs_start();
   comet_init();


perexg:
This is wrong. What url paths do you have in src/webui/static/tvh.css.gz ?

RustyDust:
Sorry for the late reply, I put the unpacked file up here: http://whocares.de/tvh.css

perexg:
The contents of the file is wrong. Actually, with latest master, the file should be like:

<pre>
-rw-rw-r--  1 perex perex    842 May 23 20:16 tvh-tv.css.gz
</pre>

unpacked:

<pre>
-rw-rw-r--   1 perex perex 130854 23. kvě 20.16 tvh.css
</pre>

You may give a look to rules in Makefile.webui for this files and determine the failure point for your environment. I tested the python helpers with both python2 and python3 - same result here.

RustyDust:
What should the pathes look like? Prefixed with /static ?

perexg:
URLs should be releative (without first '/' character).

ksooo:
Broeken for me as well with latest tvheadend master. No icons in webui. Using openelec latest master build environment.

@stefansaraev any ideas what's going wrong?

stefansaraev:
@ksooo no clue. I stopped caring long time ago.

ksooo:
@stefansaraev too bad that tvheadend is so "volatile".

RustyDust:
I fixed it for my build environment with the patch in pull request #654: https://github.com/tvheadend/tvheadend/pull/654

perexg:
Fixed in 247dd3a09f0f7c84df858e27e71f0706d50ccbe5 . Thanks for info.

RustyDust:
Confirming that https://github.com/tvheadend/tvheadend/commit/247dd3a09f0f7c84df858e27e71f0706d50ccbe5 works for me.

perexg:
@ksooo: Use release/4.0 branch for not so volatile sources.

perexg:
@RustyDust : thanks for the confirmation.

PR #650, opened 2015-05-21
python client: Fix broken int2bin() in htsmsg

ruuk:
The int2bin() method was incorrectly shifting the bits by 16 for the third character instead of 8, which was causing issues for any messages with a length greater than 256. The fixes that :)


perexg:
Merged. Thanks.

PR #651, opened 2015-05-21
[en50494] mark as experimental for release

InuSasha:
The UniCable support is not stable enought.
Actual, only same installations are working, but i can't find a
solution.
But it does not brother user, who no use unicable. So we don't have to
remove it from source. I hope we can find one expert, who can help us to
improve the situation.


perexg:
Merged. Thanks.

PR #652, opened 2015-05-22
http: fix Accept-Encoding parsing

stbenz:
http_encoding_valid() only worked, if the requested encoding was at the beginning of the Accept-Encoding header. This usually worked, as the function is only called with encoding "gzip" and "gzip" is usually at the beginning of the Accept-Encoding header.

But, when the HTML5 web player is opened in Google Chrome, it sends "Accept-Encoding: identity;q=1, *;q=0", which causes an endless loop.

This fix should do the correct parsing of the Accept-Encoding header according to http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -1,6 +1,6 @@
 /*
  *  tvheadend, HTTP interface
- *  Copyright (C) 2007 Andreas Öman
+ *  Copyright (C) 2007 Andreas Ã–man
  *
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

perexg:
Merged.

PR #653, opened 2015-05-22
transcoding: optimized video and audio encoder settings

stbenz:
This has the following optimizations:
- removed unnecessary settings
- better defaults (for bitrate = 0)
- use correct parameters for bitrate < 64k in video encoder (global_quality/crf instead of qmin/qmax)
- trade off some encoding latency for better compression in video encoder for bitrate > 64k

I only tested this with --enable-libffmpeg_static and also not with much devices (only browser and VLC). So it should be tested with linking against libav in debian and maybe some additional devices.

### DIFF
--- a/docs/html/config_streamprofile.html
+++ b/docs/html/config_streamprofile.html
@@ -1,41 +1,138 @@
 <div class="hts-doc-text">
 
-<p>
-<br>
-<hr>
-<b>Buttons</b>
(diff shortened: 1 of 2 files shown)


perexg:
Merged. Thanks. Please, watch for compilation/usage errors in the issue tracker to fix them ASAP.

PR #654, opened 2015-05-23
Fix#2 for build in symlinked dirs (fixes #647)

RustyDust:
Found the location where the path names were built incorrectly when building in a symlinked dir.
Tested with builds in real dir and symlinked dir, produces same output in both now.
The old pull request can be safely ignored/closed.


perexg:
Fixed in 247dd3a09f0f7c84df858e27e71f0706d50ccbe5 . Thanks for hint.

PR #656, opened 2015-05-24
[htsp][autorec,timerec] Handle "retention", "startExtra", "stopExtra"…

ksooo:
… defaults the same way as for normal dvr entries (https://github.com/tvheadend/tvheadend/blob/master/src/htsp_server.c#L687). Do return the current dvr config values instead of 0, which cannot be handled/interpreted by htsp clients (there is no htsp dvr config API).


perexg:
Merged (to both 4.1 and 4.0). Thanks.

PR #659, opened 2015-05-28
Add trusty (14.04 LTS) build target

dreamcat4:
debian: So that ubuntu package names will appear as '4.0.0-11~gd149c5b~trusty amd64'

It reflects more accurately now the packages were build on `trusty` because that is the current LTS (Long-Term-Service) release. And nobody uses `precise` anymore.

This commit is needed on both the master + stable branches. Tested on my fork:

https://gist.github.com/dreamcat4/f470ed87f650e6752ef8#comment-1462233

### DIFF
--- a/Autobuild/trusty-amd64.sh
+++ b/Autobuild/trusty-amd64.sh
@@ -1,0 +1,3 @@
+AUTOBUILD_CONFIGURE_EXTRA="${AUTOBUILD_CONFIGURE_EXTRA:-} --arch=x86_64"
+DEBDIST=trusty
+source Autobuild/debian.sh
--- a/Autobuild/trusty-i386.sh
+++ b/Autobuild/trusty-i386.sh
@@ -1,0 +1,3 @@
+AUTOBUILD_CONFIGURE_EXTRA="${AUTOBUILD_CONFIGURE_EXTRA:-} --arch=i686"
+DEBDIST=trusty
+source Autobuild/debian.sh


ProfYaffle:
@dreamcat4 - are we up to date with other supported releases - utopic, vivid (though it looks like utopic dies in the next month or two)?

We'll soon be getting requests for Whispering Wombat or whatever 15.10 is called...

It might also be worth thinking about the Mint releases. They should declare the underlying Ubuntu version, but they don't, possibly by design - maybe a simple symlink would sort that out for them, or a copy/rename of the files.

dreamcat4:
This PR is just about being transparent, to show up in `apt-get` tool the correct `trusty` substring which is not misleading. Since currently it says `precise` - which is no longer supported.

@ProfYaffle the `.debs` are being built on trusty however they should work on `14.10` and `15.04` just the same. The reason why I'm not building 3 seperate versions is because that would triple the server load (compiling everything 3 times).

However if a user reports any issue with these trusty builds on a newer version of ubuntu, I would be happy to add the extra, seperate build target. It's just not needed right now.

For example: I run today the trusty deb on my `15.04` vivid host. It seems OK. More thorough testing will eventually be done (by me) on docker images with the 2 newer ubuntu versions.

Again - when `15.10` come out it's same story. We will make a seperate build chain only if user has reported an issue that can only be solved by compiling on that (newer) ubuntu version. E.g. shared lib version mismatch etc. Which is not very likely ATM because the resulting package has so few shared lib deps. Even when everything enabled.

dreamcat4:
I have not thought about Mint releases. So not sure what you mean be marking them as ubuntu etc. As I've never used mint. But AFAIK the pkgs aught to work no mint for same reasons as above. Perhaps even debian too.

So it's the same plan:

Let some people to try with the trusty pkg on Mint / Debian / etc first of all. If it works then that's fine. Not much for us to worry about. Or then if it didn't work, then that's an appropriate time to consider taking extra steps and/or building a seperate line of `.deb` packages specifically for those affected distro(s). 

As I've stated elsewehere (in the associated build docs), I myself am only interested in making `.deb` pkgs for ubuntu. However I will fully support anyone else who wants to set them up with the necessary guidance / etc.

ProfYaffle:
MMmm, my point was that apt-add repository by default will simply append the distro name (from lsb_release -a, I think) and then subsequent apt-gets will look for packages of that name. So, while the debs will work on other distros more than likely, apt-get won't find them as it'll be looking for the wrong package ==> support overhead unless it's clearly in an FAQ.

The only options that I see are:

1. Ignoring this and telling people to edit /etc/sources by hand to add a Trusty (or whatever) target - maybe tricky for new Linux users

2. Rebuilding under the correct target distro in a different VM/image - probably the right answer but the greatest overhead

3. Renaming the debs after compilation so they appear under the right names - a complete frig IMO but the easiest

The Mint angle is unique, in that both Mint Qiana and Mint Rebecca will look for a package for that name rather than the 14.04 'buntu base distro, which would be Trusty, as above. I don't know how anyone else handles this, as I consider that to be a deliberate quirk on Mint's part - I suspect most folks (e.g. Kodi, mkvmerge, LibreOffice) just ignore it as Somebody Else's Problem.

dreamcat4:
@ProfYaffle I didn't see any such issue today when I installed them on 15.04 vivid. It worked fine. The target field has now been changed to specify the branch, not the distro version as was previously. So v2 pkg will be a choice of `<master|unstable|testing|stable> main` and not `<trusty|vivid> main`. It's better that way as fewer repos needed (just 1 instead of 4).

perexg:
Merged.

dreamcat4:
@perexg Can you please also add this commit to the `release/4.0` branch ?

perexg:
Done.

PR #661, opened 2015-05-31
config: make server name configureable (avahi, satips)

dmarion:
This should avoid confusion when multiple tvh instances are running on the network.
Whatever is configured as server name will show up in avahi and sat>ip announcements.

To support this I added ability to restart avahi service announcements.


perexg:
Merged. Thanks.

PR #663, opened 2015-06-05
Fixed cross-compilation errors (master)

nurtext:
Especially when compiling for Synology DSM.

### DIFF
--- a/src/descrambler/capmt.c
+++ b/src/descrambler/capmt.c
@@ -652,10 +652,11 @@ capmt_queue_msg
   sbuf_append(&msg->cm_sb, buf, len);
   msg->cm_adapter = adapter;
   msg->cm_sid     = sid;
-  if (flags & CAPMT_MSG_FAST)
+  if (flags & CAPMT_MSG_FAST) {
     TAILQ_INSERT_HEAD(&capmt->capmt_writeq, msg, cm_link);
-  else
+  } else {
     TAILQ_INSERT_TAIL(&capmt->capmt_writeq, msg, cm_link);
+  }
   tvh_write(capmt->capmt_pipe.wr, "c", 1);
 }
 
--- a/src/epggrab/otamux.c
+++ b/src/epggrab/otamux.c
@@ -236,10 +236,11 @@ epggrab_ota_start ( epggrab_ota_mux_t *om, mpegts_mux_t *mm )
   int grace;
 
   /* In pending queue? Remove.. */
-  if (om->om_q_type == EPGGRAB_OTA_MUX_PENDING)
+  if (om->om_q_type == EPGGRAB_OTA_MUX_PENDING) {
     TAILQ_REMOVE(&epggrab_ota_pending, om, om_q_link);
-  else
+  } else {
     assert(om->om_q_type == EPGGRAB_OTA_MUX_IDLE);
+  }
 
   TAILQ_INSERT_TAIL(&epggrab_ota_active, om, om_q_link);
   om->om_q_type = EPGGRAB_OTA_MUX_ACTIVE;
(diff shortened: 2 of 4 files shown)


perexg:
Uff. It looks like that the used TAILQ macros are wrong. The problem is that there may be more errors than compilation ones. Could you upload somewhere the used <sys/queue.h> file ?

nurtext:
@perexg I didn't used any file special file and the macros already did exist, except the TAILQ_EMPTY one which I added from the linux docs.

perexg:
Sorry, appearently &lt;&gt; must be typed in html. Could you upload somewhere the used <sys/queue.h> file ?

nurtext:
@perexg I still don't get it. Sorry. I just added the missing TAILQ_EMPTY macro. No special file, just the macro with #ifndef statement just in case sys/queue.h is missing it - as you can see in http://man7.org/linux/man-pages/man3/queue.3.html

perexg:
The problem is that macros should define the contents in {} blocks to not void the if something then something without brackets. Usually 'do {} while (0)' construct is used.

nurtext:
@perexg Okay, but almost every source for sys/queue.h which can be found is using the macros without brackets. See:

http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/queue.h
http://unix.superglobalmegacorp.com/xnu/newsrc/bsd/sys/queue.h.html
etc.

perexg:
The 'do {} while(0)' is OK. If you have this for TAILQ_INSERT_HEAD() than the compiler should not complain. What's the compilation error in your case?

nurtext:
@perexg Seems we got things terribly wrong in this discussion. I will clearify: TAILQ_EMPTY is missing in sys/queue.h when cross compiling for Synology DSM. Compiler was complaining about this. Someone else decided to hardcode the macro in form of a sourcecode patch during cross compilation. I thought it would be better to add the missing TAILQ_EMPTY macro using the #ifndef directive. In addition to that, the compiler complains about missing curly brackets, which I added and included into this PR as well. After doing this all cross compilation errors are gone and tvheadend runs fine.

perexg:
My all questions were for "In addition to that, the compiler complains about missing curly brackets, which I added and included into this PR as well. ". There's no reason that compiler complains when macros are using the 'do { code } while(0)' construct. It should already cover the {} need. The TAILQ_EMPTY extension is fine.

nurtext:
@perexg Sadly the compiler does complain about this :cry: 
```
src/service.c: In function 'service_destroy':
src/service.c:860: error: expected expression before 'else'
src/service.c: In function 'service_create0':
src/service.c:916: error: expected expression before 'else'
```

perexg:
@nurtext : Then, could you show me the used &lt;sys/queue.h&gt; file ?

nurtext:
Sure:
https://gist.github.com/nurtext/1cec1dcf90bd0ebda78a

From Synology's official cross-compiling source:
http://sourceforge.net/projects/dsgpl/files/DSM%204.3%20Tool%20Chains/Intel%20x86%20Linux%203.2.40%20%28Bromolow%29

nurtext:
@perexg And sorry for misunderstanding you in the first place :wink: Hope you're not mad with me now…

perexg:
Thanks. I added own copy of sys/queue.h version 8.5 to the tvheadend sources and removed all duplicate definitions from src/queue.h .

The problem is that old version 8.3 uses define {} instead define do {} while(0) which causes problems like:

<pre>
  if (cond)
    { macro_code };
  else
   something;
</pre>

instead:

<pre>
  if (cond)
   do { macro_code } while(0);
 else
  something;
</pre>

The ';' in first example means end-of-condition, so 'else' is correctly reported at wrong place.

nurtext:
@perexg Could please make a backport for release 4.0?

PR #666, opened 2015-06-09
webui: add build time and configure options to status page

stbenz:
Note, that with this change two C-files are auto generated. I successfully tested building with Linux Mint and Debian, but it could still break building on other boxes...

Fixes #2928 (at least some of it. I don't know, if we really should add system info not directly related to tvheadend)


perexg:
Merged. Thanks.

PR #667, opened 2015-06-09
Build system: some minor improvements

stbenz:
Added static hdhomerun build dir to .gitignore.
Delete data/dvb-scan on `make distclean`.
Add dependency for Makefile to static hdhomerun and ffmpeg, so they get rebuilt when the Makefile changes.
Removed some clutter from the build output of static hdhomerun and ffmpeg.
Updated static libvorbis.


perexg:
Merged. Thanks.

PR #668, opened 2015-06-11
WIP: Webhelp revisions

ProfYaffle:
Right, so people can see it - while it's mergeable, it will need more updating and would benefit from some discussion (and probably correction, I don't pretend to be a CSS expert). 

So...  **"caveat patcher"**.

This PR basically:

1. Replaces the bundled html webUI help files with auto-generated versions of the User Guide chapters

  **These will constantly change until we have stable documentation - you can see the gaps in the examples below**

  **The use of pandoc as a translator is also under review, but the sense of 'write once, use many' is valid**

2. Adds a chunk of CSS to style the webUI help (I did just find some 6-year-old styling, but I think it's getting ignored - I did this from scratch anyway)

  **I'm fairly happy with this, but I may have committed some terrible crime against W3C**

3. Tidies up a few dead files that don't appear to be used any more.

Examples of styled help screens:

image

image


perexg:
I merged and modified (again) the css styles. Also, "Additional screenshot" and "Tidying up orphaned webUI files" were pushed to master.

I'll work on makefile changes to sync the webui html pages automatically and intl support (redirections).

perexg:
I moved docs/html/ files to docs/html/en subtree and imported the initial version generated from markdown manually. Just for testing..

Flole998:
This is super outdated and it's pretty much superseded by the new documentation and also the way the docs are/should be handled now

PR #672, opened 2015-06-18
FIX for EPG broadcasted on different PID for provider Bulsatcom

zinonino:
Fix for EPG in Simulcrypt channels from Dolce for Provider Bulsatcom 39E


src/input/mpegts.h
```diff
@@ -355,6 +355,7 @@ enum mpegts_mux_epg_flag
   MM_EPG_ONLY_UK_FREESAT,
   MM_EPG_ONLY_UK_FREEVIEW,
   MM_EPG_ONLY_VIASAT_BALTIC,
+  MM_EPG_ONLY_BULSATCOM_39E,
```
perexg: This break compatibility - new values must be last, because these values are used in config files, too.

zinonino:
OK. Thanks for note. If all other is right, do you merge to master ?

perexg:
Merged. Thanks.

PR #680, opened 2015-06-21
intl: 'edit' translation changed

alexkn:
need different translation for german


perexg:
Merged and pushed new po template to transifex.

PR #684, opened 2015-06-23
Disable legacy and deprecated OSCam capmt modes

manio:
I think it would be nice to start disabling those modes...

### DIFF
--- a/src/descrambler/capmt.c
+++ b/src/descrambler/capmt.c
@@ -2161,10 +2161,12 @@ caclient_capmt_class_oscam_mode_list ( void *o, const char *lang )
   static const struct strtab tab[] = {
     { N_("OSCam net protocol (rev >= 10389)"), CAPMT_OSCAM_NET_PROTO },
     { N_("OSCam pc-nodmx (rev >= 9756)"),      CAPMT_OSCAM_UNIX_SOCKET },
+#if 0
     { N_("OSCam TCP (rev >= 9574)"),           CAPMT_OSCAM_TCP },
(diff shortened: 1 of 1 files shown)


Jalle19:
@manio I'd like this as well, but I think all the code should be removed too, not just ifed out.

manio:
@Jalle19 
Me too, but the ifdefing was a @perexg idea, when we was discussing it on IRC. The plan was to disable it the way to to be able to easily revert in case of problems, then after some time when no one complains, clean the code up.

@perexg any news on this?

perexg:
I plan to merge this to 4.3 .

Trujulu:
We are on 4.3 now :o

perexg:
Removed in 2403a543a9aef03266edff004c10e2458acaeb30 .

manio:
Thanks! :)

renedis:
Not happy.. Still using the rev-9756 mode with an Oscam/Irdeto setup. Does not work anymore!

manio:
Any reason to not updating your oscam?

renedis:
Oscam is running on:
OSCam: 1.20-unstable_svn Build: r11401 with rev-9756 and "/tmp/camd.socket" option.

When using "OSCam new pc-nodmx (rev >= 10389)" the descrambler only works the first time (first channel). After that TVHeadend needs to be restarted in order to descramble again.

I'm not the first with this issue:
https://tvheadend.org/issues/4038

perexg:
Try the standard TCP connection instead named socket. Appearently, the oscam code remains somewhat broken for the named socket.

renedis:
Took some investigation but got TCP connection working now.

PR #690, opened 2015-06-27
Fix invalid usages of types, invalid memory usage.

KyleSanderson:
Pretty simple PR, used a static analyzer and looked at the detections. Couple char comparisons against pointers, few NULL pointer usages.


perexg:
Merged. Thanks.

PR #691, opened 2015-07-01
Smal fix translate

PiterEL:
I translate this "0" zero because i can not translate this strings on Transifex. I dont know why? Please update this translation to Transifex.


perexg:
It may be browser issue. Fixed now.

PR #693, opened 2015-07-03
Add per network tags

TheUlpio:
This is useful in multi-adapter equipments. Adding the network (source) tag to each channel lets to group in clients like kodi simply choosing the proper (source) tag in group list (eg. DTT, HorBird, Astra, etc.). This is also useful in rotor setups, grouping channels by satellite avoids the continuous move of the dish that happens in an unique enormous list of mixed-sources channels


perexg:
Merged as one commit 51f4b179a2d89198bb281b2cf6d46e350969d1ca . Thanks for the contribution.

PR #695, opened 2015-07-08
rpm: git is required by dvb-scan

daniviga:
Without `git` in the `BuildRequires` the RPM build will fail during `./configure` with the error
```bash
  fetching dvb-scan files ...                       fail
ERROR: Failed to fetch dvb-scan data (use --disable-dvbscan)
```
because the `getmuxlist` requires `git`: https://github.com/tvheadend/tvheadend/blob/master/support/getmuxlist#L21

With this fix I was able to build RPMs for F21 and F22 using `mock`.


perexg:
Added. Thanks.

PR #696, opened 2015-07-09
Adding CORS headers

shukydvir:
Hi,

This implements 

https://tvheadend.org/issues/852

I'm currently building a web UI and I want it to be static HTML no server side code.

Thanks

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -231,9 +231,13 @@ http_send_header(http_connection_t *hc, int rc, const char *content,
   htsbuf_qprintf(&hdrs, "%s %d %s\r\n", 
 		 http_ver2str(hc->hc_version), rc, http_rc2str(rc));
 
-  if (hc->hc_version != RTSP_VERSION_1_0)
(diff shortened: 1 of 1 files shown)


perexg:
I don't like it. The allowed IPs / hostnames should be configurable.

Jalle19:
I don't see the problem with this, these headers are only used by browsers.

Jalle19:
Any news on this? Now that there's been more work on unifying the API there are probably people interested in creating third-party web interfaces for tvheadend. CORS is a hard requirement if that's going to be possible.

@perexg http://enable-cors.org/ has more information

perexg:
I'm busy with the API rewrite for configuration tabs at the moment, but the goal of all of these rewrites is to allow to add new various configuration switches to the configuration tabs in an easy way. So, hold on.

perexg:
Merged and added: 198ecc37f7ef8b5af75f82ef8035d49ee8996fbc

PR #697, opened 2015-07-13
linuxdvb: handle possible bad pilot config

adamsutton:
This is no doubt a result of some previous configuration migration issues.

However this change should be fairly benign and will stop FE's reporting
unecessary tuning errors.

Not sure though if the best error is to fix the configuration?


adamsutton:
@perexg I did the fix in a bit of a rush to get myself upgraded to v4.x and yes it does make things work, however it's fixing the wrong thing. The bug is in the table definition, I will fix, push and close this.

However we need to decide how best to manage the release branches etc...

PR #698, opened 2015-07-16
distinguist 'May' and its short version

diraimondo:
Add the possibiliy to translate the short version of May in other languages.


diraimondo:
I'm not sure if it is necessary to apply the split function to remove the #-postfix or if it will be applied later...

perexg:
Applied. Thanks.

PR #699, opened 2015-07-22
htsp: Extend 'timeshiftStatus' server-to-client method by start and end as time

ksooo:
... includes htsp version bump to 23


ksooo:
Just discovered that it does not yet work. Closing PR for now.

PR #700, opened 2015-08-06
Fix GMT offset during DST on vanilla glibc systems

metaron-uk:
Requested by @ksooo while investigating a pvr.hts issue (https://github.com/kodi-pvr/pvr.hts/pull/85)

Tested on my gentoo server (not a vanilla glibc system) but I can confirm that it is using the new code path successfully and provides the correct time offset when tested against https://github.com/kodi-pvr/pvr.hts/pull/85 in the Europe/London timezone during DST.

The idea for this change came from http://stackoverflow.com/questions/635780/why-does-glibc-timezone-global-not-agree-with-system-time-on-dst

Known 'vanilla glibc' systems include rasbian (aka debian) and openElec.
Known 'not vanilla' glibc systems include gentoo and ubuntu.

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -65,7 +65,7 @@
 
 static void *htsp_server, *htsp_server_2;
 
-#define HTSP_PROTO_VERSION 22
+#define HTSP_PROTO_VERSION 23
 
 #define HTSP_ASYNC_OFF  0x00
 #define HTSP_ASYNC_ON   0x01
@@ -1088,13 +1088,33 @@ htsp_method_getSysTime(htsp_connection_t *htsp, htsmsg_t *in)
   htsmsg_t *out;
   struct timeval tv;
   struct timezone tz;
+  int tz_offset;
+  struct tm serverLocalTime;
 
   if(gettimeofday(&tv, &tz) == -1)
     return htsp_error("Unable to get system time"); 
 
+  if (!localtime_r(&tv.tv_sec, &serverLocalTime))
+    return htsp_error("Unable to get system local time");
+#if defined(HAS_GMTOFF)
+  tz_offset = - serverLocalTime.tm_gmtoff / (60);
+#else
+  // NB: This will be a day out when GMT offsets >= 13hrs or <11 hrs apply
+  struct tm serverGmTime;
+  if (!gmtime_r(&tv.tv_sec, &serverGmTime))
+    return htsp_error("Unable to get system gmt");
+  tz_offset = (serverGmTime.tm_hour - serverLocalTime.tm_hour) * 60;
+  tz_offset += serverGmTime.tm_min - serverLocalTime.tm_min;
+  if (tz_offset > 11 * 60)
+    tz_offset -= 24 * 60;
+  if (tz_offset <= -13 * 60)
+    tz_offset += 24 * 60;
+#endif
+
   out = htsmsg_create_map();
   htsmsg_add_s32(out, "time", tv.tv_sec);
-  htsmsg_add_s32(out, "timezone", tz.tz_minuteswest / 60);
+  htsmsg_add_s32(out, "timezone", tz_offset/60);
+  htsmsg_add_s32(out, "gmtoffset", -tz_offset);
   return out;
 }
 
--- a/configure
+++ b/configure
@@ -185,6 +185,17 @@ int test(void)
 }
 '
 
+check_cc_snippet gmtoff '
+#include <time.h>
+#define TEST test
+int test(void)
+{
+  struct tm x;
+  x.tm_gmtoff = 0;
+  return 0;
+}
+' -DHAS_GMTOFF
+
 check_cc_snippet recvmmsg '
 #define _GNU_SOURCE
 #include <stdlib.h>


metaron-uk:
@ksooo if you can point me in the direction of how to bump the version like you requested I'll add it as a second commit, but it wasn't immediately obvious.

metaron-uk:
@ksooo: Never mind I found it !DOH! 
Squashed into the one commit so you can see more easily what changed the version.

metaron-uk:
I'm re-thinking this change. 

1. Doing a bit more digging it probably isn't portable enough
2. The test I've used really ought to check for the presence of tm_gmtoff rather than 'linux'. (It's originally a BSD parameter!)
3. the conversion outside 'linux' was adding rather than subtracting DST.

http://bytes.com/topic/c/answers/221484-obtaining-time-offset-utc

metaron-uk:
With an HTSP client in a different timezone, in order to use several of the htsp parameters which are specified as minutes after midnight server local time, you need to know what the server thinks midnight is in UTC. GetSysTime includes the 'timezone' paramater for this.
```
time               s64  required   UNIX time.
timezone           s32  required   Hours west of GMT.
```
 Unfortunately:

1. timezone offsets aren't always multiples of 1 hour (https://en.wikipedia.org/wiki/List_of_UTC_time_offsets)
2. vanilla glibc systems don't include DST offsets as part of tz_minuteswest (http://stackoverflow.com/questions/635780/why-does-glibc-timezone-global-not-agree-with-system-time-on-dst)

There is a bsd/linux extension to the tm structure (tm_gmtoff) which contains this information but it isn't always present, so a compile time check is added, with a fallback option which works unless the server is >=GMT+13 or <GMT-11 (Chatham and Samoa at certain times of the year).

The existing 'timezone' response to getSysTime is 'fixed' and a new reply 'gmtoffset' is added to allow offsets which aren't multiples of an hour to be catered for. The htsp protocol version has also been incremented. 
```
time               s64  required   UNIX time.
timezone           s32  required   Hours west of GMT.    (Not reliable before version 23)
gmtoffset          s32  required   Minutes east of GMT.  (Added in version 23)
```

If this change is accepted, the wiki (https://tvheadend.org/projects/tvheadend/wiki/Htsp) will need to be updated to introduce the new parameter and indicate that 'timezone' did not work reliably on vanilla glibc based servers prior to protocol 23.

If it is thought better to scrap 'timezone' completely, I'm happy to adjust and rebase.

The following test harness was used to develop this change:
```
#include <stdio.h>
#include <time.h>
#include <sys/time.h>
int main ()
{
  setenv("TZ", "Europe/London", 1); check_tz();
  setenv("TZ", "Europe/Berlin", 1); check_tz();
  setenv("TZ", "Australia/Adelaide", 1); check_tz();
  setenv("TZ", "US/Hawaii", 1); check_tz();
  setenv("TZ", "Canada/Newfoundland", 1); check_tz();
  setenv("TZ", "US/Samoa", 1); check_tz();
  setenv("TZ", "Pacific/Chatham", 1); check_tz();

}
int check_tz ()
{
  printf("TZ: %s:\n",getenv("TZ"));
  tzset();
  struct timeval tv;
  struct timezone tz;

  struct tm timeNow;
  struct tm gmtPlusOffsetTimeNow;
  time_t now = time(NULL);
  struct tm testTime = *localtime(&now);
  testTime.tm_year = 2016 - 1900;
  testTime.tm_mon = 0;
  testTime.tm_mday = 365; // Day of the year if month = 0
  testTime.tm_hour = 23;
  testTime.tm_min = 59;
//  now = mktime(&testTime);
//  now -= 6*31*24*60*60;
  localtime_r(&now, &timeNow);

  if (gettimeofday(&tv, &tz) == -1)
  {
    printf("gettimeofday failed\n");
    return -1;
  }

  time_t nowPlusOffset = now - tz.tz_minuteswest * 60;
  gmtime_r(&nowPlusOffset, &gmtPlusOffsetTimeNow);

  struct tm gmTimeNow;
  gmtime_r(&now, &gmTimeNow);
  int tz_offset = 0;
  tz_offset += (gmTimeNow.tm_hour - timeNow.tm_hour) * 60;
  tz_offset += gmTimeNow.tm_min - timeNow.tm_min;
  if (tz_offset > 11 * 60)
    tz_offset -= 24 * 60;
  if (tz_offset <= -13 * 60)
   tz_offset += 24 * 60;

  printf(" timezone: %d minutes west of GMT, ", tz.tz_minuteswest);
  if (timeNow.tm_isdst)
  {
    if (timeNow.tm_hour == gmtPlusOffsetTimeNow.tm_hour &&
        timeNow.tm_min  == gmtPlusOffsetTimeNow.tm_min)
      printf("DST already included\n");
    else
      printf("adjusted for tm_isdst: %d minutes west of GMT\n", tz.tz_minuteswest - timeNow.tm_isdst * 60);
  }
  else
   printf("DST Not Currently Active\n");
  printf(" gmtoff: %d minutes east of GMT or GMT%+d:%d\n", - timeNow.tm_gmtoff / 60, timeNow.tm_gmtoff / (60 * 60), timeNow.tm_gmtoff / 60 % 60);
  printf(" tz_offset: %d minutes east of GMT, or GMT%+d:%d\n", tz_offset, - tz_offset / 60, -tz_offset % 60);

  return 0;
}
```

metaron-uk:
It has been pointed out to me that gmtime_r and localtime_r don't always exist either, so I've re-written the commit to use localtime and gmtime instead. Hopefully this version will actually be portable :blush:

adamsutton:
@metaron-uk localtime() and gmtime() should be used with GREAT care and only when you're sure you know the entire code base from top to bottom. Misuse of these in a threaded application will cause some very difficult to track problems (trust me I've been bitten).

Better to write wrappers around them. Detect lack of support for necessary functions and use a safeguard (mutex etc..) in those cases.

metaron-uk:
@adamsutton thanks for the pointers. I can see time is a very murky subject. Not only complicated, but issues with code portability and full of 'features'. A couple of extra configure code snippets to use localtime_r() and gmtime_r() with fallback to localtime() / gmtime() and a mutex would be well worth it for code stability. Will have a go but it might take a day or two...

Q1: Before I start, what platforms does tvheadend need to support? would requiring localtime_r / gmtime_r support be acceptable?

Q2: I note you have a slightly different licence from most opensource projects I've come across. Does that mean pinching a code snippet from some other open source app which does this already would be a no-no?

adamsutton:
@metaron-uk 
1. Honestly I'm not sure what we're pitching at at the moment, officially we've only ever supported ubuntu/debian (and possibly rpi) though I think @perexg is a Redhat/Fedora? user so that's probably consider official now. Anything outside of that and the relevant maintainers can spot issues and provide patches (IMHO).

2. I wouldn't worry too much about the contributor agreement. It is actually pretty standard in larger projects, I mainly introduced it to try and give the project team more control. But nothing ever came of it. In general code snippets from another project are unlikely to be a big issue, if you're taking whole files / large chunks things get a bit murky (for all projects, most just ignore the issue until it's too late!).

But I would bare in mind that localtime_r() and gmtime_r() are already used fairly extensively in tvheadend. So no point adding a check/special for something that's already in widespread usage and causing no problems.

metaron-uk:
@adamsutton - In that case my previous commit is probably the best solution - uses localtime_r / and tm_gmtoff if supported or a 95% solution using  localtime_r / gmtime_r if not.
I've reverted this PR to that version. Let me know if there is anything else it needs.

ksooo:
@perexg ping

src/htsp_server.c
```diff
@@ -1083,13 +1083,34 @@ htsp_method_getSysTime(htsp_connection_t *htsp, htsmsg_t *in)
   htsmsg_t *out;
   struct timeval tv;
   struct timezone tz;
+  int tz_offset;
 
   if(gettimeofday(&tv, &tz) == -1)
     return htsp_error("Unable to get system time"); 
 
+  struct tm serverLocalTime;
+  tzset();
```
perexg: tzset should be removed from here...

src/htsp_server.c
```diff
@@ -1083,13 +1083,34 @@ htsp_method_getSysTime(htsp_connection_t *htsp, htsmsg_t *in)
   htsmsg_t *out;
   struct timeval tv;
   struct timezone tz;
+  int tz_offset;
 
   if(gettimeofday(&tv, &tz) == -1)
     return htsp_error("Unable to get system time"); 
 
+  struct tm serverLocalTime;
```
perexg: move this to declarations

src/htsp_server.c
metaron-uk: Sorry for taking so long to respond @perexg . I've been away for a week.
I included this due to several internet references that say tzset should be called before localtime_r for portability. (E.g. https://mm.icann.org/pipermail/tz/2008-January/014789.html)

src/htsp_server.c
metaron-uk: Will do.

src/htsp_server.c
perexg: Sorry, but my understanding is to call tzset only one time in app until the timezone (TZ variable) changes. It's called already in src/main.c and I don't see any benefit / requirement to call it more than one time in tvh.

metaron-uk: OK works for me. Squashed and rebased.

perexg:
Merged. Removed the HTSP version change - clients may detect the change using new field gmtoffset.

metaron-uk:
@perexg - thanks.
You should be aware that @ksooo specifically asked me to include a version change so the kodi pvr client would know when this issue had been fixed.
I agree that the presence of the new field gmtoffset is indicative and the other part could be considered a bugfix, but it will require special code to identify a compatible API.

ksooo:
Why no version bump? Without the bump I cannot deny an "incompatible" server during "hello". But I want to do exactly this - deny connection to a server if it is in another time zone than the client if the server does not support to expose it's timezone reliably.

perexg:
You may close the connection right after the getSysTime call when the gmtoffset field is missing in the reply from client. Basically, I don't like to change the protocol version after each pico-change. This change is not really incompatible with the previous implementation / usage.

ksooo:
Dont't like that, but yes, it will work.

ksooo:
I updated the htsp spec wiki page for this, btw.

PR #701, opened 2015-08-08
DVR: autorec directory name, fixes #3037

deu439:
I added a new parameter 'ondemand' to the cleanup_filename() function. If ondemand == 0 the function doesnt apply dvr_whitespace_in_title, dvr_clean_title and dvr_windows_compatible_filenames substitutions.
This allows me to disable this features for explicitly specified directory name in autorec entry.

'de->de_directory == NULL' determines whether the processed directory name was specified in autorec entry.

Fixes https://tvheadend.org/issues/3037.

### DIFF
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -172,9 +172,10 @@ dvr_rec_unsubscribe(dvr_entry_t *de)
 
 /**
  * Replace various chars with a dash
+ * - ondemand specifies if user demanded substitutions are performed
  */
 static char *
-cleanup_filename(dvr_config_t *cfg, char *s)
+cleanup_filename(dvr_config_t *cfg, char *s, int ondemand )
 {
   int len = strlen(s);
   char *s1, *p;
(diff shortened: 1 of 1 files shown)


src/dvr/dvr_rec.c
```diff
@@ -607,7 +611,7 @@ pvr_generate_filename(dvr_entry_t *de, const streaming_start_t *ss)
       break;
     *(dirsep - 1) = '\0';
     if (*x) {
-      s = cleanup_filename(cfg, x);
+      s = cleanup_filename(cfg, x, de->de_directory == NULL);
```
perexg: You cannot do 'de->de_directory == NULL' check. It affects also other directory components (like title/channel subdirectories).

There is no easy fix for this issue, because at this point, you don't know which part of the path is the de_directory.

deu439:
According to the help, the specifying de_directory overrides all other directory settings, so there isnt any other directory component.

src/dvr/dvr_rec.c
deu439: According to the help, specifying the de_directory overrides all other directory settings, so there isnt any other directory component.

perexg: But de_directory also accepts $ substitutions, too. You should remove these substitutions to make things work as you expect, too.

deu439: The second commit removes the substitutions.

perexg:
The problem is that I like the functionality for the explicit autorec directory with substitutions .

My implemenations is in : e4aa0911cd79659550e22425d12e1f61e16f9403

Your code was merged (with a little cleanups).

PR #703, opened 2015-08-27
Add systemd socket activation support

tsmetana:
Hello.

This patch enables the systemd socket activation. I've tested the code on my Fedora 22 and it seems to work OK. The functionality needs to be enabled explicitly with --enable-libsystemd_daemon configure option.
The new code checks when creating a TCP server whether systemd didn't pass the application a socket listening on the requested address/port and eventually reuses the given file descriptor.

Let me know what you think.

### DIFF
--- a/src/tcp.c
+++ b/src/tcp.c
@@ -41,6 +41,10 @@
 #include "access.h"
 #include "dvr/dvr.h"
 
+#if ENABLE_LIBSYSTEMD_DAEMON
+#include <systemd/sd-daemon.h>
+#endif
+
 int tcp_preferred_address_family = AF_INET;
 int tcp_server_running;
 th_pipe_t tcp_server_pipe;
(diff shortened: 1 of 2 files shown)


src/tcp.c
```diff
@@ -766,11 +778,90 @@ tcp_server_create
...
+  (const char *bindaddr, int port, tcp_server_ops_t *ops, void *opaque)
+{
+  int sd_fds_num, i, fd;
+  struct sockaddr_storage bound;
+  tcp_server_t *ts;
+  struct in_addr addr4;
+  struct in6_addr addr6;
+  int found = 0;
+
+  sd_fds_num = sd_listen_fds(0);
```
perexg: Uff.. I just wonder why these three lines and the bellow for loop are not in the systemd library / API. It looks terrible to implement this stuff for all required apps. Have you proposed to include this code to systemd ?

Something like: sd_listen_fds2(bindaddr, port) or so ?

perexg:
Merged. Thanks.

tsmetana:
You were fast... I'll try to consult the patch and the possibility of systemd API extension with the systemd guys and eventually come with another pull request.

The patch itself could be simplified if only the port would have been checked (not the address) with [sd_is_socket_inet (3)](http://www.freedesktop.org/software/systemd/man/sd_is_fifo.html) API function. I was aiming for bigger flexibility though and could not think out anything simpler.

Thank you.

tsmetana:
And for the record:
https://github.com/systemd/systemd/issues/1188

PR #706, opened 2015-08-31
HTSP: Added the "comment" and "episode" fields.

agitate:
I've added in the comment and episode fields to the dvr entries in HTSP responses.

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -732,10 +732,15 @@ htsp_build_dvrentry(dvr_entry_t *de, const char *method, const char *lang)
     htsmsg_add_str(out, "subtitle", s);
   if(de->de_desc && (s = lang_str_get(de->de_desc, lang)))
     htsmsg_add_str(out, "description", s);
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

Hedda:
Does it already have a "season" field? (also, was this PR merged? GitHub sais it was closed)

agitate:
No, but the season information is included in the episode field. Note that the format of the episode field stored in the dvrentry (which is a full text description, such as, "Season 2, Episode 3") does not match that returned in an event (which would be 2.3). Neither of these match the 0 based numbering found in the XMLTV format. It would be nice to consolidate them (and I'll probably look in to that next).

This change was indeed merged (commit ae8565a02f3857f0201ed0d78bce2ebe2d3a1ae1).

agitate:
Sorry, minor correction: dvrentries store it like this "Season 2.Episode 3" (with a dot between as opposed to a comma and a space).

PR #707, opened 2015-09-01
xmltv: handle timezone attached to time with no space padding

dhead666:
These changes needed for xmltv downloaded from a VBox Gateway device.

I added a check for alphabetic characters but not actually the parsing of timezone strings.

xmltv example attached
https://www.dropbox.com/s/i7sr7omjfn15rbx/vboxXmltv.xml?dl=0

### DIFF
--- a/src/epggrab/module/xmltv.c
+++ b/src/epggrab/module/xmltv.c
@@ -17,6 +17,7 @@
  */
 
 #include <assert.h>
+#include <ctype.h>
 #include <sys/stat.h>
 #include <errno.h>
 #include <string.h>
@@ -72,13 +73,15 @@ static time_t _xmltv_str2time(const char *in)
   str[sizeof(str)-1] = '\0';
 
   /* split tz */
-  while (str[sp] && str[sp] != ' ')
+  while (str[sp] && str[sp] != ' ' && str[sp] != '+' && str[sp] != '-')
+    sp++;
+  if (str[sp] == ' ')
     sp++;
 
   /* parse tz */
   // TODO: handle string TZ?
   if (str[sp]) {
-    sscanf(str+sp+1, "%d", &tz);
+    sscanf(str+sp, "%d", &tz);
     tz = (tz % 100) + (tz / 100) * 3600; // Convert from HHMM to seconds
     str[sp] = 0;
   }


perexg:
Timezone strings: Look to https://sourceware.org/bugzilla/show_bug.cgi?id=11620 , comment # 8. It seems like a complex job and using a library makes sense here.

dhead666:
Well I didn't intended in this PR to adding timezone string parsing ;)

It doesn't clear to me if the project at the link making available an interface for a timezone string conversion function, it look like not (but I'm a bit inexperienced so I might missed it), maybe it would be better to just offload the parsing of the whole str string to a external library (not sure about ISO 8601 compatibility)?

I'm guessing the other option is to add a dictionary / hash table to parse the strings.

Anyway ATM I'm only interest to get in the no space handling, so if it's appropriate and acceptable I'll just drop the `isalpha()` tests to avoid confusion.

perexg:
OK. The isalpha tests are extra, I agree and perhaps this construct is more readable:

<pre>
diff --git a/src/epggrab/module/xmltv.c b/src/epggrab/module/xmltv.c
index 458bfca..fa1d9b2 100644
--- a/src/epggrab/module/xmltv.c
+++ b/src/epggrab/module/xmltv.c
@@ -72,13 +72,15 @@ static time_t _xmltv_str2time(const char *in)
   str[sizeof(str)-1] = '\0';
 
   /* split tz */
-  while (str[sp] && str[sp] != ' ')
+  while (str[sp] && str[sp] != ' ' && str[sp] != '+' && str[sp] != '-')
+    sp++;
+  if (str[sp] == ' ')
     sp++;
 
   /* parse tz */
   // TODO: handle string TZ?
   if (str[sp]) {
-    sscanf(str+sp+1, "%d", &tz);
+    sscanf(str+sp, "%d", &tz);
     tz = (tz % 100) + (tz / 100) * 3600; // Convert from HHMM to seconds
     str[sp] = 0;
   }
</pre>

dhead666:
Done.

perexg:
Merged. Thanks.

dhead666:
Thanks, can this backported into 4.0.x?

PR #708, opened 2015-09-01
include in/out fields in subscription_create_msg()

Jalle19:
Without this potential clients will have to use the poll mechanism to get continuous updates on the bitrate of subscriptions.

@perexg I'd like to get this into the 4.0 branch as well, should I PR it or will you cherry-pick it?

### DIFF
--- a/src/subscriptions.c
+++ b/src/subscriptions.c
@@ -957,6 +957,7 @@ subscription_done(void)
 void subscription_add_bytes_in(th_subscription_t *s, size_t in)
 {
   atomic_add(&s->ths_bytes_in, in);
+  atomic_add_u64(&s->ths_total_bytes_in, in);
 }
 
 /**
@@ -965,6 +966,7 @@ void subscription_add_bytes_in(th_subscription_t *s, size_t in)
 void subscription_add_bytes_out(th_subscription_t *s, size_t out)
 {
   atomic_add(&s->ths_bytes_out, out);
+  atomic_add_u64(&s->ths_total_bytes_out, out);
 }
 
 /**
--- a/src/subscriptions.c
+++ b/src/subscriptions.c
@@ -457,11 +457,11 @@ subscription_input_direct(void *opauqe, streaming_message_t *sm)
     th_pkt_t *pkt = sm->sm_data;
     s->ths_total_err += pkt->pkt_err;
     if (pkt->pkt_payload)
-      s->ths_bytes_in += pkt->pkt_payload->pb_size;
+      subscription_add_bytes_in(s, pkt->pkt_payload->pb_size);
   } else if(sm->sm_type == SMT_MPEGTS) {
     pktbuf_t *pb = sm->sm_data;
     s->ths_total_err += pb->pb_err;
-    s->ths_bytes_in += pb->pb_size;
+    subscription_add_bytes_in(s, pb->pb_size);
   }
 
   /* Pass to output */
@@ -950,6 +950,22 @@ subscription_done(void)
 /* **************************************************************************
  * Subscription control
  * *************************************************************************/
+
+/**
+ * Update incoming byte count
+ */
+void subscription_add_bytes_in(th_subscription_t *s, size_t in)
+{
+  atomic_add(&s->ths_bytes_in, in);
+}
+
+/**
+ * Update outgoing byte count
+ */
+void subscription_add_bytes_out(th_subscription_t *s, size_t out)
+{
+  atomic_add(&s->ths_bytes_out, out);
+}
 
 /**
  * Change weight
--- a/src/subscriptions.c
+++ b/src/subscriptions.c
@@ -913,14 +913,14 @@ subscription_status_callback ( void *p )
   LIST_FOREACH(s, &subscriptions, ths_global_link) {
     /* Store the difference between total bytes from the last round */
     uint64_t in_prev = s->ths_total_bytes_in_prev;
-    uint64_t in_curr = s->ths_total_bytes_in;
+    uint64_t in_curr = atomic_add_u64(&s->ths_total_bytes_in, 0);
     uint64_t out_prev = s->ths_total_bytes_out_prev;
-    uint64_t out_curr = s->ths_total_bytes_out;
+    uint64_t out_curr = atomic_add_u64(&s->ths_total_bytes_out, 0);
 
-    atomic_exchange(&s->ths_bytes_in_avg, (in_curr - in_prev));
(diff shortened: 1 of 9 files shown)


Jalle19:
Now that the message sent from the callback is identical to the "basic" message, maybe we should drop the duplicate code in there? Let me know if you want me to do it.

perexg:
It's a wrong implementation. Note that atomic_exchange(x, 0) clear the statistics, so the concurrent statistics reading will be impossible.

I think that one way is to change usage and semantics for in/out variables - use absolute (not relative) values which will be added in subscription_create_msg() with a precise timestamp, so the client can compute the average bitrate itself.

The second possible solution might be to cache the last value for unforced (non-comet) GET requests. It might be easier to implement (update the cached values in th_subscription_t before subscription_status_callback() call in subscription_status_callback()).

Also the errors field should be moved to subscription_create_msg(), too.

Jalle19:
You're right, I noticed a similar discrepance with the inputs "bps" field.

The second approach sounds reasonable to me, I'll see if I can implement it.

Jalle19:
Actually the first approach kinda makes more sense - if the total bytes are reported then a client who polls more frequently than once per second can still calculate the bitrate.

Jalle19:
@perexg mind taking another look? I ended up implementing both ideas since IMO they're both useful. The `in` and `out` fields now contain the total bytes from the previous "round" (aka. the last second). In addition there are two new fields, `total_in` and `total_out` which tracks the byte count for the lifetime of the subscription.

I bumped the HTSP version to 23 since new fields have been added.

Here's an example of the new output of `/api/status/subscriptions`.

```
{
  "entries":[
    {
      "id":1,
      "start":1441143508,
      "errors":0,
      "state":"Running",
      "hostname":"192.168.76.1",
      "username":"admin",
      "title":"VLC/2.2.0 LibVLC/2.2.0",
      "channel":"Sjuan Stockholm",
      "service":"IPTV/Sjuan Stockholm/Sjuan Stockholm",
      "in":426008,
      "out":426008,
      "total_in":11160244,
      "total_out":11160244
    }
  ],
  "totalCount":1
}
```

Jalle19:
Now the big question is how any of this can be backported to 4.0 since we can't really bump the HTSP version there.

src/subscriptions.c
```diff
@@ -952,6 +954,24 @@ subscription_done(void)
  * *************************************************************************/
 
 /**
+ * Update incoming byte count
+ */
+void subscription_add_bytes_in(th_subscription_t *s, size_t in)
+{
+  atomic_add(&s->ths_bytes_in, in);
+  atomic_add_u64(&s->ths_total_bytes_in, in);
```
perexg: Unfortunately, this may cause a little synchronization issue (no lock for this update operation).

I would prefer to update only ths_total_bytes_in and ths_bytes_in should be the last read value, so the atomic_exchange calls should appear like:

<pre>
prev = s->ths_bytes_in;
s->ths_bytes_in_prev = atomic_exchange(&s->ths_bytes_in, s->ths_total_bytes_in) - prev;
</pre>

src/subscriptions.c
```diff
@@ -952,6 +954,24 @@ subscription_done(void)
...
+}
+
+/**
+ * Update outgoing byte count
+ */
+void subscription_add_bytes_out(th_subscription_t *s, size_t out)
+{
+  atomic_add(&s->ths_bytes_out, out);
+  atomic_add_u64(&s->ths_total_bytes_out, out);
```
perexg: Similar issue as for input...

src/subscriptions.c
Jalle19: I'm not really sure what you mean here. Aren't atomic operations synchronized by default? Plus in the snippet you wrote the `in` parameter isn't used at all.

perexg: Atomic operation is synchronized, but it can be used only for one variable. The "reader" might see updated ths_bytes_in but not ths_total_bytes_in - so the result for both in fields may be unsynchronized.

Look to my proposed code once more - I changed semantics of ths_bytes_in as "total bytes from last update" instead bytes from the last update operation.

Jalle19: `ths_bytes_in_prev` is/should only be updated by the callback since it has to be updated only once per second. I guess I just still don't understand your reasoning here. IMO it doesn't really matter if the total byte count differs slightly.

Jalle19: In your snippet it also seems like `ths_total_bytes_in` will never be updated?

perexg: OK. I think that my explanation was not understandable. I would keep only atomic_add_u64(&s->ths_total_bytes_in, in); in subscription_add_bytes_in() and change subscription_status_callback() as proposed - new exchange code.

Jalle19: I think I understand what you mean now, I have some code that almost works, but there is no `atomic_exchange` for `uint64_t` so I don't know how the exchange you suggested could work.

Jalle19:
@perexg I changed the logic a bit, what you suggested didn't really work though since the value from `atomic_exchange` minus the value it previously had will obviously always be zero. The new approach works too, but I don't see how it is any better than the old one.

src/subscriptions.c
```diff
@@ -906,14 +911,18 @@ subscription_status_callback ( void *p )
...
+    /* Store the difference between total bytes from the last round */
+    uint64_t in_prev = s->ths_total_bytes_in_prev;
+    uint64_t in_curr = s->ths_total_bytes_in;
+    uint64_t out_prev = s->ths_total_bytes_out_prev;
+    uint64_t out_curr = s->ths_total_bytes_out;
+
+    atomic_exchange(&s->ths_bytes_in_avg, (in_curr - in_prev));
+    atomic_exchange_u64(&s->ths_total_bytes_in_prev, s->ths_total_bytes_in);
+    atomic_exchange(&s->ths_bytes_out_avg, (out_curr - out_prev));
+    atomic_exchange_u64(&s->ths_total_bytes_out_prev, s->ths_total_bytes_out);
```
perexg: All four atomic operations here are not required. The access is already protected using "global_lock".

The only "missing" protection is above - the read operation should be protected (it's probably not an issue for x86, but for other CPUs, the 64-bit operations might be more complex):

<pre>
int64_t in_curr = atomic_add(&s->ths_total_bytes_in, 0);
</pre>

Jalle19: What about out_curr, does it need atomics too?

Jalle19:
@perexg better now? Is this what you had in mind?

perexg:
Merged. Thanks.

PR #710, opened 2015-09-03
socket activation: saner IPv6 address comparison

tsmetana:
This replaces the for loop with memcmp when comparing the IPv6 addresses. Same functionality, fewer lines, memcmp is most likely optimized for a particular architecture.


perexg:
Merged. Thanks.

PR #713, opened 2015-09-18
WEBUI: change size windows detales in DVR

PiterEL:


src/webui/static/app/ext.css
```diff
@@ -577,7 +577,7 @@
 
 .x-epg-prefix {
     display: inline-block;
-    width: 12em;
+    width: 20em;
```
perexg: This looks ugly.

perexg:
Merged only window size changes.

PR #716, opened 2015-09-23
webui: redir to parent instead of root

herrnst:
Fixes the webui when running in an Apache ProxyPass/ProxyPassReverse scenario, where e.g. Apache is used as SSL host and serves the TVHeadend webserver from port 9981 to a subdir of the SSL host (https://host/tvheadend <-> http://host:9981). Without the patch, the webui won't load due to the missing locale causing a JS error.

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -1717,7 +1717,7 @@ http_redir(http_connection_t *hc, const char *remain, void *opaque)
       if (lang) {
         snprintf(buf, sizeof(buf), "src/webui/static/intl/tvh.%s.js.gz", lang);
         if (!http_file_test(buf)) {
-          snprintf(buf, sizeof(buf), "/static/intl/tvh.%s.js.gz", lang);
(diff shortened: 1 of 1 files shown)


perexg:
Why proxy does  not rewrite the redirects, too ?

herrnst:
Regarding the Apache case, it (the non-rewritten HTTP 302) might be a config issue (for me, all worked fine until I upgraded to a recent GIT commit a few weeks ago - doing this upgrade stuff very seldom as the software generally runs very stable and does what it should), but with this change, TVH will run with all-relative files/links again, and it'll help users who are running proxies that are not capable of absolute rewriting at all.

perexg:
I believe it's fixed in a12ec36830c0ee8d34cd66c39b93577c903ebd16 .

herrnst:
Sorry for the late reply. It seems not fully fixed, tvh.js.gz is now fetched from the subdir, but the language file still is tried from the webroot, resulting in a blank page.

PR #718, opened 2015-09-27
4 code cleanup/improvement patches

oliv3r:
2 simple code clean patches just change some whitespace/macro's, no code changes took place

1 patch changes how default files are written, by removing the eXecute bit from files, but leaving them for directories.

1 patch should fix localhost from not working with getaddrinfo()

See each commit message for details.


perexg:
Thanks. Merged first and last patch. The middle two patches can be reduced to: 0d155566038b17fefc893d408e8dae199913827c . It's always expected that the path is a filename in these routines.

oliv3r:
Are you sure? Doesn't the directory creation routines walk this codepath as well?
You know the code much better then me so I'll take your word for it ;) I got triggered by the fact that the files and directories created during first launch in /etc/tvheadend all had the same permissions.

PR #720, opened 2015-09-29
libav: transcode in HEVC - prologue

lekma:
- setup HEVC ids for transcode
- allow HEVC stream in MPEGTS (afaik it is possible)
- close hevc bitstream filter on muxer destroy
- balk on SCT_NONE and SCT_UNKNOWN in codec enumeration

### DIFF
--- a/src/libav.c
+++ b/src/libav.c
@@ -81,6 +81,9 @@ streaming_component_type2codec_id(streaming_component_type_t type)
   case SCT_VP9:
     codec_id = AV_CODEC_ID_VP9;
     break;
+  case SCT_HEVC:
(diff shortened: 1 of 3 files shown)


lekma:
Unfortunately that is as far as I can go about tuning the default options without burning my cpu.
For reference I tested these settings on an old AMD Phenom II X4 955 processor with 4 Gb RAM.
It is still a net gain for me as with x264 scaled down from 576p to 480p I get about 900Mb-1Gb per one hour show. With x265 and these settings I now get 500Mb-600Mb files without scaling down.

perexg:
Merged. Thanks.

PR #721, opened 2015-09-30
Makefile(s): fix libmfx support

lekma:


lekma:
might be needed:
sudo apt-get install libva-dev autoconf libtool

L-S-D:
on Debian, libtool-bin is also needed
still not working for me, have to check if Intel Media SDK is installed correctly

lekma:
what is not working?
does it build?

L-S-D:
it is building fine, no problem with that, but I can't use QSV, see http://tvheadend.org/issues/3080

perexg:
Merged. Thanks.

PR #722, opened 2015-10-01
Added configure options and builddeps to support HEVC in the RPM Build

ncorrare:
I've added cmake as a Builder, required by libx265, and the configure option to enable libx265.


perexg:
Merged. Thanks.

PR #723, opened 2015-10-02
libav_muxer: copy over header for HEVC as well

lekma:
This allow streaming hevc, but breaks recording using the matroska libav muxer (h264 recording with the same muxer is also broken).
Maybe we should remove the container 'Matroska (mkv)/av-lib'?


perexg:
Merged. Thanks.

PR #733, opened 2015-10-20
Simplify an English translation of Spanish language

trujulu:
I think "Spanish" is a more acurrate translation for "Spanish; Castilian" language in the menu


perexg:
Merged. Thanks.

PR #734, opened 2015-10-21
ATSC EPG support

laurimyllari:
This PR is for https://tvheadend.org/issues/3190

These changes are enough to get somewhat working EPG data from ATSC broadcasts. Please see https://tvheadend.org/boards/12/topics/6573 for discussion and user reports.


perexg:
Thanks. I rebased and pushed your code to the master tree of tvheadend with some compilation and logic fixes (value compatibility). It's not tested yet. We can continue work in https://tvheadend.org/issues/3190 . Many thanks for your work.

PR #738, opened 2015-10-26
Implement max schedules limit for autorrecordings.

traxxxto:
Hello. I have implemented a max number of schedules for autorrecords. As it's name states, It limits the number of schedules an autorrecord can create in one sigle search. For example:

Suppose in EPG of one channel.
- C.S.I MIAMI from 22:30 to 23:30
- C.S.I NEW YORK from 23:30 to 0:30
- C.S.I LAS VEGAS from 0:30 to 1:30.

An autorrecord for recording "C.S.I. " but limited only to one scheduled record, plus 240 mins of margin at the end. This gives me the posibility of record my favorite series only for watching later and jump the comercials, but in one big file without interruptions. When the recording finishes a new search is triggered to schedule the remainings if there are any.

Another example: 

Record all the videoclips from "Beyonce, Taylor Swift and Red Hot Chilli Pepers" from a music channel, but keep the upcoming grid "clean", as this music channel puts in EPG every single videclip for the next 5 days and there were 20 schedules only in the first 3 days (real case ;) ).

Besides this new feature, I've found a bug in dvr_autorec_changed function. When a autorec change was triggered and there were disabled schedules for this autorec, all of them inverted their states. I'ts now fixed.

Of course, I'm a newbie in C++ and git, so this code can have lots of bugs.

Sorry for my english ;)

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -306,6 +306,8 @@ typedef struct dvr_autorec_entry {
   uint32_t dae_retention;
   uint32_t dae_removal;
   uint32_t dae_max_count;
+  uint32_t dae_max_sched_count;
+
 
   time_t dae_start_extra;
   time_t dae_stop_extra;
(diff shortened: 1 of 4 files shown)


perexg:
Merged (with some minor - coding style) tweaks. Thanks.

PR #740, opened 2015-10-28
[htsp] fix htsp not informed when dvr changes

Glenn-1990:
When editing a dvr entry through the webui, the htsp clients are not updated until a restart.

@perexg 
Can you check if this is right, although it's working? 
Also, do you know if updates are passed to the htsp clients when parameters are changing automatically (by epg), like start and stop time?


Glenn-1990:
Possible to backport to 4.0?

ksooo:
> Possible to backport to 4.0?

Yes, please backport this.

perexg:
Applied to both branches.

PR #742, opened 2015-10-30
WebUI: Add thetvdb search, icon buttons, DVR database search

ProfYaffle:
This PR:

* Adds a search link for events to [thetvdb.com](http://thetvdb.com) (c/f the existing [imdb.com](http://imdb.com) search).

* Extends the search function from EPG-only to the DVR entries (scheduled/completed/failed) as I can't see why you wouldn't want to look things up from there either.

* Removes a condition that existed in the EPG that meant you couldn't look something up that was already-recording or scheduled for recording (*if (!recording && !scheduled)*).

* Finally, the extra buttons make the dialog a bit cramped, so I've made them icon-only (still with tooltip, but no text) and widened the EPG 'Broadcast Info' dialog slightly. No change to the DVR dialog, as that doesn't have the same button frenzy.

**Note: the functions searchIMDB() and searchTheTVDB() are now replicated in epg.js and dvr.js. I don't know if they should really be moved to some master, shared "#include"-style file?**

**Updated EPG View:**

image

**Updated DVR View:**

image

**Icon credits**

**IMDB**: modified by me from [this original source](http://www.softicons.com/social-media-icons/social-media-icons-by-paul-robert-lloyd/imdb-icon). Attribution as per accompanying readme:

>Thanks for downloading these icons. If you decide to use them, please place the following attribution link:

> <http://paulrobertlloyd.com/2009/06/social_media_icons>

> Licenced under a Attribution-Share Alike 2.0 UK: England & Wales Licence. I claim no right of ownership to the company logos used in these icons.

> <http://creativecommons.org/licenses/by-sa/2.0/uk/>

**TheTVDB**: Site favicon from [here](thetvdb.com).

> This work is licensed under a Creative Commons Attribution-NonCommercial 4.0 International License.

> https://creativecommons.org/licenses/by-nc/4.0/


ProfYaffle:
Closing so I can clean up my local branches.

Hedda:
@ProfYaffle maybe off-topic but would it not be better to use The Movie Database (TMDb) as reference instead of TheTVDB.com and IMDb.com? Reason I ask is because TheMovieDB.org are have a more updated/versitile API, uses a Creative Commons license, plus provide metadata for both movies as well as TV shows/series/episodes from the one site?

https://www.themoviedb.org
https://www.themoviedb.org/documentation/api

ProfYaffle:
@Hedda I really don't have an opinion. I went with adding TheTVDB simply because we already had IMDb, so it balanced - and they're typically the ones I'm familiar with, so it didn't occur to me to use anything else. If we're not copying anything, and only referencing, I don't think the licence really matters.

In my experience, so much gets replicated anyway. Indeed, I just checked on TMDb and I can see several images that I originally uploaded to TheTVDB, so I suspect they all converge eventually.

It's easy to change if anyone does have strong views, though - it's just a logo and a search string...

PR #744, opened 2015-11-01
WebUI: Add Ctrl+A/select-all to DVR grids - fixes #3232

ProfYaffle:
As it says - this adds a keyboard shortcut to the DVR grids (upcoming, finished, failed) so you can select all events at the same time. 

Useful it e.g. you're clearing out a load of old recordings, or if you have many that have failed (perhaps because they were disabled?).


perexg:
Applied.

PR #745, opened 2015-11-01
WIP htsp additions

Glenn-1990:
Stille wip, but I can use some feedback @perexg @ksooo 

@perexg 
Is it somehow possible to add conflict management to active htsp subscriptions and recordings?

i.e. If a running recording is in conflict (no free tuner), would it be possible to send the subscription (recording or live) responsible for it to the htsp client?

User example on client:
1) Recording could not be started because of no free adapter, the user can open a conflict dialog with the following options:
-> do nothing
-> record and stop recording x
-> record and stop recording y
-> record and stop live stream on ipz

) Livestream get aborted by a starting recording x, a dialog will pop up with the following options:
-> Keep watching and abort recording x
-> keep watching and abort recording y
-> do nothing

Basically the only thing the client has to know are the conflicting recordings, livesteams and nothing if the dvr or subscription is running fine...
I suppose we have to check which service is mapped to our stalled (no free adapter) subscription, check on which used adapter we can subscribe this service and then list the running subscriptions used by all these adapter?

Any suggestion? It would be a nice feature ;-)

### DIFF
--- a/src/dvr/dvr_autorec.c
+++ b/src/dvr/dvr_autorec.c
@@ -273,51 +273,9 @@ dvr_autorec_create(const char *uuid, htsmsg_t *conf)
 
 
 dvr_autorec_entry_t*
-dvr_autorec_create_htsp(const char *dvr_config_name, const char *title, int fulltext,
-                            channel_t *ch, uint32_t enabled, int32_t start, int32_t start_window,
-                            uint32_t weekdays, time_t start_extra, time_t stop_extra,
-                            dvr_prio_t pri, int retention, int removal,
-                            int min_duration, int max_duration, dvr_autorec_dedup_t dup_detect,
-                            const char *owner, const char *creator, const char *comment, 
-                            const char *name, const char *directory)
+dvr_autorec_create_htsp(htsmsg_t *conf)
 {
   dvr_autorec_entry_t *dae;
-  htsmsg_t *conf, *days;
-
-  conf = htsmsg_create_map();
-  days = htsmsg_create_list();
-
-  htsmsg_add_u32(conf, "enabled",     enabled > 0 ? 1 : 0);
-  htsmsg_add_u32(conf, "retention",   retention);
-  htsmsg_add_u32(conf, "removal",     removal);
-  htsmsg_add_u32(conf, "pri",         pri);
-  htsmsg_add_u32(conf, "minduration", min_duration);
-  htsmsg_add_u32(conf, "maxduration", max_duration);
-  htsmsg_add_s64(conf, "start_extra", start_extra);
-  htsmsg_add_s64(conf, "stop_extra",  stop_extra);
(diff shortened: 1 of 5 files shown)


src/htsp_server.c
```diff
@@ -487,6 +487,152 @@ htsp_user_access_channel(htsp_connection_t *htsp, channel_t *ch)
...
+ * Converts htsp input to internal API
+ * @param htsp, current htsp connection
+ * @param in, the htsp message input from client
+ * @param autorec, true for autorecs, false for timerecs
+ * @param add, true for new instances, false for update calls
+ * @return the htsmsg_t config to be added or updated with idnode
+ */
+static htsmsg_t *
+serierec_convert(htsp_connection_t *htsp, htsmsg_t *in, channel_t *ch, int autorec, int add)
+{
```
Glenn-1990: @perexg A function like this?
This will reduce the amount of code.

perexg: I like it. It's something I thought about. Any pending changes?

src/dvr/dvr_autorec.c
```diff
@@ -330,6 +288,16 @@ dvr_autorec_create_htsp(const char *dvr_config_name, const char *title, int full
   return dae;
 }
 
+void
+dvr_autorec_update_htsp(dvr_autorec_entry_t *dae, htsmsg_t *conf)
+{
+  idnode_update(&dae->dae_id, conf);
```
Glenn-1990: @perexg do we need a lock here for idnode?

src/htsp_server.c
Glenn-1990: Yes I'm still thinking about the coflict stuff

Glenn-1990:
@perexg 
https://github.com/Glenn-1990/tvheadend/blob/htsp/src/htsp_server.c#L924
Here I want something like this:

        htsmsg_t *conflicts = htsmsg_create_list();
        FOREACH(ths_conflict, de->de_s->ths_conflicts)
        {
          htsmsg_t *map = htsmsg_create_map();
          htsmsg_add_u32(map, "id", ths_conflict->id); // conflict id
          htsmsg_add_u32(map, "type", ths_conflict->type); // livestream or dvr
          htsmsg_add_str(map, "name", ths_conflict->name; // recording x, livestream y
          htsmsg_add_msg(conflicts, NULL, map);
        }

src/dvr/dvr_autorec.c
perexg: It's protected by global_lock, too. So it's already locked.

perexg:
What if user should not know about the used tuners (permissions) ?

perexg:
I merged your code. Close this pull when we finish the discussion or we may move to the bug tracker (feature request).

Glenn-1990:
@perexg shouldn't we bump the htsp version and update the htsp wiki?
How can I edit the wiki?

PR #747, opened 2015-11-02
transcode: update to latest FFmpeg API

mlauss:
This patch updates the transcoder to the the latest ffmpeg/libav API, fixes
build against ffmpeg-2.9/git

Beware: upstream ffmpeg removed the long deprecated deinterlace module,
so deinterlacing is no longer done in tvheadend.

Signed-off-by: Manuel Lauss <manuel.lauss@gmail.com>

### DIFF
--- a/src/plumbing/transcoding.c
+++ b/src/plumbing/transcoding.c
@@ -24,6 +24,7 @@
 #include <libavutil/opt.h>
 #include <libavutil/audio_fifo.h>
 #include <libavutil/dict.h>
+#include <libavutil/imgutils.h>
 
 #if LIBAVUTIL_VERSION_MICRO >= 100 /* FFMPEG */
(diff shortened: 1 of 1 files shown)


CvH:
@mlauss deinterlace is somewhat important (if you stream to lowend devices)
any alternatives ?

mlauss:
Probably need to implement a way to hook the vf_yadif filter before scaling/encoding.
However I have no idea (yet) how to do that just now.
The old "avpicture_deinterlace()" algorithm is gone, and not coming back I assume,
that's why I removed it.

mlauss:
Looks like the whole transcode code should be overhauled to create a filter chain.
I'll look into it, but no promises :)

mlauss:
I've implemented the filter chain method instead, so this pull request is obsolete.

PR #749, opened 2015-11-03
include filters in static ffmpeg build, fixes #3261

mlauss:
enable the filters used by the transcoder in the static ffmpeg build as well


perexg:
Merged. Thanks.

PR #750, opened 2015-11-04
transcoding: add pixfmt constraints to ouput in filter chain

mlauss:
fixes color distortion with the intel qsv encoder as reported
by lsd on irc.


perexg:
Merged. Thanks.

PR #751, opened 2015-11-04
Fix compilation issues in Fedora 23

daniviga:
To compile Tvheadend on Fedora 23 with `--enable-libffmpeg_static` the compilation hardening should be disabled. Maybe it's not the best approach, but at least for now it works.

Hardening is now enabled by default in Fedora 23 (see https://fedoraproject.org/wiki/Changes/Harden_All_Packages)

Fixes https://tvheadend.org/issues/3252

### DIFF
--- a/rpm/tvheadend.spec.in
+++ b/rpm/tvheadend.spec.in
@@ -2,6 +2,10 @@
 %global commit @COMMIT@
 %global shortcommit %(c=%{commit}; echo ${c:0:7})
 
+#Allow static builds on Fedora >= 23
+#see https://fedoraproject.org/wiki/Changes/Harden_All_Packages
+%undefine _hardened_build
+
 Name:           tvheadend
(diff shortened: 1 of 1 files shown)


perexg:
Won't be better to handle -fPIE in Makefile.ffmpeg ? I don't like this solution.

daniviga:
Agreed. I'm not an expert in C source compilation, if someone can fix it in a better way I will be very happy. For now I've fixed it in this way because Fedora 23 is out and my home server was awaiting for tvheadend.

zuzzurro:
well, actually not according to the document you linked above:

 |Non-hardening|Hardening
--------------------|-------------|-----
Compile for static linking |(nothing)	| -fPIE |
Link executable 	|(nothing) 	|-pie -Wl,-z,now

perexg:
Fixed in 9237d88dc713bdc87a0c19a3800d643e0c40e47b .

PR #752, opened 2015-11-04
fix potential crash in constcw descrabler destructor

mlauss:
gcc's lto noted that free() is passed a pointer to a pointer
to memory.


perexg:
Merged. Thanks.

PR #753, opened 2015-11-04
fix compile error with ffmpeg-git due to type changes

mlauss:
...in struct AVCodecContext

The bit_rate member was extended to 64 bits, adjust the printf format
string accordingly. Works on both ffmpeg-2.8 and -git.


perexg:
Merged. Thanks.

PR #758, opened 2015-11-13
Provide network interface list for IPTV, factor out service auto-enable function

swegener:
I've got another patch for the auto-enable function that also adds it to dvb_pat_callback(), as the service is also created there, which would obsolete it in psi_parse_pmt().
I've got some data services that end up being auto-disabled, because the service_has_audio_or_video() check in psi_parse_pmt() fails for them, but the services are in PAT and the auto-disable message is a bit misleading in this case.

### DIFF
--- a/configure
+++ b/configure
@@ -251,6 +251,16 @@ if enabled_or_auto dvben50221; then
   fi
 fi
 
+check_cc_snippet ifnames '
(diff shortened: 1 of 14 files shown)


perexg:
Merged. Thanks.

PR #759, opened 2015-11-15
describe the transliteration of channel icons with two examples

HolgerJeromin:
This would make the transliteration easier to understand. 
My first thought was that the unsave chars are discarded and not replaced.

### DIFF
--- a/docs/html/en/config_misc.html
+++ b/docs/html/en/config_misc.html
@@ -62,7 +62,7 @@
 <tbody>
 <tr>
 <td><strong>%C</strong></td>
-<td>The transliterated channel name in ASCII (safe characters, no spaces etc.)</td>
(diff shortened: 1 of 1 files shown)


ProfYaffle:
@HolgerJeromin Many thanks for the PR, but we'd need to push this somewhere else... if you notice, the first line of the HTML is:

    

The reason for that is that the web interface help files are generated from the [documentation files](http://docs.tvheadend.org/) so they're consistent between the online documentation and the application web help.

I think the piece you're commenting on is actually here now:

http://docs.tvheadend.org/webui/config_general/#picons

... so I've gone ahead and made an equivalent change here for you. As we finalise documentation for the 4.2 release, it will find its way into the HTML help as well.

HolgerJeromin:
Thank you! I should have read carefully.

PR #760, opened 2015-11-22
[DVR] keep forever and keep until space needed

Glenn-1990:
I've made a list of predefined values for retention and removal (1 day, 3 days, ... , until space needed, forever). But all other integer values lower than (INT_MAX -1)  are still accepted.

File system checking and removal of the "until space needed" recordings runs periodically (15 secs) when there is an ongoing active recording.

For retention I have a "Until file removal" now as this seems more user friendly to me.
(otherwise the user needs to change both fields (retention and removal) If he want to keep the recording a bit longer)

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -240,7 +240,7 @@ api_dvr_entry_create_by_event
                                        e, 0, 0,
                                        perm->aa_username,
                                        perm->aa_representative,
-                                       NULL, DVR_PRIO_NORMAL, 0, 0, comment);
+                                       NULL, DVR_PRIO_NORMAL, DVR_RET_DVRCONFIG, DVR_RET_DVRCONFIG, comment);
         if (de)
           dvr_entry_save(de);
       }
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -54,6 +54,7 @@ typedef struct dvr_config {
   uint32_t dvr_extra_time_post;
   uint32_t dvr_update_window;
   int dvr_running;
+  uint32_t dvr_cleanup_threshold;
 
   muxer_config_t dvr_muxcnf;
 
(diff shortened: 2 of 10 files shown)


perexg:
I'd like to see u32 and overflow cleanups rather than this. 0 is perfect and value like 8760000 practically means "unlimited" (1000 years).

Glenn-1990:
@perexg then we do get '8760000' in the retention fields to the user, this is not very user friendly.
So if the user adds an autorec through the webui, he should add 8760000 in the retention field, or can we solve this in a different manner?

perexg:
The GUI allows preset lists (like for start after/stop before in autorec). So we can have here 'one day, one week, two weeks, one month, six months, one year, two years, ten years'.

Glenn-1990:
@perexg also added "until space needed" now and updated PR description

perexg:
Some more comment: The "cleanup" should be protected (to not do simultaneous operations - see deferred_unlink() in dvr_entry_delete(). If you call "cleanup" multiple times within a short time interval, some files might not be removed from the disk yet so more files can be removed than necessary.

Glenn-1990:
@perexg Removed the cleanup to the disk space stuff, I also added a threshold in MB for each dvr config. 
But now, I've a problem with locking, I think I need to hold the global lock when deleting a dvrentry, but doing that, the main thread freezes, also after releasing the lock. Not locking results in a crash with an assert in "dvr_entry_delete".. I'm a total noob with locking :p

perexg:
Merged and recoded (added also used space maintenance) . Thanks for your contribution.

The only thing I don't like is to call the vfsstat from the global_lock . It might be optimized later.

Glenn-1990:
@perexg 
Thanks for fixing the locking

I wonder what the benefit is of the used space maintenance option?
So if I have a disk af 100GB, only 2GB is used (with the default setting)?
That 2 GB can already be consumed by other non tvh files and the the recording gets deleted immediately after finish?

Let's say I set the used space to 50GB for my disk of 100GB:
- If the used size exceeds 50GB, tvh will cleanup to 50GB, even if that 50GB are used by other non tvh related files.
- This is actually the same as setting the free space to 50GB, so both settings do the same but in a different manner.. Or am I missing a point here?

PR #762, opened 2015-11-24
When deleteing a dvr entry, try to also delete any KODI EDL files.

carpa5:
Hi. I am new at this so please be patient. 
I have a job that runs after a recording has finished. It uses "Comskip" to create a KODI compatible EDL file. When I was deleting a recording the video file got removed, but the EDL file did not. This commit fixes that situation by deleting the associated EDL file. 

Richard

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -50,6 +50,7 @@ static void dvr_entry_start_recording(dvr_entry_t *de, int clone);
 static void dvr_timer_start_recording(void *aux);
 static void dvr_timer_stop_recording(void *aux);
 static int dvr_entry_rerecord(dvr_entry_t *de);
+static void dvr_delete_edl_file(const char *filename, const char *rdir);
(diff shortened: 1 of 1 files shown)


perexg:
I don't like this. It would be better to run a custom script when the entry was removed - like 'post-processor command'.

perexg:
But thanks for your contribution!

carpa5:
OK, if this is the wrong way, can you point me to more info on  'post-processor command' after submitting a delete on a dvr entry. I have just recently switched from mythtv as the backend and so this is all very new to me. 

Thanks

perexg:
I implemented post-remove command in 0fda5efee3e768259a6cbe9d4177209778c12714 . The usage is similar to post-processor command. See help for the post-processor command details.

carpa5:
Thanks for the quick fix. It is great that you are so receptive to requests. I am implementing the post-remove script now.

PR #763, opened 2015-11-25
JS: Changed various strings to sentence case and changed some titles/text

mpmc:
Pretty much as title really - @ProfYaffle Could you take a look please? :)

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -39,9 +39,9 @@ tvheadend.dvrDetails = function(uuid) {
         if (episode)
           content += '<div class="x-epg-title">' + episode + '</div>';
         if (start_real)
-          content += '<div class="x-epg-time"><div class="x-epg-prefix">' + _('Scheduled Start Time') + ':</div> ' + tvheadend.niceDate(start_real * 1000) + '</div>';
+          content += '<div class="x-epg-time"><div class="x-epg-prefix">' + _('Scheduled start time') + ':</div> ' + tvheadend.niceDate(start_real * 1000) + '</div>';
         if (stop_real)
-          content += '<div class="x-epg-time"><div class="x-epg-prefix">' + _('Scheduled Stop Time') + ':</div> ' + tvheadend.niceDate(stop_real * 1000) + '</div>';
+          content += '<div class="x-epg-time"><div class="x-epg-prefix">' + _('Scheduled stop time') + ':</div> ' + tvheadend.niceDate(stop_real * 1000) + '</div>';
         if (duration)
           content += '<div class="x-epg-time"><div class="x-epg-prefix">' + _('Duration') + ':</div> ' + parseInt(duration / 60) + ' ' + _('min') + '</div>';
         if (desc)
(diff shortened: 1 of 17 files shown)


src/webui/static/app/tvheadend.js
```diff
@@ -489,7 +489,7 @@ function accessUpdate(o) {
         var chepg = new Ext.TabPanel({
             activeTab: 0,
             autoScroll: true,
-            title: _('Channel / EPG'),
+            title: _('Channel list'),
             iconCls: 'television',
```
ProfYaffle: If this is the config tab, doesn't it still cover both channels and EPG settings? Or am I on an old version?

src/webui/static/app/epggrab.js
```diff
@@ -27,7 +27,7 @@ tvheadend.epggrab_base = function(panel, index) {
         labelWidth: 200,
         tbar: [triggerButton],
         help: function() {
-            new tvheadend.help(_('EPG Grab Configuration'), 'config_epggrab.html');
+            new tvheadend.help(_('EPG grabber configuration'), 'config_epggrab.html');
```
ProfYaffle: Is it the configuration of the grabber, of of grabbing in general? If the latter, then "EPG grab" is probably correct vs "EPG grabber" - or even "EPG source" if you want to avoid it entirely

src/webui/static/app/dvr.js
```diff
@@ -490,8 +498,8 @@ tvheadend.dvr_failed = function(panel, index) {
         url: 'api/dvr/entry',
         gridURL: 'api/dvr/entry/grid_failed',
         readonly: true,
-        titleS: _('Failed Recording'),
-        titleP: _('Failed Recordings'),
+        titleS: _('Failed recording'),
+        titleP: _('Failed recordings'),
         iconCls: 'exclamation',
```
ProfYaffle: Are they now "failed" or "missed" (or similar)? Disabled recordings appear here as well, but they're not strictly failures, are they...

src/webui/static/app/dvr.js
```diff
@@ -56,7 +56,7 @@ tvheadend.dvrDetails = function(uuid) {
         if (autorec_caption)
           content += '<div class="x-epg-meta"><div class="x-epg-prefix">' + _('Autorec') + ':</div> ' + autorec_caption + '</div>';
         if (timerec_caption)
-          content += '<div class="x-epg-meta"><div class="x-epg-prefix">' + _('Time Scheduler') + ':</div> ' + timerec_caption + '</div>';
+          content += '<div class="x-epg-meta"><div class="x-epg-prefix">' + _('Time scheduler') + ':</div> ' + timerec_caption + '</div>';
```
ProfYaffle: This is "time scheduler" as in "Which time schedule rule triggered this", isn't it... I wonder if there's a better wording, e.g. "Autorec rule:" and "Time schedule rule:"

src/webui/static/app/tvheadend.js
mpmc: The tab includes Channel tags, EPG grabber modules etc. So yes I think you're right. I just didn't think Channels / EPG was correct either.  I suppose my change makes even _LESS_ sense now.

ProfYaffle:
No real problems - as you say, it's self-explanatory. The only general points I'd raise would be that I'd probably keep tab titles as title case, as they're headings - so "Help for Timeshift Configuration", for example, or "DVB Inputs".

src/webui/static/app/dvr.js
mpmc: Suggestions on a better title? :) "Skipped/missed recordings" perhaps?

Or maybe "Recordings with issues"?.

mpmc:
Hmm, I think you may be right, I've looked at a build I just made with the changes and it doesn't look as nice.

I'll close this PR and have a think.

PR #764, opened 2015-11-27
ignore comments in m3u8-playlist

trondmm:
comments in m3u8-files are currently interpreted as a url for a video segment. This patch skips such comments


perexg:
Commit f6ac3c88c4e949baa0189d8de950660a743c0c2f should already handle this.

trondmm:
Ah. So it does :)

PR #765, opened 2015-11-30
Naming adjustment and uppercase fixes

diraimondo:
Looking at the web interface on tab "Digital Video Recording", the name "DVR Autorec Entries" doesn't match with the style of the other tab names. This is a personal proposal for a more coherent naming of this feature.


diraimondo:
In the meanwhile I pushed other string fix commitments

ProfYaffle:
I was wondering if "Automatic Schedules" is clear enough... it's better than "DVR Autorec Entries", though, I know, but my problem is that the timerec rules ("Time Schedules") are *also* automatic. So we have two tabs, "Automatic" and "Also Automatic" :smile:

The difference is really "stuff that matches things in the EPG" and "stuff that matches the time on the clock", I suppose: almost "EPG Event Rules" vs "Time Rules". I'm not happy with those terms either, though...

diraimondo:
At least we agree that the prefix "DVR" should go: it is redundant with the name of the main tab.
"EPG Event Rules" vs "Time Rules" looks nice too. Considering also a simplified "EPG Rules" vs "Time Rules".

ProfYaffle:
"EPG Rules" vs "Time Rules" sounds like a good way to go - similar length, and they tell you what they mean.

Is there anything else you'd ever have a rule for... just wondering if it scales... or does it matter...

+1 for this as it stands, though.

Re: capitalisation, @mpmc may want to take a look as he's been doing a chunk on the strings as well.

mpmc:
I'm not sure I like "Time rules" I'm not sure I like the usage of rules TBH.  I don't know it just doesn't sound right you don't hear people going around saying "Oh I'm just going to add <programme> to my EPG rule" do you? Why not something even simpler like planner (This is what Sky call their recordings list) and timer (or even timers)? You know when VHS was popular "Oh I must set a timer to record <programme>".

diraimondo:
@mpmc What about "EPG Schedules" / "Time Schedules"

mpmc:
We want something that sounds universally familiar really - Like what you would find on an STB hence why I said Planner & Timer as for most in the UK and maybe even the US would know what they mean.

Planner may be a better replacement for "Upcoming / Current recordings" though thinking about it as the auto rec entries end up there anyway.

Why not just say it as it is "EPG auto record" and "Timers"?

At the end of the day as long as it makes sense I don't suppose it matters.

perexg:
Note that 'autorec' is very related to 'autopilot' . Just my two cents. Personally, I like this naming.

mpmc:
@perexg I don't mind it either, I think it's more to do with differentiating between Autorec and the Timed recordings more than anything. Autorec and Timers is what I would prefer.

diraimondo:
@perexg I would propose "tardis" for the "time schedule"! :-)

We are almost stuck: it is a personal tasting question. 
I didn't like the original 'autorec' because because it has different style/length with respect to the rest of the DVR tabs. It is also difficult to translate in other languages.

I'm the last man arrived here, so I can't decide. 
@perexg in the meanwhile consider the idea to pull the fixes on uppercase strings. When the question will be close, you can fix the final strings in 2 secs.

ProfYaffle:
I think it was the "DVR" bit that first raised an eyebrow, as they're both DVR-relate, and in the Digital Video Recorder section, so it's unnecessary.

That leads to "Autorec Schedule" and "Timer Schedule", perhaps, or just "Autorec" and "Timers", "Autorec Events", "Timer Events" or anything that's more consistent.

Feels like we're in danger of bikeshedding now, though, I must confess...

mpmc:
Lets take a vote? :laughing: 

I vote for keeping Autorec (which could be easily changed for other languages anyway) but changing "Time Schedules" to "Timers".

diraimondo:
Given last commits, I can close this.

PR #768, opened 2015-12-04
[DVR] retention corrections after last PR

Glenn-1990:
- fixed possible overflow as we are now using INT_MAX instead of UINT_MAX
- moved "log retention" to expert level as it is set to "on file removal" by default
and the user is probably more interested how long the file stays on disk instead of the log.
- used space was 2GB by default, so when 2GB of the total disk space was used (almost nothing), dvr's where deleted, added a possibility to disable this "used size" setting
- fixed wrong logging: dvr_vfsmgr.c line 203
- fixed one dvr config was skipped: dvr_vfsmgr.c line 218, incrementing was already done on line 181
- cosmetics

@perexg Can you explain the "used storage size" feature a bit, because I don't exactly get what you try to achieve with this?

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -133,7 +133,7 @@ typedef enum {
   DVR_RET_1MONTH    = (30+1),
   DVR_RET_2MONTH    = (60+2),
   DVR_RET_3MONTH    = (90+2),
-  DVR_RET_6MONTH    = (180+2),
+  DVR_RET_6MONTH    = (180+3),
   DVR_RET_1YEAR     = (365+1),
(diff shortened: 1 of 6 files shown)


perexg:
Merged. Thanks.

Used threshold - yep, the implementation is wrong - the actual "used space" value should be calculated using a sum of all DVR entries placed in one VFS which may be time consuming.

Glenn-1990:
@perexg Maybe we can drop the used threshold stuff then?

Glenn-1990:
Is it possible to get write access to the wiki? the htsp specs are a bit outdated...

PR #769, opened 2015-12-04
Add some more Tooltips!

mpmc:
@ProfYaffle Could you take a look please? :cool:

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2006,10 +2006,10 @@ const idclass_t config_class = {
       .type   = PT_STR,
       .id     = "cors_origin",
       .name   = N_("HTTP CORS origin"),
-      .desc   = N_("HTTP Cross-origin resource sharing. This option is "
-                   "usually set when Tvheadend is behind a proxy. "
-                   "Enter a domain (or IP) to allow cross-domain "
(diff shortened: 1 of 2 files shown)


ProfYaffle:
Happy to look later, but immediate thought is that some of these seem very long - how do they appear on screen, have you tested at all? I'm just wondering if I'm reviewing for content or brevity...

mpmc:
@ProfYaffle Yes I have tested them & they look fine to me on screen they only appear when you mouse over an element. like this..
Example

Yes some can be long and some have been stolen from the help but it's not like they prevent you from using the element, I also suggested to @perexg to add an option to hide them for those that find them annoying but I think in the long run it'll help with people constantly asking "What does xyz do?".

*I'm just wondering if I'm reviewing for content or brevity* both really it's always good to have another set of eyes IMO.

I was also thinking that we could use these somehow to generate the help docs and where necessary (in the docs) add more information.

perexg:
Merged. Thanks. I also added the possibility to disable quick tips in the general config.

perexg:
BTW: Generation of the docs templates from the idnode data is not a bad idea.

PR #770, opened 2015-12-06
[htsp] add new dvr lifetime options to htsp api 24

Glenn-1990:
Otherwise we need a htsp version bump again and v24 is only a month old.
I guess this is ok for dev builds?


perexg:
Applied. Thanks.

PR #772, opened 2015-12-07
Continuing to add tooltips.

mpmc:
@perexg You may want to have a quick scan through some of these as I had to guess what they were. :laughing:


mpmc:
Anyway, here are a few more tooltips :cool:

perexg:
All merged.

PR #774, opened 2015-12-24
Only delete once count is greater than max_count.

scottharrisaus:
[dvr] When autorec has max count set only remove recordings after count exceeds max count. If max count is set to 5,  currently only 4 recordings are kept not 5 as expected.


perexg:
Thanks. Applied in effbf60ec3cf186c192bf028cf7e23c3ab6f04bb .

PR #775, opened 2015-12-26
Adding Australian LNB 10700

Ferni7:
Same change committed to master branch but would like in 4.0.x release so that I can use with kodi addon builds please (when next 4.0.9 release is done)...

https://github.com/tvheadend/tvheadend/pull/766


perexg:
Added in 38e10fc802c5d4937fe94e05f5dce8d6252a95c1 .

PR #777, opened 2015-12-26
[htsp] use s64 channelId for autorec/timerec add

Glenn-1990:
- update autorec/timerec uses a s64 channelid
- add autorec/timerec uses a u32 channelid

Fix this inconsistency and use s64 for both methods (htsp v25+)


perexg:
Merged. Thanks.

PR #778, opened 2016-01-03
bump htspmon reported HTSP version

Jalle19:
Servers may omit messages if they know the client is too old to support them. Since this tool is for monitoring what the server sends it makes sense to report the highest possible version number.


perexg:
Merged. Thanks.

PR #779, opened 2016-01-03
[descrambler] don't attempt to open IP addresses as files when the highest OSCam mode is used

Jalle19:
Without this patch the "pc-nodmx" mode can't be used, tvheadend will fail spectacularly with something like `Could not open 127.0.0.1 (No such file or directory). Do you have OSCam running`.

@manio what is this mode supposed to do? Judging by the log both this and "mode 5" seem to work the same (both report "protocol version 2").

### DIFF
--- a/src/descrambler/capmt.c
+++ b/src/descrambler/capmt.c
@@ -484,7 +484,9 @@ capmt_connect(capmt_t *capmt, int i)
   if (!capmt->capmt_running)
     return -1;
 
-  if (capmt->capmt_oscam == CAPMT_OSCAM_TCP || capmt->capmt_oscam == CAPMT_OSCAM_NET_PROTO) {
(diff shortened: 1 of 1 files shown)


manio:
This mode was introduced by @perexg. To be honest I'd love to get rid of the oscam modes from the capmt tab as it is starting to be a big mess, old modes don't even work, not mentioning the LD_PRELOAD hack. I created a PR for the very beginnig to clear this: https://github.com/tvheadend/tvheadend/pull/684. That is also why I introduced a network mode with some kind of protocol versioning, as it is flexible (can be used over a network as well as locally when used with 127.0.0.1), but @perexg created another one mode which is the same as network mode but made over a unix socket, so I think that @perexg can explain/review it better as this mode is his work and he has own vision for this...

perexg:
Merged. Thanks. The modes only differ in unix socket (named pipe) / IP socket communication. Nothing else.

PR #782, opened 2016-01-09
Channel-EPG mapping should ignore case

atamariya:
When channel is mapped with a grabber, it should ignore case for maximum automatic possible matches.


perexg:
Applied partly (only change for src/epggrab/channel.c). The channel_find_by_name fcn is not used for epggrab auto-mapping.

PR #784, opened 2016-01-11
Bugfix/coverity checks

adamsutton:
@perexg just a few minor fixes here and there. Not sure the httpc.c is correct, i.e. does it / should it hold global lock. Not sure why coverity didn't spot this one, it actually showed up in one of @mpmc 's valgrind traces.

I'm planning to do something about that code more generally, but struggling for time as usual.


perexg:
@adamsutton : Could you push this to master?

adamsutton:
I'll push sometime today along with #788

PR #785, opened 2016-01-11
[htsp] include subcriptionId in descrambeInfo messages

Jalle19:
Without this a client with two subscriptions can't differentiate between the message. This is in line with all other subscription-related messages.


perexg:
Merged, Thanks.

PR #788, opened 2016-01-15
Feature/build system

adamsutton:
Includes updates to the build scripts to cache staticlibs on bintray (and re-use if available).

A few strange issues occurred when getting this all to work on travis-ci.org so there are some slightly annoying fixes.

In theory once this is integrated the nightly builds (given how I've configured the system) should produce and upload ffmpeg and hdhomerun static libs so that others can re-use them in their own builds.

Some of the scripts could probably do with hardening, but I think there good enough.


perexg:
Definitely a step forward. Could you push this ?

adamsutton:
Yeah I'll push sometime today ready for builds tonight.

PR #790, opened 2016-01-16
[htsp] "epgMaxTime" should define a timespan instead of being static

Glenn-1990:
Shouldn't "epgMaxTime" define a window.
Currently, when setting "epgMaxTime" for example to 1 day in the future, you will run out of epg data within 1 day.
1 day should be 1 day ahead of the current time, right?

Also the event add/update procedure didn't take this setting into account.
Now the window will be updated on certain intervals.

@perexg maybe there is a better way to do this? Please review :-)

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -77,6 +77,8 @@ static void *htsp_server, *htsp_server_2;
 #define HTSP_ASYNC_AUX_AUTOREC 0x03
 #define HTSP_ASYNC_AUX_TIMEREC 0x04
 
+#define HTSP_ASYNC_EPG_INTERVAL 30
+
 #define HTSP_PRIV_MASK (ACCESS_HTSP_STREAMING)
 
 extern char *dvr_storage;
@@ -93,6 +95,7 @@ static struct htsp_connection_list htsp_connections;
 
 static void htsp_streaming_input(void *opaque, streaming_message_t *sm);
 const char * _htsp_get_subscription_status(int smcode);
+static void htsp_epg_window_cb(void *aux);
 
 /**
  *
@@ -140,6 +143,10 @@ typedef struct htsp_connection {
   char *htsp_username;
   char *htsp_clientname;
   char *htsp_language; // for async updates
+
+  int64_t  htsp_epg_window;      // only send async epg updates within this window (seconds)
+  int64_t  htsp_epg_lastupdate;  // last update time for async epg events
+  gtimer_t htsp_epg_timer;       // timer for async epg updates
 
   /**
    * Async mode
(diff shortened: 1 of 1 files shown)


adamsutton:
I should probably take a look at this. The whole async EPG stuff actually had a pretty big flaw which I just never got around to look at, that it wouldn't work if you selected a period that didn't include everything.

Maybe that's what you've fixed here, in which case that would be excellent!

Glenn-1990:
@adamsutton I'll provide you some background information.

Currently the default way to get epg events into kodi is synchronously, the async epg option is off by default. It might happen that epg events were deleted/added between the kodi epg update intervals (minimum 15 minutes). In this way, the epg window isn't always correct and wrong timers can be added.

The solution to this is switching to async mode, but it's very inefficient in 2 ways.
1) tvh: kodi will display a predefined window, but with async epg we have to parse all events, so also the ones far behind the point where kodi displays them, this also increases network load and  import times. To solve this we should set "epgMaxTime" to the maximum time of the kodi epg window, this doesn't work as desired as described above -> this PR solves this
2) kodi: when an async epg update was sent to pvr.hts, it will trigger an kodi update for that particular channel, so all epg data gets updated instead of a single event, this will also introduce unneeded system load. -> kodi v17 will support updates/deletes/adds for single events to solve this

So with 1) and 2) together we will have the "perfect" solution i.e. asynchronous epg will be more efficiently as synchronous epg updates. Currently it's the other way around.

Maybe helpful, this is what I used to test this PR:
https://github.com/Glenn-1990/pvr.hts/commit/4fea355d76ea60708f550810ea5367c379a05bab
This is not based on the new kodi methods as that part isn't merged yet into kodi. 
Basically the initial async epg import time is limited now by setting "epgmaxTime" to 5 hours, after initial import, it will be increased to the current kodi window.

adamsutton:
@Glenn-1990 what you're describing does sound very much like the limitation I was aware of but never got around to resolving. So if this does indeed fix that problem then that's excellent news, I will try and take a look when I find some time.

src/htsp_server.c
```diff
@@ -1423,14 +1441,20 @@ htsp_method_async(htsp_connection_t *htsp, htsmsg_t *in)
 
   /* Send EPG updates */
   if (epg) {
```
perexg: This code block appears almost same as htsp_epg_window_cb(). I would create a fcn with the shared code called from here and from htsp_epg_window_cb().

src/htsp_server.c
```diff
@@ -1367,10 +1374,21 @@ htsp_method_async(htsp_connection_t *htsp, htsmsg_t *in)
...
-  htsmsg_get_s64(in, "lastUpdate", &lastUpdate);
-  htsmsg_get_s64(in, "epgMaxTime", &epgMaxTime);
+  if (!htsmsg_get_s64(in, "lastUpdate", &lastUpdate))   // 0 = never
+    htsp->htsp_epg_lastupdate = lastUpdate;
+  if (!htsmsg_get_s64(in, "epgMaxTime", &epgMaxTime)) { // 0 = unlimited window
+    if (htsp->htsp_async_mode & HTSP_ASYNC_EPG) {
+      /* Only allow to increasing the window as pulling back events doesn't make sense */
+      if (htsp->htsp_epg_window && epgMaxTime > (dispatch_clock + htsp->htsp_epg_window)) {
+        htsp->htsp_epg_window = epgMaxTime-dispatch_clock;
+        gtimer_arm(&htsp->htsp_epg_timer, htsp_epg_window_cb, htsp, 1); // Force update
```
perexg: It's ineffective when epg == 1 (the next gtimer_arm will rearm timer with HTSP_ASYNC_EPG_INTERVAL). I'm not sure if it's intended. The simplest solution will be to set 'epg = 1' here and let the 'if (epg) {' code block in this fcn do this job.

perexg:
Merged and implemented my suggestions in 4785faf5e843867710718c0296596cdf5ba800c2 .

PR #794, opened 2016-02-01
Fixed bug-3507 - incorrect dereference of argv when performing variab…

drivera73:
Fixed bug-3507 - incorrect dereference of argv when performing variable interpolation for pipe:// URLs in IPTV.  This affects all versions and should be backported throughout.


perexg:
Thanks. Merged to both 4.0 and 4.1 trees..

PR #795, opened 2016-02-03
Makefile: move inline patch outside of makefile.

lekma:
- add a PATCH command to Makefile.static
- move patch to dedicated support/patches folder


perexg:
Applied. Thanks.

PR #796, opened 2016-02-03
Tooltips again..

mpmc:
@perexg 

Few more descriptions, I've added what I can and will continue to do so however there are a few options I don't understand fully , these are:-

linuxdvb_switch.c:118
iptv.c:707:820

I'll leave these for you to describe :)


perexg:
Applied. Thanks.

PR #797, opened 2016-02-04
Fix crash when selecting once per day in duplicate handling. fixes #3552

scottharrisaus:
Attempt at fixing crash (issue #3552) when using once per day in duplicate handling.

After applying fix autorec does not crash server and only 1 recording is scheduled to record for the day with the rest marked as duplicates of it.


perexg:
Applied. Thanks.

PR #800, opened 2016-02-08
Tooltips

mpmc:
Same as before :+1: 

Note: I've not finished with the wizard just yet, but I'm working on it!


perexg:
Merged all.

PR #803, opened 2016-02-18
libaesdec: fix incomplete packet handling

spdfrk:
A calculation error (offset / length) was made which only was a problem for incomplete packets. 
Also some slight optimization was possible.


perexg:
Merged. Thanks.

PR #805, opened 2016-02-22
Add service type override.

meijjaa:
Some providers don't set the service type correctly. This causes a problem when, for instance,
the client doesn't know if it's Radio or TV and chooses the wrong one.

This override will let the user set the service type.

### DIFF
--- a/src/service.c
+++ b/src/service.c
@@ -150,13 +150,11 @@ static htsmsg_t *
 service_type_auto_list ( void *o, const char *lang )
 {
   static const struct strtab tab[] = {
-    { N_("Override disabled"),  ST_UNSET },
-    { N_("None (0)"),           ST_NONE },
-    { N_("MPEG2 SD TV (0x01)"), 0x01 },
-    { N_("Radio       (0x02)"), 0x02 },
-    { N_("MPEG2 HD TV (0x11)"), 0x11 },
-    { N_("H.264 SD TV (0x16)"), 0x16 },
-    { N_("H.264 HD TV (0x19)"), 0x19 }
+    { N_("Override disabled"), ST_UNSET },
+    { N_("None"),              ST_NONE  },
+    { N_("Radio"),             ST_RADIO },
+    { N_("SD TV"),             ST_SDTV  },
+    { N_("HD TV)"),            ST_HDTV  }
   };
   return strtab2htsmsg(tab, 1, lang);
 }
--- a/src/service.c
+++ b/src/service.c
@@ -150,9 +150,12 @@ static htsmsg_t *
 service_type_auto_list ( void *o, const char *lang )
 {
   static const struct strtab tab[] = {
-    { N_("Override disabled"),  -1 },
-    { N_("Radio (0x02)"), 0x02 },
-    { N_("MPEG2 HD TV (0x11)"),  0x11 },
+    { N_("Override disabled"),  ST_UNSET },
+    { N_("None (0)"),           ST_NONE },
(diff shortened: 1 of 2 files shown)


src/service.c
```diff
@@ -1134,9 +1144,14 @@ service_has_audio_or_video(service_t *t)
 int
 service_is_sdtv(service_t *t)
 {
-  if (t->s_servicetype == ST_SDTV)
+  char s_type;
+  if(t->s_type_user > -1)
```
perexg: Add new ST_ define to service.h like ST_UNSET and use it here...

src/service.c
```diff
@@ -211,6 +211,15 @@ const idclass_t service_class = {
       .get      = service_class_caid_get,
       .opts     = PO_NOSAVE | PO_RDONLY | PO_HIDDEN | PO_EXPERT,
     },
+    {
+      .type     = PT_INT,
+      .id       = "s_type_user",
+      .name     = N_("Type override"),
+      .desc     = N_("Service type override. This value will override the "
+                     "service type provided by the stream."),
+      .off      = offsetof(service_t, s_type_user),
```
perexg: .list is missing - find 'service_class_auto_list' references for an example.

src/service.c
```diff
@@ -1148,9 +1163,14 @@ service_is_sdtv(service_t *t)
 int
 service_is_hdtv(service_t *t)
 {
-  if (t->s_servicetype == ST_HDTV)
+  char s_type;
+  if(t->s_type_user > -1)
```
perexg: Use new ST_ define here.

src/service.c
```diff
@@ -1166,9 +1186,14 @@ int
 service_is_radio(service_t *t)
 {
   int ret = 0;
-  if (t->s_servicetype == ST_RADIO)
+  char s_type;
+  if(t->s_type_user > -1)
```
perexg: Use new ST_ define here.

meijjaa:
Hi,

I updated the auto list but I'm not sure what you mean with a new ST_ define. Should I define a new service_t within these functions?

perexg:
<pre>
/**
   * Service type
   */
  enum {
    ST_UNSET = -1, /* ADD THIS */
    ST_NONE = 0,
    ST_OTHER,
    ST_SDTV,
    ST_HDTV,
    ST_RADIO
  } s_servicetype;
</pre>

meijjaa:
Thanks, I updated the enum and used it. I also extended the type list.

src/service.c
```diff
@@ -146,6 +146,21 @@ service_class_auto_list ( void *o, const char *lang )
...
+service_type_auto_list ( void *o, const char *lang )
+{
+  static const struct strtab tab[] = {
+    { N_("Override disabled"),  ST_UNSET },
+    { N_("None (0)"),           ST_NONE },
+    { N_("MPEG2 SD TV (0x01)"), 0x01 },
+    { N_("Radio       (0x02)"), 0x02 },
+    { N_("MPEG2 HD TV (0x11)"), 0x11 },
+    { N_("H.264 SD TV (0x16)"), 0x16 },
+    { N_("H.264 HD TV (0x19)"), 0x19 }
```
perexg: The above definitions are wrong. The type from DVB PSI tables is converted to ST_xxxx enum - see dvb_servicetype_lookup() in dvb_psi_lib.c , So use only ST_xxxx enums here.

meijjaa:
I see now, I updated the list to match the mappings.

perexg:

meijjaa:
Happy to do, keep up the good work!

PR #806, opened 2016-02-28
 RTSP: Use GET_PARAMETER in keep alive loop to prevent time out

spdfrk:
Use GET_PARAMETER in keep alive loop to prevent time out. Currently OPTIONS is used but the connection will still time out. With an empty GET_PARAMETER request the timer does reset.

### DIFF
--- a/src/http.h
+++ b/src/http.h
@@ -391,8 +391,14 @@ void http_client_unpause( http_client_t *hc );
  * RTSP helpers
  */
 
-int rtsp_send( http_client_t *hc, http_cmd_t cmd, const char *path,
-               const char *query, http_arg_list_t *hdr );
+int rtsp_send_ext( http_client_t *hc, http_cmd_t cmd, const char *path,
+               const char *query, http_arg_list_t *hdr, const char *body, size_t size );
+
+static inline int
+rtsp_send( http_client_t *hc, http_cmd_t cmd, const char *path,
+               const char *query, http_arg_list_t *hdr ) {
+  return rtsp_send_ext( hc, cmd, path, query, hdr, NULL, 0 );
+}
                       
 void rtsp_clear_session( http_client_t *hc );
 
@@ -421,7 +427,10 @@ rtsp_teardown( http_client_t *hc, const char *path, const char *query ) {
 }
 
 static inline int rtsp_get_parameter( http_client_t *hc, const char *parameter ) {
-  return rtsp_send(hc, RTSP_CMD_GET_PARAMETER, NULL, parameter, NULL);
+  http_arg_list_t hdr;
+  http_arg_init(&hdr);
+  http_arg_set(&hdr, "Content-Type", "text/parameters");
+  return rtsp_send_ext(hc, RTSP_CMD_GET_PARAMETER, NULL, NULL, &hdr, parameter, strlen(parameter));
 }
 
 int rtsp_describe_decode( http_client_t *hc );
--- a/src/http.h
+++ b/src/http.h
@@ -426,12 +426,7 @@ rtsp_teardown( http_client_t *hc, const char *path, const char *query ) {
   return rtsp_send(hc, RTSP_CMD_TEARDOWN, path, query, NULL);
 }
 
-static inline int rtsp_get_parameter( http_client_t *hc, const char *parameter ) {
-  http_arg_list_t hdr;
-  http_arg_init(&hdr);
-  http_arg_set(&hdr, "Content-Type", "text/parameters");
-  return rtsp_send_ext(hc, RTSP_CMD_GET_PARAMETER, NULL, NULL, &hdr, parameter, strlen(parameter));
-}
+int rtsp_get_parameter( http_client_t *hc, const char *parameter );
 
 int rtsp_describe_decode( http_client_t *hc );
 static inline int
--- a/src/http.h
+++ b/src/http.h
@@ -110,6 +110,7 @@ typedef enum http_cmd {
   RTSP_CMD_TEARDOWN,
   RTSP_CMD_PLAY,
   RTSP_CMD_PAUSE,
+  RTSP_CMD_GET_PARAMETER,
 } http_cmd_t;
 
 #define HTTP_CMD_OPTIONS RTSP_CMD_OPTIONS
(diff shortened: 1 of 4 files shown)


src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -53,7 +53,7 @@ iptv_rtsp_alive_cb ( void *aux )
   iptv_mux_t *im = aux;
   rtsp_priv_t *rp = im->im_data;
 
-  rtsp_send(rp->hc, RTSP_CMD_OPTIONS, rp->path, rp->query, NULL);
+  rtsp_get_parameter(rp->hc, "");
```
perexg: This is wrong for some RTSP servers. For example, SAT>IP specification defines only OPTIONS for the RTSP connection keep-alive signaling.

spdfrk:
This is annoying, it does not work for an Edgeware Orbit server. I guess there is no real standard for this? A bit more detailed, the STB will request the position parameter every second but if I do that than Tvheadend will lock up somewhere because of the unexpected body in the response. As a workaround an empty request every 30 seconds works as well.

Might I suggest we make the keep alive request type configurable? A suggestion where to pass this option would be very welcome so I can prepare a proposal.

perexg:
> This is annoying, it does not work for an Edgeware Orbit server. I guess there is no real standard for this? A bit more detailed, the STB will request the position parameter every second but if I do that than Tvheadend will lock up somewhere because of the unexpected body in the response. As a workaround an empty request every 30 seconds works as well.

It should be fixed. There's already hc_data_received / iptv_rtsp_data callback, so it should work.

'--trace httpc' and wireshark might give some hints.

> Might I suggest we make the keep alive request type configurable? A suggestion where to pass this option would be very welcome so I can prepare a proposal.

Perhaps, the client should try GET_PARAMETER and if the server returns an error, fallback to OPTIONS.

spdfrk:
Hi,

> Perhaps, the client should try GET_PARAMETER and if the server returns an error, fallback to OPTIONS.

Thanks, I added the fallback.

> It should be fixed. There's already hc_data_received / iptv_rtsp_data callback, so it should work.
> 
> '--trace httpc' and wireshark might give some hints.

Strange, the callback is properly called but I think there is still something going wrong in httpc, this is how the keep alive trace looks:
2016-03-02 21:21:03.000 [  TRACE]:httpc: 0001: sending RTSP/1.0 cmd
2016-03-02 21:21:03.004 [  TRACE]:httpc: 0001: received RTSP/1.0 answer (len = 150)
2016-03-02 21:21:03.004 [  ERROR]:iptv: unknown data 24 received for '...'
2016-03-02 21:21:32.000 [  TRACE]:httpc: 0001: sending RTSP/1.0 cmd
2016-03-02 21:21:32.004 [  TRACE]:httpc: 0001: received RTSP/1.0 answer (len = 150)
2016-03-02 21:21:32.004 [  ERROR]:iptv: unknown data 150 received for '...' 
2016-03-02 21:22:01.000 [  TRACE]:httpc: 0001: sending RTSP/1.0 cmd
2016-03-02 21:22:30.000 [  TRACE]:httpc: 0001: sending RTSP/1.0 cmd
2016-03-02 21:22:51.637 [  TRACE]:httpc: 0001: client flush -5

24 would be the actual body but the 2nd time the buffer contains both header and body. The 3th and 4th time the request never gets send (checked with Wireshark). Some additional debugging shows the 1st and 2nd time httpc returns 5 (HTTP_CON_SENT), after that 4 (HTTP_CON_SENDING).

In the next commit I will add the body support with an actual request for the position. It is an UNIX timestamp which I probably need later on to get the remote timeshifting (skipping) working.

spdfrk:
Hi,

I tried to do some debugging in httpc.c, what I noticed in http_client_run0:
- the first reply gets processes properly and http_client_data_received() is called at line 1070;
- the second reply gets no processing and http_client_data_received() is called at line 963, for some reason hc->hc_in_data == -1.

I can't seem to find where hc->hc_in_data is set to -1, maybe you have an idea what is going on?

spdfrk:
I isolated the problem a little further. hc->hc_rpos still contains the length of the previous message and hc->hc_in_data = -1, if I reset hc->hc_in_data and hc->hc_rpos manually it works flawless. But there probably is some different root cause for this?

```
diff --git a/src/httpc.c b/src/httpc.c
index 9882c83..8ae79bb 100644
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -899,6 +899,8 @@ http_client_run0( http_client_t *hc )
   ssize_t r;
   size_t len;
 
+  hc->hc_in_data = 0;
+  hc->hc_rpos = 0;
   if (hc->hc_shutdown) {
     if (hc->hc_ssl && hc->hc_ssl->shutdown) {
       r = http_client_ssl_shutdown(hc);
```

perexg:
Could you try this patch?

<pre>
iff --git a/src/httpc.c b/src/httpc.c
index 9882c83..46b4d2d 100644
--- a/src/httpc.c
+++ b/src/httpc.c
@@ -1070,8 +1070,10 @@ header:
   res = http_client_data_received(hc, hc->hc_rbuf + hc->hc_hsize, len, 1);
   if (res < 0)
     return http_client_flush(hc, res);
-  if (res > 0)
+  if (res > 0) {
+    hc->hc_in_data = 0;
     return http_client_finish(hc);
+  }
   goto retry;
 
 rtsp_data:
</pre>

spdfrk:
I tried but it didn't work, if I also set hc_rpos back to 0 it works.

At line 1054 hc_rpos is already set to 0 but somehow at line 1070 in http_client_data_received hc_rpos is set back to the body length. 
If I bypass  http_client_data_received (res = 1) then it works without any additional change. So I guess  http_client_data_received thinks the message isn't completed yet?

perexg:
Fixed in 5a94400a0f7a17128a8b7244e6f3c9c222d99151 .

spdfrk:
That fixed it, thanks. Is it possible to merge this pull request now?

src/http.h
```diff
@@ -409,10 +417,22 @@ rtsp_play( http_client_t *hc, const char *path, const char *query ) {
...
+static inline int
 rtsp_teardown( http_client_t *hc, const char *path, const char *query ) {
   return rtsp_send(hc, RTSP_CMD_TEARDOWN, path, query, NULL);
 }
 
+static inline int rtsp_get_parameter( http_client_t *hc, const char *parameter ) {
+  http_arg_list_t hdr;
+  http_arg_init(&hdr);
+  http_arg_set(&hdr, "Content-Type", "text/parameters");
+  return rtsp_send_ext(hc, RTSP_CMD_GET_PARAMETER, NULL, NULL, &hdr, parameter, strlen(parameter));
```
perexg: Too much code for the inline function. Please, move the body to rtsp.c .

src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -78,6 +82,13 @@ iptv_rtsp_header ( http_client_t *hc )
     return 0;
   }
 
+  if (hc->hc_cmd == RTSP_CMD_GET_PARAMETER && hc->hc_code != HTTP_STATUS_OK) {
+    tvherror("iptv", "GET_PARAMETER command returned invalid error code %d for '%s', "
```
perexg: I would use 'tvhtrace' here.. It's not a fatal error.

spdfrk:
Sounds good, I included your remarks.

perexg:

PR #807, opened 2016-02-28
RTSP: Also allow RTP/AVP/UDP stream type.

spdfrk:
It is possible that the server appends UDP to the type so just check for RTP/AVP*.


perexg:
Merged. Thanks.

PR #808, opened 2016-02-28
Add remote timeshift support for RTSP sources

spdfrk:
IPTV providers provide RTSP servers for timeshifting and cloud recoding. This pull request creates an option to pass timeshift commands to the remote RTSP server.

Already added:
Headers set the position in the stream buffer (UNIX timestamp (start-end)) and playback speed (-16 - 16).
Option for service to pass through commands.
Pause and get_parameter (to get the current position) command from pull #806.

Still missing is the integration into the TvHeadend timeshift system. I'm not sure at which point a intersection can be created to send the command to the timeshift or rtsp client. Also the local timeshift buffer should be disabled for this service.

### DIFF
--- a/src/http.h
+++ b/src/http.h
@@ -346,6 +346,8 @@ struct http_client {
   char        *hc_rtsp_user;
   char        *hc_rtsp_pass;
   char         hc_rtsp_keep_alive_cmd;
+  int          hc_rtsp_range_start;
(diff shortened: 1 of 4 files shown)


spdfrk:
Superseded by https://github.com/tvheadend/tvheadend/pull/1347

PR #809, opened 2016-02-28
Fixed some typos (apostrophes for accents where appropriate)

NicApicella:
A bunch of accents used as apostrophes haven been replaced by the proper mark.

Of course, this hasn't been done blindly: I checked and modified only the marks in texts; backticks and code-related instances have been left as-is.

Also, I only edited the .pot-file (re: internationalisation)—I'm pretty sure that the .po-files will get updated (and thus the changes will be propagated) automagically.


perexg:
Merged. Thanks.

mpmc:
@NicApicella These grave accent goofs are entirely my fault @ProfYaffle did point them out and I corrected most of them in a different commit but it looks like I missed some! 

Thanks for pointing them out :)

ProfYaffle:
I presumed a case of "damn you, auto-correct!"

PR #811, opened 2016-03-02
config: Exclude timeshift buffer from configuration backup

mamarley:
This fixes the issue where if TVHeadend was streaming live TV with
timeshift turned on and suffered an unclean shutdown followed
immediately by an upgrade, it would try to put the entire timeshift
buffer into the configuration backup.


perexg:
Merged. Thanks.

PR #814, opened 2016-03-05
fix time_t overflow on a 32-bit OS

Glenn-1990:
https://github.com/kodi-pvr/pvr.hts/issues/201

Don't actually know how overflows should be detected, but this should do the trick.

@perexg


perexg:
Could you review 055be40c0c3dffae117f1be2ce22885eb819f02c ?

PR #818, opened 2016-03-07
Small styles fix DVR info to looks better

pepeEL:


perexg:
Merged with little changes.

PR #819, opened 2016-03-07
Small styles fix EPG detailes to looks better

pepeEL:


perexg:
Merged with little changes.

PR #822, opened 2016-03-12
Don't request all tiers on every key request from server

Ferni7:
Change provided here:
https://tvheadend.org/issues/3458#change-17132


perexg:
Added.

Ferni7:
Was this closed without merging on purpose?

PR #823, opened 2016-03-16
Small style correct show EPG info

pepeEL:


perexg:
Some subjective comments:

* colors - no, no...
* italic / normal font - I don't have a strong meaning...

pepeEL:
Hi
Why colors no ? Ok i can remove but i think that looks better because looks better when program start and finish.

And normal font looks better... Italic is no sync with other...

ProfYaffle:
I'm also no fan of colours... we should use colored icons in the dialog if necessary before we re-paint the text IMO.

Re: italics, I think that's just to separate the program text (description) from the "metadata". If it's not italicised, I'd suggest that it needs to be clearly separated in some other way.

pepeEL:
But it is separate by line <hr>. For me italic no looks good on looks all.

perexg:
In master (another way).

PR #826, opened 2016-03-16
libav: try to tidy up the mess that is Makefile.ffmpeg

lekma:
- fix broken libx265 build
- rename libffmpeg_static to ffmpeg_static
- rename libmfx option to qsv (which is what one really wants)
- update to ffmpeg 3.0
- drop support for non-ffmpeg libav libs (kinda needed for libswresample, but might force users
  to upgrade)
- drop vdpau config option (unused)
- add multiple options (libvpx, libvorbis, ...) to customize builds to one's needs
- add libfdk-aac support
- add libtheora support
- try to cleanup the libav section in Makefile
- libx264 now supports swscale and avfilter
- patch libmfx to support the correct path to libmfx(s|h)w64.so on ubuntu (other distro users
  might not be pleased with that one)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,62 +62,83 @@ CFLAGS  += -Wno-unused-value -Wno-tautological-constant-out-of-range-compare
 CFLAGS  += -Wno-parentheses-equality -Wno-incompatible-pointer-types
 endif
 
-ifeq ($(CONFIG_LIBFFMPEG_STATIC),yes)
 
-CFLAGS  += -I${BUILDDIR}/ffmpeg/build/ffmpeg/include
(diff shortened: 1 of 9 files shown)


lekma:
@perexg I'm not absolutely sure you need (or even want) to merge that one, but there are some good bits in here you might be interested in.

CvH:
Just out of curiosity, no opus support (aka kind of the best audio codec out there + oss)?

lekma:
@CvH if it makes sense in a tvheadend context I can add it. It will only be build support though (for the time being).

perexg:
Merged. Many thanks.

PR #828, opened 2016-03-22
Support Android 5.1

BenEfrati:
Fix some compilation error when compiling with android-21.


src/tvheadend.h
```diff
@@ -44,6 +44,9 @@
       (char *) memcpy(__new, __old, __len);                                   \
     })
 #endif
+#ifndef index
+#define index(...) strchr(__VA_ARGS__)
```
perexg: I would move this to src/compat.h including strdupa() .

perexg:
Merged all patches into one and pushed to master. Thanks.

PR #829, opened 2016-03-24
webui: Fix display of signal strength if scale differs from SNR scale

herrnst:
Typo or C/P probably, resulting in wrong display of the signal strength column. If the SNR property reports FE_SCALE_DECIBEL but the strength property reports FE_SCALE_RELATIVE, strength is shown as dBm instead of a percentage bar.


perexg:
Merged. Thanks.

PR #832, opened 2016-03-28
transcoding: disable swscale and avfilter support in libx264

lekma:
- it seems the way I implemented it broke parallel building :( (might be worth investigating why later)


perexg:
Merged. Thanks.

PR #835, opened 2016-03-31
Fix epggrab to close accepted socket after grabbing

axfree:
This patch fixes #3678
https://tvheadend.org/issues/3678


perexg:
Merged. Thanks.

PR #836, opened 2016-03-31
Do not crash on retention overflow.

Glenn-1990:
When a user enters for example '10000' in the dvr retention/removal field (on a 32-bit machine), tvheadend will crash due to an overflow when the recording finishes. 
With the abort() removed, an overflow would be handled as forever.

Also there was a small typo.

@perexg


perexg:
Merged.

PR #837, opened 2016-04-02
Markdown help additions and changes.

mpmc:
@perexg: Please don't commit these unless they're correct if not please point out any mistakes :smile:. 

I think I may be doing this backwards.. Should I be adding docs to docs/class/<title>.md?


mpmc:
OK I've figured it out, I am doing it backwards!

md_to_c.py does generate the docs_inc.c file!

Closing for now :)

PR #838, opened 2016-04-02
ffmpeg: revert minimum required version numbers

bkuhls:
This commit reverts the minimum version numbers required by configure to
the values before
https://github.com/tvheadend/tvheadend/commit/1359effe28a0381b8c9cbd362d6e144fb87b00fc#diff-e2d5a00791bce9a01f99bc6fd613a39dL486
in order to allow compilation with older versions of ffmpeg again.

Please note that the previous version numbers of all ffmpeg libs, with
the exception of libavfilter, are those of ffmpeg 2.0.7:
https://ffmpeg.org/olddownload.html

Therefore I synced the minimum version number required for libavfilter to
3.79.101 and not to 4.0.0 as previous in tvheadend's configure script.

Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>

### DIFF
--- a/configure
+++ b/configure
@@ -512,13 +512,13 @@ else
   if enabled_or_auto libav; then
     has_libav=true
 
-    check_pkg libavfilter   ">=6.31.100"  || has_libav=false
-    check_pkg libswresample ">=2.0.101"   || has_libav=false
-    check_pkg libavresample ">=3.0.0"     || has_libav=false
-    check_pkg libswscale    ">=4.0.100"   || has_libav=false
-    check_pkg libavformat   ">=57.25.100" || has_libav=false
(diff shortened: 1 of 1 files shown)


lekma:
did you test transcoding with these specific old libs?

bkuhls:
@lekma: No, until two weeks ago I used ffmpeg 2.8.6 but I reverted to the version numbers in use before so I guess they were ok. Otoh we could change the required version to ffmpeg 2.8.6 as present in Debian stretch.

bkuhls:
Closing this PR after buildroot bumped ffmpeg to 3.1, this commit is not needed anymore from my POV.

PR #841, opened 2016-04-03
Enabling multiple channel names and adding a channel number to the XMLTV EPG grabber

axfree:
This patch enables xmltv epg to have a channel number and mutiple channel names.
http://wiki.xmltv.org/index.php/XMLTVFormat

Example:
```
<display-name>21 JTBC Sports</display-name>
=> channel number: 21
=> channel name: "JTBC Sports"

<display-name>21</display-name>
=> channel number: 21

<display-name>JTBC Sports</display-name>
=> channel name: "JTBC Sports"
```


perexg:
Merged. Thanks.

agitate:
My XMLTV has a display-name like this:

`<display-name lang="en">5 USA</display-name>`

but it isn't channel number 5 (it is channel 21). I also have these in my XMLTV:

`<display-name lang="en">102.2 Smooth Radio</display-name>
<display-name lang="en">5*</display-name>
<display-name lang="en">102.2 Smooth Radio</display-name>
<display-name lang="en">96-106 Capital FM</display-name>'

From what I can see, from the XMLTV spec, you can have multiple display-name elements, but the content doesn't seem to be interpreted in any way other than as a string to represent the display-name. What do you think?

axfree:
These are not treated as channel number. only numbers followed by space is treated as such.
```
<display-name lang="en">102.2 Smooth Radio</display-name>
<display-name lang="en">5*</display-name>
<display-name lang="en">96-106 Capital FM</display-name>
```
But this one is not what I expected :)
`<display-name lang="en">5 USA</display-name>
`

To handle that case, I've decided not to extract channel number if it's uncertain.
Channel number extraction is now done only if there's a channel number only `<display-name>` exists like this.
```
<display-name lang="en">5 USA</display-name>
<display-name lang="en">5</display-name>
```

This is how I done. code is little messy, but works.
https://github.com/axfree/tvheadend/commit/af78b3255f439e17c9833dfce78374b0ba2567b9

PR #842, opened 2016-04-05
[xmltv] fixed not to extract channel number from display name if it's uncertain

axfree:
https://tvheadend.org/issues/3692

See also:
https://github.com/tvheadend/tvheadend/pull/841

Thanks.


perexg:
I don't like this. I made this extraction configurable in 929517e8f1b9bad347119d7640ac59aa9c74e49e and also added possibility to extract numbers only when display-name tag contains only digits in 3b0db5767c3e18c1bb12b136c421d6fef1f5a14d .

axfree:
I didn't noticed that change. Having an option is a good idea. I second that 👍

PR #848, opened 2016-04-09
More changes & additions to mdhelp plus a small 'fix' for the webui

mpmc:
As title..
Not sure you'll agree with the messagebox fix but it was starting to bother me a little ;)


perexg:
Merged. Thanks.

PR #849, opened 2016-04-10
webui: change property->get_opts() signature to accept a property ptr argument

lekma:
this would allow us to extend property options.
ex.:
```
uint32_t
my_get_opts(void *obj, const property_t *prop)
{
    uint32_t opts = prop->opts;

    if (obj && !(((mystruct *)obj)->enabled)) {
        opts |= PO_RDONLY;
    }
    return opts;
}
```
@perexg: please do not merge if you feel/know there is a better way to achieve the same result.


lekma:
After thinking about it, I think there is something better to do here.
Closing.

PR #850, opened 2016-04-10
input: tvhdhomerun: Handle the device changing IP addresses

mamarley:
Previously the device detection loop would always ignore HDHomeRuns
with the same UID as devices that had already been detected.  This
would not detect if an HDHomeRun changed IPs for any reason (e.g.
APIPA, misconfigured DHCP server) while TVHeadend was running,
requiring a restart to resume functionality.  Instead, if
a newly detected device has the same UID as an previously-detected
one but a different IP address, destroy the old device and recreate
it with the new IP address.

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
@@ -21,6 +21,7 @@
 #include "input.h"
 #include "htsbuf.h"
 #include "settings.h"
+#include "tcp.h"
 #include "tvhdhomerun.h"
 #include "tvhdhomerun_private.h"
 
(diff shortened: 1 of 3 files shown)


src/input/mpegts/tvhdhomerun/tvhdhomerun.c
```diff
@@ -86,7 +86,7 @@ tvhdhomerun_device_class_get_title( idnode_t *in, const char *lang )
   static char buf[256];
   tvhdhomerun_device_t *hd = (tvhdhomerun_device_t *)in;
   snprintf(buf, sizeof(buf),
-           "%s - %s", hd->hd_info.friendlyname, hd->hd_info.ip_address);
+           "%s - %s", hd->hd_info.friendlyname, inet_ntoa(hd->hd_info.ip_address));
```
perexg: Pls, use tvh's tcp_get_str_from_ip() function here... inet_ntoa might not allow recursive calls (static return buffer).

src/input/mpegts/tvhdhomerun/tvhdhomerun_private.h
```diff
@@ -35,7 +35,7 @@ static struct hdhomerun_debug_t* hdhomerun_debug_obj = 0;
 
 struct tvhdhomerun_device_info
 {
-  char *ip_address;         /* IP address */
+  struct in_addr ip_address;         /* IP address */
```
perexg: Use complete 'struct sockaddr' here..

perexg:
Re to IRC: See - http://stackoverflow.com/questions/16010622/reasoning-behind-c-sockets-sockaddr-and-sockaddr-storage . You may retype between struct sockaddr_storage , sockaddr , struct sockaddr_in and struct sockaddr_in6 (according family value).

If you like to compare IPv4 - use sin_addr member from sockaddr_in .

perexg:
Merged. Thanks for this contribution.

PR #852, opened 2016-04-13
mdhelp

mpmc:
Some more changes!

I decided that it's better to have the upcoming, finished and failed tab markdown in one file as they're near enough the same and they use the same property class/items. Not sure what you guys will make of it but it can always be reverted if you don't agree. :)


perexg:
Merged.

PR #853, opened 2016-04-13
settings: minor code cleanup

oliv3r:
Fixes a whitespace issue and puts a variable declaration at the start
of its scope rather then mid-scope which can improve portability.


perexg:
Merged.

PR #856, opened 2016-04-19
Mdhelp

mpmc:
More help docs :smile: 

@perexg You may want to check this one over before committing as I think I've missed some classes / made mistakes etc.


perexg:
Merged. Thanks.

PR #859, opened 2016-05-05
Fix compilation error with Android NDK

BenEfrati:
pthread_yield is not implemented on Android NDK.
Using sched_yield instead


perexg:
Pls, put this code to src/compat.h .

perexg:
Merged. Thanks.

PR #861, opened 2016-05-11
Markdown help (yet again!)

mpmc:
These commits mostly complete the help docs, even if they're basic. Some things do need adding and no doubt there are a number of mistakes I haven't caught :laughing: .

@perexg you may want to review these fully before committing. 

The "webui: display an error message when help doc is not available/fails to load" commit is one you may want to omit if you don't think it's a good idea.


perexg:
Merged. Thanks.

PR #862, opened 2016-05-12
replace pthread_yield with POSIX sched_yield

clandmeter:
See: http://linux.die.net/man/3/pthread_yield

### DIFF
--- a/src/api/api_idnode.c
+++ b/src/api/api_idnode.c
@@ -651,7 +651,7 @@ api_idnode_handler
       }
       pthread_mutex_unlock(&global_lock);
       if (destroyed)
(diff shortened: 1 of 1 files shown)


clandmeter:
@perexg I forgot to mention this change will prevent the following error on musl libc.

```
src/api/api_idnode.c: In function 'api_idnode_handler':
src/api/api_idnode.c:654:9: error: implicit declaration of function 'pthread_yield' [-Werror=implicit-function-declaration]
         pthread_yield(); /* delete penalty */
         ^
```

clandmeter:
@perexg any reason this is not being merged, other then you are too busy?

perexg:
Merged. Thanks.

PR #864, opened 2016-05-17
Transcoding: preliminary work for codec settings/profile

lekma:
These are needed for the upcoming codec settings.
All of it has minimal impact, the only visible change is the grouping of esfilters under one tab.


CvH:
@lekma as you doing work at transcoding - it would be nice if you could drop an eye to the new ffmpeg where 
> - VAAPI-accelerated H.264/HEVC/MJPEG encoding
- VAAPI-accelerated format conversion and scaling
- libnpp/CUDA-accelerated format conversion and scaling


is included -> drop mfx dispatcher and it could be used at "all" Intels with vaapi support. This would be a major improvement and work out of the box :)

lekma:
> drop mfx dispatcher and it could be used at "all" Intels with vaapi support

that's the plan :) (not to drop libmfx right now though)

CvH:
@lekma is intel qsv still needed? It should work through the intel gpu driver obviously. It is more likely that ppl have a gpu driver then qsv/mfx installed ;)

lekma:
@CvH I agree, but I think we'll drop qsv only when vaapi has had a good test drive. Right now I'm still working out kinks when frames go through filters. That's why I needed a way to set up codecs "dynamically" to test out various combination of settings (hence the upcoming codec settings/profile tab). Unfortunately all that work does take a lot of time and I only have so much to spare.

lekma:
closing for missing parts :(
(too many local test trees -> mixups are bound to happen)
sorry

PR #865, opened 2016-05-24
Fix C/Ku Label and Change switch frequency

sielickin:
* The C/Ku label was 5150/11700, in spite of the fact that the local
  oscillator frequencies were 5150/10750. This label is now changed to
  reflect the correct frequencies. Just to be clear: the frequencies
  themselves are unchanged, still at 5150/10750. Only the description
  changed.
* To accomodate feeds operating outside of the standardized band
  (11.7GHz - 12.2Ghz), (eg: 11175 V 4200 @ 71.8W,) the switch is now set
  to operate at an arbitrary midpoint of 7.95, rather than the lower
  edge of the standard Ku freq.

Signed-off-by: Nicholas Sielicki <sielicki@bitwise>

### DIFF
--- a/src/input/mpegts/linuxdvb/linuxdvb_lnb.c
+++ b/src/input/mpegts/linuxdvb/linuxdvb_lnb.c
@@ -328,7 +328,7 @@ struct linuxdvb_lnb_conf linuxdvb_lnb_all[] = {
   },
   {
      { {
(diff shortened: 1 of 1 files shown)


perexg:
Merged (only half) - see 70c39f59896c508d70176a8b76a84b800f825eee .

sielickin:
I'm fine with that. Thanks much.

PR #868, opened 2016-06-03
tvhdhomerun: fix full mux pid filter

beralt:
I came across a warning message during some debugging that caught my attention. The original code performed a check for raw (full) PID subscriptions that did not work, and moreover the resulting call to the hdhomerun API failed because it simply refuses to subscribe to PIDs above 0x1FFF.

My proposed fix: detect the full mux subscription correctly, set the PID filter to 0x0 - 0x1FFF in that case. I also added a mask to prevent specific subscriptions above 0x1FFF, those will not work anyway.

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
@@ -331,7 +331,10 @@ static void tvhdhomerun_device_open_pid(tvhdhomerun_frontend_t *hfe, int pid) {
   tvhdebug("tvhdhomerun", "current pfilter: %s", pfilter);
 
   /* make sure the pid maps to a max of 0x1FFF, API will reject the call otherwise */
-  pid = (pid & 0x1FFF);
+  if(pid > 0x1FFF) {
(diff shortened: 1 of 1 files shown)


src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
```diff
@@ -325,6 +330,9 @@ static void tvhdhomerun_device_open_pid(tvhdhomerun_frontend_t *hfe, int pid) {
 
   tvhdebug("tvhdhomerun", "current pfilter: %s", pfilter);
 
+  /* make sure the pid maps to a max of 0x1FFF, API will reject the call otherwise */
+  pid = (pid & 0x1FFF);
```
perexg: It will be probably better to show a warning here if pid is above 0x1fff and/or simply return.

beralt: Good idea to log an error in that case, even if it is an edge case which we will probably never see. There is going to be a migration on my DVB-C network and I need mux detection to work, hence this fix.

perexg:
Merged. Thanks.

PR #871, opened 2016-06-09
config: set http digest auth to off by default, fixes #3848

mpmc:


mpmc:
Closing as a proper fix has been found, thanks @perexg  :)

PR #874, opened 2016-06-19
webui: add program image (when available) to epg detailed window

Leyorus:
Show the program image in the EPG detailed window if available (depending on XMLTV grabber used)
epg_with_preview

### DIFF
--- a/src/webui/static/app/epg.js
+++ b/src/webui/static/app/epg.js
@@ -95,6 +95,12 @@ tvheadend.epgDetails = function(event) {
     if (event.start && event.stop && event.stop - event.start > 0)
         duration = (event.stop - event.start) / 1000;
 
+    if (event.image != null && event.image.length > 0) {
+        content += '<div>';
+        content += '<img class="x-epg-image" src="' + event.image + '">';
+        content += '</div>';
+    }
+
     if (event.channelIcon != null && event.channelIcon.length > 0) {
         content += '<img class="x-epg-chicon" src="' + event.channelIcon + '">';
         chicon = 1;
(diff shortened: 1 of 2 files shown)


Saentist:
This will need free tuner when not on same mux/transponder and will take 3-5 seconds
This usual is used in multicast streamers

Leyorus:
The idea is to use the program image provided by some XMLTV grabbers (see https://tvheadend.org/issues/3108) in the web UI.

The feature to retain the image URL was implement some months ago, but 
the image was not exploited in web ui.

Saentist:
Program Icon aka logo/picon is totally different from image aka zap (image taken from live stream).
In https://tvheadend.org/issues/3108 is request for program logo

Leyorus:
In Feature #3108, Phil Tu refers to a program image provided by its XMLTV grabber (see his second post).

I confirm that some XMLTV grabbers provide this information (e.g. **tv_grab_fr_telerama**)

In my example, it corresponds to the Judoka's picture (while Canal+ is the channel logo).

The image URL is currently stored by tvheadend and even forwarded to the front end (Kodi), but not exploited in the web ui. This can be improved easily by implementing something inspired by my proposal.

perexg:
If someone comes with a little more design-like solution, I'll include it. For example - the image may be be smaller bellow the channel logo and it may be resized to EPG window info size when the mouse pointer hovers in the small image or on click etc.

tunip:
Something like this?
details_1

mouse over on program image
details_2

perexg:
Yes, it looks nice.

tunip:
see #892

perexg:
Merged #892.

PR #876, opened 2016-06-23
webui: fix idnode_filter_fields

lekma:


lekma:
the fix is in the r.splice() call

perexg:
Merged. Thanks.

PR #880, opened 2016-07-30
Fix FSF address

melmorabity:
The license headers have old addresses for the FSF. This trivial patch fixes such addresses.


perexg:
Merged. Thanks.

BTW: If you can prepare 4.1 packages for rpmfusion, it would be great.

PR #882, opened 2016-08-05
SAT>IP: reached max_pids_count force pids=all

cykedev:
When the configured max number of pids is reached, pids=all are requested.

This fixes https://tvheadend.org/issues/3903

### DIFF
--- a/src/input/mpegts/satip/satip_frontend.c
+++ b/src/input/mpegts/satip/satip_frontend.c
@@ -754,9 +754,11 @@ satip_frontend_pid_changed( http_client_t *rtsp,
     return 0;
   }
 
-  if (tr->sf_pids.all) {
+  mpegts_pid_weighted(&wpid, &tr->sf_pids, lfe->sf_device->sd_pids_max);
 
-all:
+  if (tr->sf_pids.all || (lfe->sf_device->sd_fullmux_ok && wpid.count > max_pids_count)) {
+    mpegts_pid_done(&wpid);
+
     i = tr->sf_pids_tuned.all;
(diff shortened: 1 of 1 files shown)


perexg:
Thanks. I tried to fix this a little bit differently, but the behaviour should be same. Could you test latest (commit 11d1ea7427e1ae88278b6cbaf7ebcf1295487676) ?

cykedev:
The commit looks good to me. Need to set up another system first to try this.

cykedev:
@xdarklight could you please have a look too?

cykedev:
I don't get the master branch running on FreeBSD, as I have to patch the 4.0.8_1 port. Have to set this up on another machine ...

cykedev:
@perexg I successfully tested your commit

xdarklight:
I just had a closer look at 11d1ea7427e1ae88278b6cbaf7ebcf1295487676 and it seems fine to me :+1: 

The initial problem (just in case anyone reads this) was that when enabling "many" (probably all) PIDs on one transponder @cykedev's satip dish stopped sending any data (traffic goes down to 0b/s).
However, the "else if" in satip_frontend_pid_changed always requested more pids from the satip dish which seems to have a limit of max allowed pids in add/del mode. This "else if" part did not check whether the number of pids exceed the limit so the satip dish started crying and stopped sending data.
11d1ea7427e1ae88278b6cbaf7ebcf1295487676 adds the pid count (and "all" mode supported) check to the mentioned "else if" statement, which now makes tvheadend request pids=all from the satip dish (which makes the dish happy).

PR #885, opened 2016-08-17
fixing HMF datatypes

ckarrie:
- added HMF_BOOL support
- changed misleading "map" to dict (map is a python funtion - not a datatype)


ckarrie:
Oh no, I forget 2 lines of code...

perexg:
Merged. Thanks.

PR #889, opened 2016-08-28
Improve handling of a manual deleted dvr [WIP]

Glenn-1990:
The following user case is not handled correctly with the current code:
1) Autorec rule with duplicate prevention is set up.
2) Let's say that there are 5 upcoming identical episodes.
3) The first one will be recorded and the others not (detected as rerun of)
4) Now, the user watches that first recording and deletes it afterwards.
5) The video file and logfile are deleted.
6) Now the rerun detection fails because of the missing logfile and the show will be recorded again.

My idea was to keep the logfile when the user deletes a recording (htsp or webui).
For the webui, the dvr file will be moved to the failed recordings with the comment that the user deleted the file.

However I'm not familiar with the new stuff as parent, child, don't reschedule and don't rerecord. Maybe one of those are meant to prevent this issue? @perexg Can you please explain these methods a bit?

@perexg

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -245,8 +245,8 @@ api_dvr_entry_create_by_event
                                        e, 0, 0,
                                        perm->aa_username,
                                        perm->aa_representative,
-                                       NULL, DVR_PRIO_NORMAL, DVR_RET_DVRCONFIG,
-                                       DVR_RET_DVRCONFIG, comment);
+                                       NULL, DVR_PRIO_NORMAL, DVR_RET_REM_DVRCONFIG,
+                                       DVR_RET_REM_DVRCONFIG, comment);
         if (de)
           idnode_changed(&de->de_id);
       }
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -69,7 +69,7 @@ static int is_dvr_entry_finished(dvr_entry_t *entry)
 {
   dvr_entry_sched_state_t state = entry->de_sched_state;
   return state == DVR_COMPLETED && !entry->de_last_error &&
-         dvr_get_filesize(entry, 0) != -1 &&
+         (dvr_get_filesize(entry, 0) != -1 || entry->de_file_removed) &&
          entry->de_data_errors < DVR_MAX_DATA_ERRORS;
 }
 
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -54,6 +54,7 @@ typedef struct dvr_config {
   int dvr_clone;
   uint32_t dvr_rerecord_errors;
   uint32_t dvr_retention_days;
+  uint32_t dvr_retention_minimal;
   uint32_t dvr_removal_days;
   uint32_t dvr_autorec_max_count;
   uint32_t dvr_autorec_max_sched_count;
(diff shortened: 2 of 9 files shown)


DJBenson:
There is already a 'DVR log file retention period' setting under the DVR profiles tab which (I think) does exactly this. By default, it's set to 'on file removal' but if you change it to something else, the log file will be kept for that period but you can still delete the recordings - that way, TVH maintains the state of previously recorded (but deleted) programmes so that they aren't recorded again once watched then deleted.

p.s. I've never used this functionality (although I never knew what I'd use it for, but now you've explained the use case, I might change from the default to something like 3 months).

Glenn-1990:
@DJBenson Yes, that's right, but that behaviour only applies to automatically deleted recordings (i.e. space needed or removal period expired). When pressing delete in kodi or the webui, both the log file and video file will be deleted, no matter what the retention and removal settings are.

ksooo:
Any chance not to have the entries for the recordings the user has deleted intentinally on the "failed recordings" tab? Actually they are not "failed", imo. It's just that the user has watched the recording, deletes the respective entry (using tvh webui or htsp) to save disk space and never again wants an autorec to record this event.

Just asking as I'm aware that this PR is still wip.

Glenn-1990:
@ksooo Can you test the current implementation? There is a new setting in the dvr config to set a minimum retention. I'd like to keep things simple, but simple doesn't exist in this case :p 
Using the existing retention setting for manually deleted recording could lead into problems as never deleted log files when using removal "maintain space"

perexg:
Merged.

ksooo:
@Glenn-1990 @perexg if i set the new setting "Minimal log retention period" to something else than "disabled" i cannot delete recordings using tvh webui any more (tried failed recordings tab and finished recordings tab).

ksooo:
<pre>
2016-10-03 22:29:14.291 dvr: delete entry 2286de05fea8898a0750a5c87920afa2 "ZDF-History" on "ZDFinfo HD" start time 2016-10-03 05:35:00, scheduled for recording by "kodi", retention "1096 days" removal "1 days"

2016-10-03 22:29:14.291 dvr: delete entry 2286de05fea8898a0750a5c87920afa2 not allowed "ZDF-History" on "ZDFinfo HD", current retention period 0, minimal retention period 366
</pre>

This does not look good. i cant remember to have set something to 1 day removal, for instance.

before i try to delete the recording, the settings look like this:

screen shot 2016-10-03 at 22 37 50

after the failed attempt to delete it, the props look different:

screen shot 2016-10-03 at 22 38 21

Glenn-1990:
@ksooo Your files should be deleted but this is not clearly visible as deleting the dvr logfile is prevented now. On a manual delete, the files are removed and the minimal retention gets checked. When the minimal retention doesn't allow a logfile delete, the retention will be modified to the minimal retention. (in your case, the logfile will be deleted automatically after 1 year according to picture 2). The file retention (removal) will also be modified to reflect the time of the manual deletion. (0 is invalid, that's why there is '1 day')

Currently the webui and htsp client can only do a soft delete, meaning that they can't ignore the minimal retention setting. This is something that might need some improvements. Also that the fact that the play and download buttons stay enabled for deleted recordings as well.

ksooo:
Okay, maybe that technically everything is correct, but my problem is that the "deleted" log entry stays visible in the tvh webui. For the user it seems nothing happened! In combination with the errors in the log file this is not good. 

Of course the "deleted" log entry should vanish from the list if recordings in the webui immediately and there should be no errors written to the logs.

Glenn-1990:
@ksooo Yes, I know it isn't perfect yet, that's why there is a "wip" in the title.

The big problem with vanishing is that the entries still exist but are undeletable. For example, an upcoming show is detected as duplicate because of a "vanished" recording. The user still wants to record the show by deleting the old dvr entry which is vanished -> not possible.

https://github.com/Glenn-1990/tvheadend/commits/minimal_ret_par2
Can you test this approach?

Principle:
1) lets say minimal retention is set to 365 days.
2) delete finished recording after 10 days -> recorded files will be removed -> retention will be set to 355 (365-10)days  -> dvr will be transferred to failed recordings.
2) delete failed recording -> minimal retention will be ignored and the db entry will be deleted

@perexg Would it be save to make virtually deleted recordings invisible? Or should we put them with the failed recordings or even make a new tab with "deleted recordings".

ksooo:
> Yes, I know it isn't perfect yet, that's why there is a "wip" in the title.

 yeah, i was really astonished to see this merged yesterday, tbh. 

@perexg maybe we can revert the merge until things got sorted out? I'm afraid that in current state no user will understand the feature and even experienced insiders will have problems. This is all just to complicated if you ask me. 3 settings, all dealing somehow with retention of recordings, and such...

 why dont we just have

a) file retention
- value in days describes when to automatically delete the recording file, relative to the time the recording finished
- other values:
-- keep forever
-- if disk space is needed

b) log retention
- describes when to automatically delete the log entry that belongs to a recording file, relative to the time the recording file gets deleted (automatically or manually makes no difference)
- other values: 
-- keep forever (even after file got deleted)
-- on recording file delete (no matter whether the file got deleted automatically or manually)

> Can you test this approach?

 the bad thing is that i have very limited time for testing, atm. :-(

PR #891, opened 2016-09-12
Added playcount/playposition + HTSP fixes

Glenn-1990:
1) Implemented last played position and playcount, useful when using multiple clients.
2) Some htsp fixes and improvements.

@perexg


perexg:
For discussion: The playback position should be probably maintaned per user which makes things a lot more complicated..

Glenn-1990:
Indeed, I'll think about a solution

PR #893, opened 2016-09-25
Correct the documented default config path

oliverhenshaw:
The documentation of the default for the config path option is wrong.

This can cause problems when migrating the location of the config path, e.g. when moving from a self-built tvheadend run with the default location to the fedora package, which is run with "-c /var/lib/tvheadend/config", as I need to know what the default path actually is to rename it. Moving/renaming $HOME/.hts does not pick up my settings whereas moving $HOME/.hts/tvheadend does.


perexg:
Thanks. Merged.

PR #896, opened 2016-10-04
Implemented "moved recordings" in order to improve the user experience, fixup #889

Glenn-1990:
@perexg, @ksooo 
This PR is a fixup for #889 (which was actually not ready for merging yet)

Behavior with this PR:
1) assume retention is set to something different than "on file removal" (this is still an expert setting)
2) recording finished -> recording is showing in the" finished recordings" list now
3) now the remove (instead of delete) button is showed, this will remove the stored files and place the recording in the "removed recordings" list.
4) when going to the "removed recordings" page (only in expert mode!), the user can delete the database entry. Otherwise the "removed recording" will be deleted automatically after the retention time has expired.

There are no changes in functioning or GUI when using basic or advanced mode.
schermafdruk van 2016-10-04 22 52 50

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -69,8 +69,14 @@ static int is_dvr_entry_finished(dvr_entry_t *entry)
 {
   dvr_entry_sched_state_t state = entry->de_sched_state;
   return state == DVR_COMPLETED && !entry->de_last_error &&
-         (dvr_get_filesize(entry, 0) != -1 || entry->de_file_removed) &&
+         dvr_get_filesize(entry, 0) != -1 && !entry->de_file_removed &&
          entry->de_data_errors < DVR_MAX_DATA_ERRORS;
+}
+
+static int is_dvr_entry_removed(dvr_entry_t *entry)
+{
+  dvr_entry_sched_state_t state = entry->de_sched_state;
+  return ((state == DVR_COMPLETED || state == DVR_MISSED_TIME) && entry->de_file_removed);
 }
 
 static int is_dvr_entry_upcoming(dvr_entry_t *entry)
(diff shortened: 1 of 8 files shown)


ksooo:
yeah. i like this.

perexg:
I think that dvr_entry_cancel() fcn should be kept for some consistency. Your implementation basically does two things (and one of them does not really mean 'cancel'). Create a new fcn dvr_entry_cancel_remove() plus /api/dvr/entry/remove API. Don't forget to call this fcn in htsp_method_deleteDvrEntry() , too.

Glenn-1990:
@perexg Can you explain what means and should do for finished recordings then? 
According to the htsp wiki cancel = "Cancel an existing recording, but don't remove the entry from the database." but the cancel function will definitely delete the database and not the stored files with the current implementation. https://github.com/tvheadend/tvheadend/blob/master/src/dvr/dvr_db.c#L3555
Or should we limit cancel to schedules and running recordings only?

This looks not right as storage might fill up while the files are deleted according tvheadend.

For the moment "cancel" was never triggered for completed recordings.

Glenn-1990:
@perexg Implemented a 'cancel_remove' function as you suggested. 
Also modified the 'cancel' function a bit as it may not delete database entries for finished recording, otherwise recorded files will become undeletable.

so:
cancel               -> cancel upcoming or abort active recording
cancel_delete   -> cancel upcoming and delete active and finished recordings (not visible anywhere)
cancel_remove -> cancel upcoming and remove active and finished recordings (visible as removed recording)

perexg:
Thanks. Merged.

ksooo:
@Glenn-1990 wanted to say thank you for this feature. Very valueable. Much appreciated.

PR #897, opened 2016-10-16
Fixed: UHD was not handled in tags and stream profiles

Glenn-1990:
Also renamed "radio" to "radio channels" in order to be consistent with "tv channels".
@perexg


perexg:
Merged. Thanks.

PR #898, opened 2016-10-18
Uilevel polish

Glenn-1990:
**Flaws with the current code:**
-  inconsistency between dvr, autorec and timerec: Some parameters are tagged as basic for autorec while they are tagged as expert for timerecs. Also parameters are listed in different orders and might not be visible at all. -> see screenshot
-  when adding a entry of any kind (channel,mux,bouquet,acces,etc), it should be enabled as creating a disabled entry doesn't make sense.
-  in many places there are no or almost no differences between the advanced and expert level
-  the user should be able to set up tvheadend when using "basic" mode. This is not always possible. example: adding a channel is possible but assigning a service not, this leaves the user with an unplayable channel.

dvr_autorec_timerec_inconsistency

**Changes with this PR**
- addressed flaws described above
- prevent the htsp profile for recordings and set the default to "pass" instead of "dvr" (which is an old leftover I guess).
- Basic ui level: (= beginner level) This level should be similar to normal STB's, the user can scan channels, watch them and record them. OTA epg is enabled but the user should not bother (enabled by default)    
- Advanced ui level: (= level for users familiar with tvheadend) The user can acces some settings that require more knowledge from tvheadend (internal/external epg grabber for example).
- Expert ui level (= level for computer/dvb/tcpIp experts) This settings should not be used in 99% of the cases. Everything with string formatting is tagged as expert, as well settings that might harm things or are very experimental,...

@perexg

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -1394,6 +1394,7 @@ const idclass_t access_entry_class = {
       .id       = "enabled",
       .name     = N_("Enabled"),
       .desc     = N_("Enable/Disable the entry."),
+      .def.i    = 1,
       .off      = offsetof(access_entry_t, ae_enabled),
     },
     {
@@ -1460,7 +1461,7 @@ const idclass_t access_entry_class = {
       .doc      = prop_doc_themes,
       .list     = theme_get_ui_list,
       .off      = offsetof(access_entry_t, ae_theme),
-      .opts     = PO_DOC_NLIST,
+      .opts     = PO_DOC_NLIST | PO_ADVANCED,
     },
     {
       .type     = PT_BOOL,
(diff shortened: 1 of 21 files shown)


CvH:
@Glenn-1990 
- I think `Image Cache` options should be expert, users tend to play there unnecessarily and normally you won't need it
- Timeshift `ram only` should be at least advanced because it is a decision like timeshift yes/no if you want it at hdd/ssd or ram (I guess you make it expert because a too high setting could kill your machine)
- debugging should be advanced, this should be as simple as possible without to annoy basic users (idk why but ppl have huge problems to find and use it)
- `rolloff` should be expert I think, this isn't something you should change normally, same for `pls_mode` (not too sure)

Glenn-1990:
@CvH added a comment for your suggestions ;-)

CvH:
I'am happy with it 👍

perexg:
Merged. Thanks for your work.

PR #900, opened 2016-11-04
epggrab: test code for only tuning muxes that require it

galeneca:


galeneca:
src/epggrab.h
src/epggrab/module/eit.c
src/epggrab/module/opentv.c
src/epggrab/otamux.c
src/epggrab/private.h
src/input/mpegts.h

PR #907, opened 2016-11-04
Fixes +  improvements

Glenn-1990:
Whats changed/fixed:
* HTSP clients can hold old data (epg, channels, tags) while not having acces to it anymore . Sending "channelDelete", "tagDelete", "eventDelete" to all clients (without checking access) should overcome this problem.
* Disabling tags and channels does not disable them in htsp clients.
* Duplicate detection should honor disabled dvr entries.
* Playcount/playposition is implemented in a bit different way than my previous PR. Owner can set and get the position and count. Other users can only get when "all dvr read" is enabled. When "all dvr read/write" is enabled, other users can also set playcount/position. Feel free to drop this commit if it doesn't fit in the tvheadend structure.
* Webui list the wrong filesize when the recording was split up.
* Send access right to clients so that they can inform users about limited access and such.
* some minors

While the filesize for DVR entries is correct now, there is still an issue. Only the newest file is playable through htsp or downloadable from the webui.
Example: 
1) Start recording "show A", filename="show A.ts"
2) Interrupt recording with an other higher priority one for a short time (so steal the tuner device from show A)
3) Recording "show A" resumes, filename="show A-1.ts"
4) When playing "show A" only the second file gets played (htsp and webui)!

@perexg

### DIFF
--- a/src/channels.c
+++ b/src/channels.c
@@ -77,6 +77,9 @@ channel_class_changed ( idnode_t *self )
   /* update the EPG channel <-> channel mapping here */
   if (ch->ch_enabled && ch->ch_epgauto)
     epggrab_channel_add(ch);
+
+  /* HTSP */
+  htsp_channel_update(ch);
 }
 
 static htsmsg_t *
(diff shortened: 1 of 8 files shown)


mpmc:
> While the filesize for DVR entries is correct now, there is still an issue. Only the newest file is playable through htsp or downloadable from the webui.

For htsp playback couldn't Tvheadend just play the files back-to-back in order? And for downloads use a dialog that lists all available files for the entry (if there are more than one)?

---

I really like these changes, especially "Duplicate detection should honor disabled dvr entries.". Some nice improvements! Keep it up!

Glenn-1990:
@mpmc Yes indeed, playing the files in order will be the best solution and it's a bit strange that tvheadend wasn't designed to do so. It's a relative big problem that htsp clients can't play all recorded files when the recording was split up.

ksooo:
@Glenn-1990 cool. keep up the good work.

ksooo:
@perexg any chance to get your review for this anytime soon?

perexg:
Merged all except the filesize (commit  27de533d4c183868d5a74428f5e92229f5dcc37a). The splitted files should be handled somehow differently.

PR #910, opened 2016-11-25
Removed recordings tweaks

Glenn-1990:
* List recordings with missing files as "removed", files may be deleted externally with the file manager, are saved on external storage, etc
* Keep the old recording status for removed recordings in order to keep recording errors visible. 
* Moved finished/failed/removed recording check to one central place to prevent inconsistencies.

@perexg


perexg:
Merged. Thanks.

PR #911, opened 2016-11-30
Update dvr_autorec.c

dave-p:
This PR is a suggested fix for my #4087
At present, in order for a 'Series Link' event to be recorded, not only must the series CRIDs match, but also the title of the event must (regex) match the title of the first event recorded in the series. This is not always the case, leading to recordings being missed. A current example on UK Freeview is the series "Who Do You Think You Are?", where each episode has the name of the 'star' added to the title. 
This patch has only been tested on UK Freeview.


perexg:
Could you test my fix in b93b557359da95c773e3df1675197f34334a186e ? This fix does not skip other checks like the time inverval.

dave-p:
It will be a few days before the next series link with varying titles fires, but I see no problem with your code changes. Many thanks for your help.

PR #912, opened 2016-12-06
Fix m3u parsing for extended  #EXTM3U header

fedux:
Some .m3u files have extendend parameters. For example:

`#EXTM3U url-tvg="http://xxx/" m3uautoload=1 cache=500 deinterlace=1`

Compare only the first 7 chars to allow this kind of headers.

Fixes `iptv: unknown playlist format for network ''` when the file is valid.


perexg:
Merged. Thanks.

PR #915, opened 2016-12-18
Small fixes and polishing

Glenn-1990:
The major changes:
1) Removed some restrictions regarding retention. 
When setting removal to "maintain space" the retention was forced to "on file removal". 
This means that the database entry will be deleted as soon as the file gets deleted. This can cause an unexpected rerecord by duplicate detection. 
The solution is to remove the retention restrictions and postpone it if necessary (when retention < removal).
@perexg Maybe we can set the default retention to some fixed value like "1 month" or so? 
2) Disabling an active recording is equal as aborting now, before disabling didn't have any effect.
3) Duplicate detection should ignore failed recordings (also the deleted ones)
4) Don't pass unplayable channels to htsp clients 

@perexg

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -115,7 +115,7 @@ api_dvr_entry_grid_removed
   dvr_entry_t *de;
 
   LIST_FOREACH(de, &dvrentries, de_global_link)
-    if (dvr_entry_is_finished(de, DVR_FINISHED_REMOVED))
+    if (dvr_entry_is_finished(de, DVR_FINISHED_REMOVED_SUCCESS | DVR_FINISHED_REMOVED_FAILED))
       idnode_set_add(ins, (idnode_t*)de, &conf->filter, perm->aa_lang_ui);
 }
 
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -33,10 +33,11 @@
 #define DVR_FILESIZE_UPDATE     (1<<0)
 #define DVR_FILESIZE_TOTAL      (1<<1)
 
-#define DVR_FINISHED_ALL        (1<<0)
-#define DVR_FINISHED_SUCCESS    (1<<1)
-#define DVR_FINISHED_REMOVED    (1<<2)
-#define DVR_FINISHED_FAILED     (1<<3)
+#define DVR_FINISHED_ALL             (1<<0)
+#define DVR_FINISHED_SUCCESS         (1<<1) 
+#define DVR_FINISHED_FAILED          (1<<2) 
+#define DVR_FINISHED_REMOVED_SUCCESS (1<<3) /* Removed recording, was succesful before */
+#define DVR_FINISHED_REMOVED_FAILED  (1<<4) /* Removed recording, was failed before */
 
 typedef struct dvr_vfs {
   LIST_ENTRY(dvr_vfs) link;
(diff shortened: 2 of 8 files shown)


perexg:
Merged. Thanks. Retention - I don't have any strong opinion about the default value.

ksooo:
@Glenn-1990 not 100% sure, but could you please check that this PR does not break aborting recordings in progress via htsp? Seems, that recording gets stopped correctly on server, but htsp client gets not notified.

Glenn-1990:
@ksooo tested with "stop recording" and "disable recording" in both cases the timer vanishes in Kodi, this is correct right?

Do you have some steps to reproduce? Or are you running pvr.hts with htsp v26 support?

ksooo:
@Glenn-1990 start a recording from kodi, stop it from kodi while still recording, it vanishes in tvheadend webui, but not in kodi.

Glenn-1990:
@ksooo Do you have a debug log from kodi?

ksooo:
Can you reproduce it? Should be easy. No log at hand atm, and I doubt it will really help. I think, the htsp command succeeds, but no timer update is notified to pvr.hts.

Glenn-1990:
@ksooo I'm not able to reproduce...
`AddOnLog: Tvheadend HTSP Client: pvr.hts - recording id:1774021302, state:completed`

Is it possible that you do not have "failed dvr" access enabled? An aborted recording is handled as failed by tvheadend and that would prevent updates when that setting is disabled. Although not related to this PR as the excising "abort recording" button in the webui behaves the same.

ksooo:
> Is it possible that you do not have "failed dvr" access enabled?

Hmm, that did the trick. Honestly, this is a real trap. I never noticed this "failed view" access thingy and it was actually switched off (because that's the default I guess). Works now, but can we do better (with the default, for instance)?

Glenn-1990:
@ksooo You probably created the access entry before this was in?
https://github.com/tvheadend/tvheadend/pull/898/commits/7c89ca05a29fad4f6179fc10c287fbdcdf4c5918
The default should be "on" for new entries now. I'll look into this but it's on low priority as the default is ok now.

ksooo:
> You probably created the access entry before this was in?

 Yes, I did.

> The default should be "on" for new entries now. I'll look into this but it's on low priority as the default is ok now.

 I guess many people will run into this problem after upgrading tvheadend. Not sure this is really low priority.

ProfYaffle:
Just a comment... I wonder if it's actually something that should be handled by any tvh upgrade routines, then? "here are some old profiles, here are some new settings, these are the defaults we'd like you to have"? 

I don't actually know that there *are* any upgrade routines, that said, but...

Glenn-1990:
@ksooo @ProfYaffle  It's only a problem for users who created an access entry when the default value was "false". Upgrading from older tvheadend versions (before the failed dvr option was implemented) should work properly and set the flag to true.

This issue with htsp and "failed dvr" is present since its implementation, but somehow nobody reported this until now :-) But I'll think about a fix ;-)

ProfYaffle:
Okay - so upgrade is only an issue for folks on an earlier master version, but stable-to-stable will work. Perhaps worth thinking about what versions have been in OE/LE as probably the biggest "regular user" population of pre-release code, though, if that is the case.

I'll vanish again now :)

ksooo:
> Okay - so upgrade is only an issue for folks on an earlier master version, but stable-to-stable will work.

 If this is the case, all is fine. Thanks guys. :-)

PR #916, opened 2016-12-22
Fix for compiling fdk-aac with GCC6

untitledxi:
As noted in issue [4043](https://tvheadend.org/issues/4043), compiling tvheadend with GCC6 fails as the standard defaults to a newer version. The suggested solution, to disable ffmpeg is unfeasible for some setups.

As the original project is rarely updated (noted in https://github.com/mstorsjo/fdk-aac/issues/45), i've implemented a hackish fix until the library is updated.

Feel free to revert/remove this PR, once the library is updated.


CvH:
we have used this "hack" to [make it work](https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/addons/service/tvheadend42/patches/tvheadend42-03_gcc6.patch)

the current fdk-aac master from https://github.com/mstorsjo/fdk-aac/ should also work (this is the upstream git for the fdk-aac project)

perexg:
Merged. Thanks.

PR #917, opened 2016-12-30
Make SAT>IP requests for pids=all work

caveman99:
pids=all would result in only pid 0 being sent. transedit uses this to analyze ts content which is again possible with this patch.


perexg:
Merged. Thanks.

PR #919, opened 2017-01-13
Fix weird channel numbers

ricardopvz:
When selecting "Use service IDs as channel numbers" in the network settings, the `s_dvb_service_id` (uint16) multiplied by `CHANNEL_SPLIT` (int64), would produce a value that would overflow the `r` variable (int) if the service id is was high enough, causing the result to be weird numbers.


perexg:
Merged. Thanks.

PR #920, opened 2017-01-22
Fixes

Glenn-1990:
Biggest changes:
1) Recordings will now be aborted when running out of space, also when the user does not have any "maintain free space" recordings.  Without this, tvheadend will keep recording until file writing fails, at this point the OS can become very unstable. 
2) Improved htsp async methods. The add/update/delete methods are very inconsistent due to different access right. Under some circumstances "update" was sent instead of "add", "delete" was sent to clients which never received the "add", "delete" was never sent due to access right changes,.. 
This does also fix the issue @ksooo reported: failed dvr access is off -> abort recording -> recording is marked as failed so the user does not have access anymore -> because of no access, "dvrEntryDelete" won't be sent to the client -> client thinks it's still recording.
@perexg Do you agree with the solution of keeping a local copy from all added dvrs/tags/channesl/events? In this way we can properly determine when to send add, update or delete.
I also added a fixme as dvr_rec.c calls some htsp functions without global lock, not related to this PR.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -559,7 +559,8 @@ dvr_entry_retention_timer(dvr_entry_t *de)
   }
   else {
     dvr_entry_retention_arm(de, dvr_timer_disarm,
-        dvr_entry_get_rerecord_errors(de) ? INT_MAX : 0); // extend disarm to keep the rerecord logic running
+        dvr_entry_get_rerecord_errors(de) ?
+            time_t_out_of_range((int64_t)de->de_stop + DVR_RET_REM_1MONTH * (int64_t)86400) : 0); // extend disarm to keep the rerecord logic running, give up after one month
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/dvr/dvr_db.c
```diff
@@ -559,7 +559,8 @@ dvr_entry_retention_timer(dvr_entry_t *de)
   }
   else {
     dvr_entry_retention_arm(de, dvr_timer_disarm,
-        dvr_entry_get_rerecord_errors(de) ? INT_MAX : 0); // extend disarm to keep the rerecord logic running
+        dvr_entry_get_rerecord_errors(de) ?
+            time_t_out_of_range((int64_t)de->de_stop + DVR_RET_REM_1MONTH * (int64_t)86400) : 0); // extend disarm to keep the rerecord logic running, give up after one month
```
perexg: This change is useless. dvr_entry_retention_arm() already uses one hour timers when dvr_entry_get_rerecord_errors() is non-zero.

perexg:
Applied 'dvr: properly abort recordings when running out of free disk space' and 'htsp: pass failed dvr right'. The timer fix is useless and I need to think about the async id filtering more.

chewitt:
Closing as the main feature in the PR was picked to commit cf818c05b82a15ae9caa4aaabc20c814c5d84548.

PR #921, opened 2017-01-23
Implement service_refresh_channel method in service.c

ricardopvz:
Correct me if I'm wrong, but without this, HTSP clients will not get notified of changes to the service from the provider like name, number, etc


perexg:
Merged. Thanks.

PR #922, opened 2017-01-23
Allow Authorization Header

r15ch13:
Allows Authorization Header to be used by XMLHttpRequest.

If the Authorization Header is not allowed, it's not possible to use Basic-Authentication from Javascript requests in the Browser (Ajax).
A Browser will throw this error: `XMLHttpRequest cannot load http://tvheadend/. Request header field Authorization is not allowed by Access-Control-Allow-Headers in preflight response.`


perexg:
Merged. Thanks.

PR #925, opened 2017-02-15
Make separate log category for EMM messages

joeusercz:
When tracing descrambler, quite often the provider is sending many EMM messages which end up flooding the trace log. Separating EMM messages allows you to choose whether or not to include them in the log.


perexg:
Merged.

PR #926, opened 2017-02-15
use png optimized images

CvH:
`AUTOBUILD_CONFIGURE_EXTRA=--enable-bundle ./Autobuild.sh -t precise-amd64`

```
19155958 tvheadend_4.1-2435~g6129d04~precise_amd64.deb
12501500 tvheadend_4.1-2439~ga436633~precise_amd64-with-png-commit.deb
```
saves ~35% (6,5mb) size overall at no real quality loss at the pics
you can use pngquant from os repo or compile yourself from gh
```
./pngquant --quality=85-90 --speed 1 --ext=.png --force *.png
``` 
to optimize these pngs.

@perexg and @mpmc would be nice if this is the new "default" because it saves so much size without visible quality drop


CvH:
rebased

perexg:
I don't like this solution, but I added on-the-fly png optimizations for the bundle builds in v4.1-2506-g8bf3932 - 8bf3932539ceadb3bfc1c9453f11175f15aefad2 . It's few lines of code :-)

CvH:
@perexg an even better solution :) 👍

PR #927, opened 2017-02-16
Various little fixes.

mpmc:
I noticed that the "Basic" streaming and DVR options weren't set when creating users via the wizard setup. This commit adds those options as default.

Edit: Additional fixes.. I noticed some browsers will try and login using cached details sometimes causing a login loop when clicking the "Default Login" link, replacing it with /logout should hopefully prevent this.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: tvheadend
 Section: video
 Priority: extra
 Maintainer: Adam Sutton <aps@tvheadend.org>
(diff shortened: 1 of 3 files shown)


perexg:
The permissions fix was not correct. I tried to fix this in b86ac2309f9b1f9332f6fa21808ae18cedb919f8 . Applied the other two commits. Merged.

mpmc:
I had a feeling it wasn't right, but it _seemed_ to work when I tried it, but hey-ho I'm no C developer (and it shows! 🤣 ). I really appreciate you fixing it and letting me know that I goofed!

PR #928, opened 2017-02-19
Timeshift broken on stream restart, issue #4096

Glenn-1990:
When a subscription gets overridden (and possible in other "stop" cases too), the timeshift reader and writer threads gets closed. The subscription stays running in idle state, but when it resumes (free tuner for example), timeshift won't be restarted.

This causes 2 problems:
1) subscription hangs and never resumes on the client after an subscription override (or other stop code), even if it is "running" according to the tvheadend status tab.
2) "sm_code" messages not sent to clients when SMT_NOSTART is issued. So the user is left in the dark why playback stopped. 

Related to:
https://tvheadend.org/issues/4096#change-20566

@perexg I'm not sure if this is a "correct" fix, please give your opinion

### DIFF
--- a/src/timeshift.c
+++ b/src/timeshift.c
@@ -370,9 +370,12 @@ static void timeshift_input
     }
 
     /* Check for exit */
-    else if (type == SMT_EXIT ||
-        (type == SMT_STOP && sm->sm_code != SM_CODE_SOURCE_RECONFIGURED))
+    else if (type == SMT_EXIT)
       ts->exit = 1;
+
+    // Don't exit timeshift on SMT_STOP as the subscription might restart later on.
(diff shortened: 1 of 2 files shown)


perexg:
The buffers (all previous data) must be removed in this case, otherwise the PTS timing will break.

Glenn-1990:
But this means that we won't be able to skip over the gaps in the timeshift buffer?

Saentist:
Its more good to be added timer "zero clients live time" 
where if client X disconnect from stream and same client X isnt change channel to have Y time to reuse stream and time shift, else timeshift buffer to be destroyed .

ljalves:
@perexg 
"...otherwise the PTS timing will break."
Doesn't that already happens when a scrambled channel fails to decode for a couple of seconds and continue to decode later on?
I believe most players can recover from a PTS discontinuity and resume playback, right?

PR #930, opened 2017-03-02
Append adapter number (An) to adapter name

joeusercz:
When people have more than one identical adapter installed, it is sometimes hard to tell which adapter is being referenced in the logs. Especially if they assign the same network to both adapters (Dual LNB, etc...)


perexg:
Merged. Thanks.  Following cosmetic changes are in 82cfcc771961b1fec039381cf359ce372b701294 .

PR #935, opened 2017-03-16
enable PROXY protocol and X-Forwarded-For

ap-satip:
As discussed in #4247
http://www.tvheadend.org/issues/4247

This supersedes Pull request #640
http://github.com/tvheadend/tvheadend/pull/640

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1676,6 +1676,7 @@ config_boot ( const char *path, gid_t gid, uid_t uid )
   config.idnode.in_class = &config_class;
   config.ui_quicktips = 1;
   config.digest = 1;
+  config.proxy = 0;
   config.realm = strdup("tvheadend");
   config.info_area = strdup("login,storage,time");
   config.cookie_expires = 7;
@@ -2124,6 +2125,20 @@ const idclass_t config_class = {
                    "It is intended to replace unencrypted HTTP basic access authentication. "
                    "This option should be enabled for standard usage."),
       .off    = offsetof(config_t, digest),
+      .opts   = PO_EXPERT,
+      .group  = 1
+    },
+    {
+      .type   = PT_BOOL,
+      .id     = "proxy",
+      .name   = N_("Use PROXY protocol & X-Forwarded-For"),
+      .desc   = N_("PROXY protocol is an extension for support incoming "
+                   "TCP connections from a remote server (like a firewall) "
+                   "sending the original IP address of the client. "
+                   "The HTTP header 'X-Forwarded-For' do the same with "
+                   "HTTP connections. Both enable tunneled connections."
+                   "This option should be disabled for standard usage."),
+      .off    = offsetof(config_t, proxy),
       .opts   = PO_EXPERT,
       .group  = 1
     },
--- a/src/http.c
+++ b/src/http.c
@@ -1100,7 +1100,13 @@ process_request(http_connection_t *hc, htsbuf_queue_t *spill)
   char authbuf[150];
 
   hc->hc_url_orig = tvh_strdupa(hc->hc_url);
+
+  v = (config.proxy) ? http_arg_get(&hc->hc_args, "x-forwarded-for") : NULL;
+  if (v)
+    tcp_get_sockaddr((struct sockaddr*)hc->hc_peer, v);
+
   tcp_get_str_from_ip((struct sockaddr*)hc->hc_peer, authbuf, sizeof(authbuf));
+
   hc->hc_peer_ipstr = tvh_strdupa(authbuf);
   hc->hc_representative = hc->hc_peer_ipstr;
   hc->hc_username = NULL;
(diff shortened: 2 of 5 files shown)


src/config.c
```diff
@@ -2128,6 +2129,20 @@ const idclass_t config_class = {
...
     {
+      .type   = PT_BOOL,
+      .id     = "proxy",
+      .name   = N_("Use PROXY protocol & X-Forwarded-For"),
+      .desc   = N_("PROXY protocol is an extension for support incoming "
+                   "TCP connections from a remote server (like a firewall) "
+                   "sending the original IP address of the client. "
+                   "The HTTP header 'X-Forwarded-For' do the same with "
+                   "HTTP connections. Both enable tunneled connections."
+                   "This option should be disabled for standard usage."),
```
mpmc: It might be a good idea to mention that it's IPv4 only.

ap-satip: Hi,

It could have IPv6 support, as the PROXY protocol supports it. Also, it could have PROXY v2 (binary) support too. However, we prefer to only integrate the most simple option at time.

Feel free to add "(Note: only IPv4 support at time)".

src/http.c
```diff
@@ -1460,6 +1466,48 @@ http_serve_requests(http_connection_t *hc)
...
+        if ( (strlen(pl) >= 7) && (strncmp(pl,"UNKNOWN",7) == 0))
+            goto error;  // Unknown PROXY protocol
+        
+        if ( (strlen(pl) < 5) || (strncmp(pl,"TCP4 ",5) != 0))
+            goto error;  // Only IPv4 supported
+        pl += 5;
+
+        // Check the SRC-ADDRESS
+        c = pl;
+        char ch;
```
perexg: Move all declaration on the start of block {} or ideally at the function start.

src/http.c
```diff
@@ -1460,6 +1466,48 @@ http_serve_requests(http_connection_t *hc)
...
+        for ( ;; ) {
+            if (strlen(pl) == 0) goto error;  // Incomplete PROXY format
```
perexg: Replace 'strlen(pl) == 0' with 'pl[0] == '\0'' - it's more efficient.

src/http.c
```diff
@@ -1460,6 +1466,48 @@ http_serve_requests(http_connection_t *hc)
...
+            ch = *pl++;
+            if (ch == ' ') break;
+            if (ch != '.' && (ch < '0' || ch > '9')) goto error;  // Not valid IP address
+        }
+        if (((pl-c) < 8) || ((pl-c) > 16)) goto error;  // Not valid IP address
+        
+        // Here 'c' points to a dotted IPv4 SRC-ADRRESS
+        char srcaddr[16];
+        memset(srcaddr, 0, 16);
```
perexg: Remove this memset() and add 'srcaddr[(pl-c)-1] = '\0'' after strncpy() call.

src/tcp.c
```diff
@@ -450,6 +450,28 @@ tcp_get_ip_from_str(const char *src, struct sockaddr *sa)
 /**
  *
  */
+int
+tcp_get_sockaddr(struct sockaddr *sa, const char *s)
```
perexg: Why you don't use already prepared tcp_get_ip_from_str() function which does the exactly similar thing ?

src/http.c
ap-satip: "ch" is declared here for better readability... is strictly local var.
If you prefer I can move after "pl" declaration.

src/http.c
ap-satip: strlen() is too more easy to read...

src/http.c
ap-satip: Sorry! Safe memset() of a String is a good coding technique.
Why you prefer to add the null-terminating char if the strncpy() function adds it?

src/tcp.c
ap-satip: Because I don't know about this other function... our patch is based on the work of Steffen Vogel.

src/http.c
linuxstb: Or move the check a line further down and use "if (ch == '\0')" or even if (!ch).

lars18th: All comments are acceptable.
However, our goal is implementing the code "safe" (=without errors or memory leaks) and "clear" (=easy to read and understand).

lars18th:
Hi @perexg ,

Please, can you accept the PR and then include your suggestions?
We fully agree with your comments, but we like to focus in other issues/functions.
As you know your code very well, it will be easy to adapt it.

src/http.c
ap-satip: All comments are acceptable.
However, our goal is implementing the code "safe" (=without errors or memory leaks) and "clear" (=easy to read and understand).

ap-satip:
Hi @perexg ,

Please, can you accept the PR and then include your suggestions?
We fully agree with your comments, but we like to focus in other issues/functions.
As you know your code very well, it will be easy to adapt it.

perexg:
Merged with heavy modifications. 366b0dd24ffabb12999bf7171b70458c73182e88 c2ec95cabbfed6a318ac010a4e1b1e1d41c02d2f

ap-satip:
Tahnk you!
We will check it. If something is wrong, then we will open a ticket.
From my point of view, all modifications are fine... the new implementation also supports IPv6.
Great!

Regards.

sly-net:
Hello,
I'm sorry to write this here but I was unable to create an account on tvheadend.org to submit the following issue (never received the validation email, even after checking spams).

I just wanted to report the following issue: https://github.com/tvheadend/tvheadend/commit/9154ea387346ad1b2864a02397a36bc2cfb9dd5c#commitcomment-49111889

PR #938, opened 2017-03-27
[Feature #4229] IPTV/RTP input: handle source specific multicast (SSM)

duncanthrax:
This adds support for source-specific multicast (SSM) for IPTV RTP input. Reference: https://tvheadend.org/issues/4229

### DIFF
--- a/src/udp.c
+++ b/src/udp.c
@@ -119,7 +119,6 @@ udp_get_ifindex( const char *ifname )
   return r;
 }
 
-#if defined(PLATFORM_LINUX) || defined(PLATFORM_DARWIN)
 static int
 udp_get_ifaddr( int fd, const char *ifname, struct in_addr *addr )
 {
@@ -138,7 +137,6 @@ udp_get_ifaddr( int fd, const char *ifname, struct in_addr *addr )
          sizeof(struct in_addr));
   return 0;
 }
-#endif
 
 static int
 udp_get_solip( void )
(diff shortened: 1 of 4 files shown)


src/udp.c
```diff
@@ -233,11 +233,48 @@ udp_bind ( int subsystem, const char *name,
       }
 #endif
 
-      if (setsockopt(fd, udp_get_solip(), IP_ADD_MEMBERSHIP, &m, sizeof(m))) {
-        inet_ntop(AF_INET, &m.imr_multiaddr, buf, sizeof(buf));
-        tvhwarn(subsystem, "%s - cannot join %s [%s]",
-                name, buf, strerror(errno));
+#if defined(PLATFORM_LINUX)
```
perexg: I would move the code outside PLATFORM_LINUX - or ms.imr_sourceaddr does not exist on BSD / Darwin ?

duncanthrax: I'd be happy to, but I can't do a test build on these platforms ATM. If that is not a problem, I'll do it.

perexg: From a quick google lookup, it seems that almost all platforms have imr_sourceaddr, so pls, remove this block outside ifdef.

duncanthrax: Done, since this means that udp_get_ifaddr() is then potentially used on platforms other than OSX and Linux, I checked for the SIOCGIFADDR as well - it seems to exists universally, so I guess that's OK.

If you prefer a new PR with a single commit let me know.

perexg:
Applied. Thanks.

PR #940, opened 2017-04-05
Doozer update

mpmc:
These commits change the doozer.json to include new targets & output deb packages.

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -19,8 +19,6 @@
         "python",
         "dvb-apps",
         "debhelper",
-        "libcurl4-gnutls-dev",
-        "curl",
         "ccache"
       ],
       "buildcmd": [
@@ -45,8 +43,6 @@
         "python",
         "dvb-apps",
         "debhelper",
-        "libcurl4-gnutls-dev",
-        "curl",
         "ccache"
       ],
       "buildcmd": [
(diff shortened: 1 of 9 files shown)


mpmc:
@perexg 

There is an issue with wget on Ubuntu 12.04 that prevents the download of ffmpeg due to certificate issues, hence the failed build on precise:i386. I did do a commit that resolved it by replacing wget with curl. However, I decided to leave it as is & let the builds fail as no one should really be using 12.04 now. 

But to prevent this in the future, would it be possible for you to change the Makefile.ffmpeg to include some sort of logic that'll attempt to download the packages using curl (if available) when wget fails due to cert issues? I would've done it myself, but I'm no bash expert & more than likely would've got it completely wrong. :rofl:

perexg:
It's enough to pass '--no-check-certificate' to wget . But it's  a question, if we trust this source. Only fdk-aac requires this. I tried to fix this in 9f12730980251e15b6c8cae461625dca8239c67e . Perhaps, you can remove curl again :-)

mpmc:
curl was already listed in the build dependencies, so I decided to leave as is. Plus I didn't want to screw with the Makefile ffmpeg in case it breaks on other distros!

perexg:
I think that curl should be removed. The RPM packaging does not use it at all.

perexg:
libcurl4-gnutls-dev package is not probably required, too

PR #942, opened 2017-04-10
Ability to selectively transcode video stream in trancoding streaming profile.

Den4t:
Ability to selectively transcode video stream in trancoding streaming profile.
New option added in transcoding page: Source video codec,
can be selected from list or separated with " ;,|", in case of codec match,
do transcode with profile settings, else pass with copy codec..


src/profile.c
```diff
@@ -2040,6 +2071,14 @@ profile_transcode_work(profile_chain_t *prch,
   props.tp_vbitrate   = profile_transcode_vbitrate(pro);
   props.tp_abitrate   = profile_transcode_abitrate(pro);
   strncpy(props.tp_language, pro->pro_language ?: "", 3);
+  
+  if(!pro->pro_src_vcodec) {
+	  strcpy(props.tp_src_vcodec, "");
+  } else if(!strncasecmp("Any",pro->pro_src_vcodec,3)) {
```
perexg: This looks wrong. Use the key match, not the language string , so the first condition should be:

  if(!pro->pro_src_vcodec || pro->pro_src_vcodec[0] == '\0') {

Den4t: Fixed.

Den4t:
Fixed in 5f8c0c9.

CvH:
@Den4t if you need inspiration to fill your free time you can head over to https://github.com/lekma/tvheadend/tree/codecs - sadly this is unfinished till date :/ 
("proper" codec profiles + gpu acceleration for Tvh - basically the last main feature Tvh still misses)

Den4t:
One test was failed, but it is not a compile error,
problem with certificate at doozer. How to run test again ?

perexg:
Merged to 4.2 and 4.3 . Thanks.

PR #943, opened 2017-04-16
Htsp playstatus improvement

Glenn-1990:
Currently the implemented htsp playstatus methods are kind of useless for some clients.
This for the following reasons:
a) The client might want to set playcount/playposition by itself. For example, some clients do this after 'x' percent/minutes of the show was watched. 
b) The property "watched count" is past tense, so strictly speaking we should increase the count after watching and not before (so on file close instead of file open). 
c) The auto increment conflicts with "htsp_method_updateDvrEntry", they both can increase the playcount and that can cause issues as playcount +2 instead of +1.

a) This will be handled with some new constants, "HTSP_DVR_PLAYCOUNT_KEEP " will ignore tvheadends auto increment when the client wants to set the playcount by itself.
b) Auto increment (when desired) will be done on file closing.
c) Also handled with 1)

@perexg


perexg:
The code and idea looks good. Merged. Thanks.

PR #944, opened 2017-04-24
mdhelp: add note about service probing for IPTV Automatic Network

mpmc:
This could be added to release/4.2 if needed.


perexg:
Merged. Thanks.

PR #946, opened 2017-05-11
fix detecting number of atsc-t satip tuners

catalinii:
Hi,

tvheadend does not detect correctly the number of atsc-t tuners.

Thank you


perexg:
Merged, Thanks. 6e99e14e7ec78c20ebc8cdfa4724363f7cf35bce

PR #949, opened 2017-05-15
Add cccam CA client

ljalves:
Based on cwc.c and oscam cccam module.

Still needs a few enhancements, being the major:
- extended mode - multiple requests at once (does not mean "no multiple channels" - it means that to send a new ecm/emm request, it has to wait for the answer to the previous one)
- emm's

Tested on my side and seems to work good

### DIFF
--- a/src/descrambler/cccam.c
+++ b/src/descrambler/cccam.c
@@ -1851,24 +1851,23 @@ caclient_cccam_nodeid_set(void *o, const void *v)
   cccam_t *cccam = o;
   const char *s = v ?: "";
   uint8_t key[8];
-  int u, l, i = 0;
+  int i, u, l;
+  uint64_t node_id;
 
-  if (!strlen(v)) {
-    uuid_random(cccam->cccam_nodeid, 8);
-    i = 1;
-  } else {
-    for(i = 0; i < ARRAY_SIZE(key); i++) {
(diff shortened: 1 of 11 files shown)


src/descrambler/cccam.c
```diff
@@ -1851,17 +1851,24 @@ caclient_cccam_nodeid_set(void *o, const void *v)
...
-
-  for(i = 0; i < ARRAY_SIZE(key); i++) {
-    while(*s != 0 && !isxdigit(*s)) s++;
-    u = *s ? nibble(*s++) : 0;
-    while(*s != 0 && !isxdigit(*s)) s++;
-    l = *s ? nibble(*s++) : 0;
-    key[7-i] = (u << 4) | l;
+  int u, l, i = 0;
+
+  if (!strlen(v)) {
```
perexg: It's better to check, if the hexa string contains all zero bytes ?

<pre>
  if (memcmp(key, "\0\0\0\0\0\0\0\0", 8) == 0) {
    uuid_random(key, 8);
  }
</pre>

ljalves: modified: now comparing result with zero
Also fixed a bug on the byte display order (endianess)

perexg:
Merged. Thanks.

PR #950, opened 2017-05-16
Debian: add libpcre3-dev build dep to control file

mpmc:


perexg:
Note that 'libpcre2' (version 10.0+) is NEWER than 'libpcre3' (version 8.0+)... !!!! It should be probably libpcre2-dev ..

perexg:
Merged.

PR #954, opened 2017-05-18
Fixed compile error if pcre is not installed.

UllrichKossow:
Not tested, but compile clean with and without pcre library installed.


perexg:
Merged. Thanks.

PR #955, opened 2017-05-21
Wizard: prevent the SOH character from appearing in the wizard text -…

mpmc:
… fixes #4282

Maybe not the best way to fix it but it works!


perexg:
Fixed in e3b90de8fb52d803f3f5df5bf4dbc0d3cac15b6e and aea3be89252e92751da21ea7df270365e26dd0b9 . But thanks for this PR - it directed me to the right place !

PR #956, opened 2017-05-23
HTSP Python: function < integer will never return false

mastacheata:
Sorry, if this is a strange Python thing I don't understand, but to me it looks like this has at best undefined behaviour and will at worst be completely useless because len < nlen+dlen will never be able to return false. (len is a function and needs a parameter to return any useful numeric information)


perexg:
Applied, thanks.

PR #959, opened 2017-06-08
Fix dvr priorities after 33cc05b

ksooo:
Use new DVR_PRIO_DEFAULT wherever it should be used (HTSP, API, ...).


perexg:
Applied as 8d3bd9dd80087b1c238e666e8ce680ebdc309523 . The _dvr_entry_update() argument DVR_PRIO_NOTSET looks fine in my eyes (the priority should not change in those cases), so I removed this from your commit.

Final commit: 8d3bd9dd80087b1c238e666e8ce680ebdc309523

Thanks.

ksooo:
Ah, okay. Thanks.

PR #961, opened 2017-06-15
Compatibility fix: Adding atomic32 support

m4tt075:
This compatibility fix has been around as patch for older platforms for a couple of years now. I don't know who came up with this. Anyways, looks clean to me, so I was wondering whether it can be upstreamed...


m4tt075:
Forgot to adapt variable names. Fixed, squashed and force pushed. Should be ok now.

perexg:
Merged. Thanks.

PR #962, opened 2017-06-15
bintray: don't automatically publish dirty builds

mpmc:
bintray: allow org override

@perexg, this should prevent apt getting confused :)


perexg:
Merged. Thanks.

PR #964, opened 2017-06-20
Compatibility fix for undefined _SC_IOV_MAX

m4tt075:
This is the last compatibility fix, which has been included in the spksrc cross-compilation framework for tvheadend (downstream) for years. 

To be honest, technically I can see what it does, but I do not know what symptoms it fixes: There might be different reasons for this. Either, because I just do not have an architecture which exposes the corresponding problems, or, because this patch is actually no longer needed. I don't know. With this in mind...

@perexg, can you carefully review, please, and let me know what you think?


perexg:
Merged. Thanks.

PR #968, opened 2017-06-27
Mdhelp - More changes!

mpmc:
@perexg Could you check to make sure the **Change parameters** info/table is correct?

Thanks.


perexg:
Merged. Thanks.

PR #969, opened 2017-07-01
sort file lists

bmwiedemann:
to generate docs_inc.c reproducibly
See https://reproducible-builds.org/ for why this matters


perexg:
Merged. Thanks.

PR #970, opened 2017-07-01
Allow to override build date

bmwiedemann:
to enable reproducible builds

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

The date calls are designed to work with both GNU date and BSD date

The alternative would be to drop the build timestamps
in favour of version numbers and/or git commit hashes.


perexg:
Merged. Thanks.

PR #975, opened 2017-07-17
Add "Now" view mode in EPG (#3411)

LubosD:
The EPG viewer in the web interface isn't currently what I would expect as a user. It basically cannot be used to show what is **currently** being broadcast on all channels, sorted by channel number (i.e. something TVHClient on Android does).

I've implemented such view mode, and I'm looking forward to hearing your comments.

I'd event suggest make this view mode default, but I'll leave it up to you.


perexg:
Merged. Thanks.

PR #976, opened 2017-07-17
Add a button to quickly play a channel in EPG view (#3926)

LubosD:
When scrolling through EPG view, I miss a button that would quickly open "Watch TV" with given channel. This implements such a feature.

### DIFF
--- a/src/webui/static/app/tvheadend.js
+++ b/src/webui/static/app/tvheadend.js
@@ -487,11 +487,17 @@ tvheadend.playLink = function(link, title) {
 /**
  * Displays a mediaplayer using the html5 video element
  */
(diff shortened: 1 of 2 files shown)


src/webui/static/app/tvheadend.js
```diff
@@ -487,11 +487,17 @@ tvheadend.playLink = function(link, title) {
 /**
  * Displays a mediaplayer using the html5 video element
  */
-tvheadend.VideoPlayer = function(url) {
```
LubosD: The url parameter seems to be unused, so I replaced it with channelId.

perexg:
Merged. Thanks.

PR #977, opened 2017-07-18
Upgrade version of x264-snapshot

BKPepe:
To the latest available according to: http://ftp.videolan.org/x264/snapshots/

//EDIT: I leave it here, but it's not necessary. I thought as it is in commit's comment it helped, but it didn't. I compiled it probably w/o ffmpeg. That's the reason, why I thought it helped me. (Funny, right?
 :D) There is no button to remove it. Only close, but it's up to you. :-)
Maybe it would be better to use stable?


perexg:
Unfortunately, the dependency on newer nasm version breaks the debian builds (check the Travis CI).

PR #978, opened 2017-07-24
Add Delete Autorec to API

dave-p:
This PR adds the ability to delete an Autorec via the HTTP API using the new URL api/dvr/autorec/cancel. Tested on UK Freeview using 'Series Link' autorecs.


perexg:
Use api/idnode/delete to destroy an idnode object. You just wrote the identical code. Closing.

PR #979, opened 2017-07-26
WebUI: add return to sort-by-date when toggling back to 'All' from 'Now'

ProfYaffle:
.. as discussed on https://github.com/tvheadend/tvheadend/commit/620c50b9883b9e767e9addcc68afc56af3d72b36#commitcomment-23309464

... this forces an EPG sort by start time when toggling back to "All" view from "Now" (currently-running) view. Without it, the sort remains by channel number; a quick test suggests that simply removing any sort order and the EPG defaults to sorting by duration, which is unlikely to be useful.


perexg:
Merged. Thanks.

perexg:
Merged. Thanks.

PR #982, opened 2017-07-30
Add series link support to htsp autorec methods.

ksooo:
I need this to be able to schedule series link autorecs via htsp client (Kodi HTSP PVR Addon - pvr.hts, in this case).


perexg:
Merged. Thanks.

PR #983, opened 2017-08-05
Issue in Fr translation

JF6:
For autorec, Sat means Saturday (Samedi in French), not Satellite.
Changed the string into Sam ("Samedi" abreviated)


Saentist:
Translation is not in this Git it is located in 
https://www.transifex.com/tvheadend-foundation-cic/tvheadend/dashboard/
there is located all translations
(synchronisation is other question)

perexg:
Yes, please, make all language specific changes on transifex. We will import texts before new 4.2.x release.

PR #985, opened 2017-08-17
Autobuild: add raspbian stretch target ready for doozer.

mpmc:


perexg:
It seems like a dup of #986, closing.

PR #988, opened 2017-09-03
Av desync

blablack:
From the dev's comments, I understand this is not a clean fix for https://tvheadend.org/issues/4507 - but can we still integrate this workaround until a proper solution is put in place?


perexg:
Fixed in 8a43edd48e00a724eba76b37418ee18840dbe60e except the reversed no_audio check.

PR #990, opened 2017-09-07
4287 scrape episode from ota eit

azlm8t:
This patch allows optional scraping of season, episode, and airdate information                                                                                                                                                                      
from the EIT summary data for EIT-based parsers (uk, nz, etc. not psip
nor opentv).                                                                                                                                                      
                                                                                                                                                                                                                                           
Broadcasters often include season and episode number in the summary                                                                                                                                                                        
text in the OTA EIT. For example UK broadcasters often have a                                                                                                                                                                              
description of "Lorem ipsum (S5 Ep 8)" or "Lorem ipsum (S3 Ep 4/9)" or                                                                                                                                                                     
"Lorem ipsum (Ep 4)", and movies have an airdate such as "Lorem (1998)".
So this can be extracted using regex.                                                                                                                                                                                           
                                                                                                                                                                                                                                           
The main logic is only about ten lines in commit 2 and is based
on the opentv logic.

The rest of the commits are re-factoring of existing regular expression parsing 
in to a separate file (commit 1) and adding GUI configuration to enable/disable the                                                                                                                                                                         
processing. Modifications to the other parsers is to pass a flag to
indicate whether they support scraping.                                                                                                                                                                                                                          
                                                                                                                                                                                                                                           
The regular expressions are loaded from a configuration file based                                                                                                                                                                         
on the module name such as uk_freesat. If that file does not exist                                                                                                                                                                         
then it tries the first component before the underscore, so "uk"                                                                                                                                                                           
or "nz". Alternatively, the user can override the configuration file                                                                                                                                                                       
in the GUI, which is useful for the generic "eit" module.                                                                                                                                                                                  
                                                                                                                                                                                                                                           
The GUI also has a tick box to enable the scraping. The intention is                                                                                                                                                                       
that in the future we might have other optional functionality such as                                                                                                                                                                      
"tidy title to remove the 'New:' prefix" which some users might not                                                                                                                                                                        
want.                                                                                                                                                                                        
                                                                                                                                                                                                                                           
To enable the functionality, go to Configuration->Channel->EPG Grabber                                                                                                                                                                         
Modules->Over the Air. Enable "Scrape episode". In the generic "DVB                                                                                                                                                                        
Grabber" I enter "uk" to cause it to load the uk configuration since I have
generic "DVB Grabber" enabled along with freeview and freesat.

I've only supplied the uk configuration. I can see issue 4509 which appears
to be for Bulsatcom and can request some more information if the pull
request is accepted.

The scraping can be disabled so should have no impact on very low
spec machines.

One minor issue I noticed is that on one of my minor channels and the sister +1 timeshift channel 
broadcast the same programme with the same crid but with different description: one with
season/episode information and one without. Since tvh lookups are by crid,
that means tvh constantly overwrites the description depending on the last one parsed.
Previously it wouldn't really be noticed by anyone that the description constantly changes.
However, this means sometimes the GUI shows a description containing season/episode when
you click info, but the season/episode field is blank due to tvh parsing the other
description in the interim. So not a bug, just an existing annoyance that I'll mention
since it looks unexpected and took a while to track down.


perexg:
Merged. thanks.

PR #991, opened 2017-09-08
DVR: Prefer autorec rule name if comment field is empty (#4500)

azlm8t:
Currently the Upcoming recordings tab has a comment field that says
"Auto recording" or "Auto recording: comment from autorec rule".
This helps to identify why a recording is scheduled.

This patch ensures we continue to use the autorec comment field if
it is non-empty (keeping existing behaviour), but fallback to using
the recording rule name. If both are empty then we keep the
existing behaviour of fallback to an empty string.

This avoids the user having to duplicate the rule name in to
the comment string for manually created autorec rules.

So, in the above case it would be "Auto recording: comment",
"Auto recording: rule name", otherwise "Auto recording".

Issue: #4500.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -1463,9 +1463,11 @@ dvr_entry_create_by_autorec(int enabled, epg_broadcast_t *e, dvr_autorec_entry_t
     }
   }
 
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

PR #992, opened 2017-09-09
eit: scrape sub-title for uk freeview freesat (#4578)

azlm8t:
The Freeview/Freesat frequently have a subtitle as part of the                                                                                                                                                                             
summary. So we have "Treehouse of Horror IX: Three scary                                                                                                                                                                                   
stories."  from which we can deduce the subtitle as "Treehouse of                                                                                                                                                                          
Horror IX".
                                                                                                                                                                                                                 
So allow scraping of this and use it as the subtitle. If we                                                                                                                                                                                
cannot scrape a subtitle then we continue the existing practice                                                                                                                                                                            
of using the summary buffer for the subtitle.                                                                                                                                                                                              
                                                                                                                                                                                                                                           
The subtitle is currently NOT removed from the summary. I experimented with removing it but the logic for tidying up the summary was becoming ugly for relatively little benefit.

This scraping is configurable via Configuration->Channel->EPG Grabber Module->Scrape Subtitle.
This ensures that people who rely on the existing dup detect for subtitle are not affected by the change unless they specifically opt-in.

I also added a bit more clarification to the documentation of scraping to indicate that they don't access the Internet, and fixed up the regex so they are parsable by JSON for the upcoming test harness.
                                                                                                                                                                                                                                           
Issue: #4578


perexg:
Merged. Thanks.

PR #993, opened 2017-09-09
EIT scraper basic test harness (#4577)

azlm8t:
(Please leave bug open for further commits as people work out their scraper settings, but if code is acceptable then commit, thanks).

This is an initial implementation of a simple Python test harness for the EIT scraping behaviour. This is done to make it easier to verify changes going forward and make it easier for non-programmers to add test cases for scrapers.

It takes the Tvheadend scraper configuration file as input and a JSON test harness file containing data similar to a broadcaster's summary and the season/episode we expect to scrape from it. It then just outputs a simple OK or FAIL for each case.

I'm hoping the syntax is easy enough that non-programmers can write examples of what they expected from what they see in the EIT.

I've added the Bulsatcom_39E test harness file but have not included the configuration file since it is still being discussed in #4509, but committing the test harness file will allow the two other people in that thread to hopefully test their local changes to the config file more easily.

The uk test harness file only works against the uk Tvheadend file in #4578 since Python had problems parsing the file. I've not added it to this patch since there would be merge conflicts.

### DIFF
--- a/data/conf/epggrab/eit/scrape/README
+++ b/data/conf/epggrab/eit/scrape/README
@@ -1,0 +1,6 @@
+The directory contains configuration file for general regular
+expressions to be applied to the EPG.
+
+There is a test harness for these files in the development tree at
+support/eitscrape_test.py with test harness files at
+support/testdata/eitscrape.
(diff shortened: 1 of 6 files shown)


Saentist:
Let's rename "Bulsatcom_39E" to "bg"
it apply to almost every Bulgarian DVB-T/C/S operators

azlm8t:
Thanks, I've applied those changes.

perexg:
Merged. Thanks.

PR #994, opened 2017-09-18
doozer: 'fix' dirty build issue (4).

mpmc:
from (but modified): https://stackoverflow.com/questions/24533390/git-reset-files-with-typechange-status#27461055

Update: CentOS has issues with this apparently :( https://doozer.io/mpmc/tvheadend/build/1255

Feel free to modify/fix (I have no idea why it won't work on centos).


perexg:
Merged. Thanks.

PR #995, opened 2017-09-18
Travis: build and publish targets including those not currently supported by doozer

mpmc:
This basically uses Docker on Travis-CI to build various targets, including some not currently supported by Doozer, It's mainly as a backup.

I've been testing this for a while & it seems to work okay, it's not the nicest in the least and may need some fixing, especially the GIT_BRANCH "hack" I added which doesn't work right yet, but that can be ignored (or even stripped out).

The uploads are done in docker, so the ENV variables [BINTRAY_USER, BINTRAY_PASS] etc need to be set (securely) in Travis, See the docker run line for the rest. These are slow builds & I mean slow, but that should be fine as a backup as the upload fails with 409 conflict if Doozer uploads first.

---
This can be rejected if not up to standard, or used as people see fit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,80 @@
-#
-# Basic Build Setup
-#
-
-sudo: required
-dist: trusty
 language: c
+sudo: required
+services:
+  - docker
+cache: ccache
+
+env:
+  global:
+    - GIT_BRANCH: "test-builds"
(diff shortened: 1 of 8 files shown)


mpmc:
I don't think this is needed anymore - closing for now :)

PR #998, opened 2017-09-21
eit: Also scrape episode from title and description (issue #4592)

azlm8t:
Previously we scraped the EIT episode information only from the summary. However several countries put this information in the title or description. So we now apply the same regex to these fields.
This is done my moving the existing logic to a separate function and calling it for each field.

I posted a variant of this patch that has been successfully tested for a few days in #4287 by Dietmar Konermann. But I modified the patch slightly since then due to a merge conflict and to rename the enums used for the changed flags.

Also added config for scraping Italy, NZ, and AU based on the opentv config and bug discussion.


perexg:
Merged. Thanks.

PR #999, opened 2017-09-21
fix storage space calculation

razzfazz:
The f_bfree and f_bavail members in struct statvfs are expressed in
units of f_frsize, not f_bsize (see the statvfs(2) manpage, e.g. here:
<https://linux.die.net/man/2/statvfs>). The two can differ quite
substantially, e.g. on ZFS.


perexg:
Merged. Thanks.

PR #1002, opened 2017-09-23
xmltv: Parse atsc style numbers in channel tag. (#4615)

azlm8t:
Previously we only parsed numbers of the format "445" and
forced the minor number to be zero. However atsc numbers
have a major and a minor so were not parsed.

We now parse "39.1" as major=39 minor=1 so we handle
and correctly associate channel sub-tags such as
 `<display-name>39.1 KNSDDT</display-name>
`
This was tested in the forum thread since I don't have atsc.

One change in behaviour is that we now expect channel number
to be non-zero which I believe to be ok.

Issue: #4615


perexg:
Merged. Thanks.

PR #1003, opened 2017-09-24
eit: Add drop-down list for selecting xmltv scraper configuration. (#4626)

azlm8t:
Suggested by Mark Clarkstone in 4509, previously you had to 
enter text in the "scraper configuration to use" drop-down on 
the EPG grabber modules/Over the air module.

Now we populate a drop-down selection box by loading the configurations
and then getting their names. This is slight overhead compared to just
fb_scandir, but it is only used briefly during GUI configuration and
has the benefit of listing configurations that are valid in case the
user modified the configuration files and caused it to be invalid.

Issue: #4626

### DIFF
--- a/src/epggrab/module.c
+++ b/src/epggrab/module.c
@@ -104,6 +104,34 @@ static int epggrab_mod_class_type_set(void *o, const void *v)
   return 0;
 }
 
+static htsmsg_t *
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

PR #1004, opened 2017-09-25
xmltv: Only parse channel number from display-name if config allows. (#4615)

azlm8t:
I'm sorry but my earlier commit (a803e6, parse atsc numbers) accidentally removed the check for user configuration option (chnum) when parsing the display-name in xmltv files. This means that leading digits in an xmltv display-name would always be stripped even if the config is disabled.

The patch is actually just re-adding the test and moving the variables inside the test, but the re-indent shows most lines changed.

I used the same commit issue as the original bug (for parsing atsc numbers).


perexg:
Merged. Thanks.

PR #1005, opened 2017-09-27
webui: fix clear all stats button thinko

mpmc:
This explains why the button didn't work! DOH!


Trujulu:
Tested! With this commit works!!! 😀

mpmc:
@Trujulu Thanks for the confirmation! I noticed it when I installed it on my main tvh system, initially I thought it was an error with Chromium but then checked other browsers/OSes and they had the same problem.

This goof was entirely my fault :disappointed: !

perexg:
Thanks. Applied.

PR #1006, opened 2017-09-28
webui: Mark upcoming recording duplicates with a line-through in grid (#4632)

azlm8t:
The advanced and expert views on the upcoming tab have a column for the
duplicate date ("rerun of"). But on the basic view there is no indication of
duplicates unless you click information on the entry.

So we now render title/subtitle for duplicates with a css style that has line-through.
This puts a strikeout line through the title/subtitle and makes it clear they are duplicates
that will not be recorded.

Issue: #4632

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -195,6 +195,21 @@ tvheadend.filesizeRenderer = function(st) {
     }
 }
 
+/** Render an entry differently if it is a duplicate */
+tvheadend.displayWithDuplicateRenderer = function(value, meta, record) {
+    return function() {
+        return function(value, meta, record) {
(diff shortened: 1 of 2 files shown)


perexg:
Applied. Thanks. Perhaps, we can also change the first (left) icon.

azlm8t:
I was thinking the same thing about the icon.

I think the iconIndex bit in the GUI code is probably taking the string sched_status and automatically applying that class from the CSS to get the icon to display. So, it's getting DVR_SCHEDULED but should perhaps get DVR_NOSTATE, but that might have a knock-on when checking duplicate start. Perhaps there's a better way to modify the icon displayed client-side. I'll keep looking to see if I can find an example.

PR #1007, opened 2017-09-29
opentv: opentv provider skyIT support new formats

anthonytex:
Recently start to appear new description format so new regex are needed to correctly scarpe Episode/Season number and subtitle.


perexg:
Applied. Thanks.

PR #1010, opened 2017-10-02
doozer.io: add raspbian jessie and stretch builds

mpmc:
Doozer can now build for jessie and stretch. This adds those in.

~~Additionally, commit 049cba99157da175163f9c6a6fa8a686f3fafa54 needs to be applied to release/4.2 to avoid the dirty tagging.  it may need modifying to fit with 4.2 doozer.json, I'm happy to do it if required.~~ Fixed in doozer so no longer required see [IRC log](http://colabti.org/irclogger/irclogger_log/hts?date=2017-10-05#l192)

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -239,6 +239,58 @@
         "support/bintray.py publish filelist.txt"
       ]
     },
+    "raspbian-jessie": {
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

PR #1012, opened 2017-10-13
dvr: allow autorec by category (#4665)

azlm8t:
The xmltv import supports categories such as "movie", "animated", "biography", so allow autorec to record via these categories. Also allow recording with a minimum star rating.

This allows you to record all movies with a rating of 80% that are animated; or all sumo wrestling, etc. and can be combined with the existing regex to filter matches further such as "all movie, comedy which match full text of Stallone".

I've not bumped the HTSP_PROTO_VERSION since I have another feature I'll submit tomorrow and then increment the version for both patches so clients know the feature is available rather than having merge conflicts on both patches changing the same version number. I believe this patch could be merged even without incrementing the version, it just means frontends won't yet know the backend supports recording by category.

### DIFF
--- a/src/api/api_channel.c
+++ b/src/api/api_channel.c
@@ -24,6 +24,7 @@
 #include "channels.h"
 #include "access.h"
 #include "api.h"
+#include "string_list.h"
 
 static int
 api_channel_is_all(access_t *perm, htsmsg_t *args)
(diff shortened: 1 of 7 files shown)


perexg:
Merged. Thanks.

PR #1014, opened 2017-10-13
dvr: Add format specifier to simplify scraping/separate movies and shows (#4667)

azlm8t:
Add a new format specifier "$q" to generate "queryable" (scrapable) names for recordings. This relies on good guide data, typically xmltv data that contains "movie".

This format specifier helps split movies and shows in to separate hierarchies without needing post-processing and makes it easier for users who want easy setup.

Examples are:
- tvmovies/Gladiator (2000)
- tvshows/Countdown/Countdown
- tvshows/Bones/Bones - S05E11 - The X in the Files

A couple of other modifiers are also added for cases where guide data is not good so recordings can be forced to be movie or tvshow for cases where a user has separate recording profiles for movies and shows.

I'm not too happy with the names for the variants $1q, $2q for forcing a programme to be formatted as a movie or show. I chose those names to make it easy to extend in the future to have other variants as necessary ($3q, $4q) and to avoid any potential confusion with other specifiers. For example having "$tq" meaning tvshow queryable would be confused with "$t" for title followed by letter q; using $TQ would then remove another letter from future specifiers. So $1q was a compromise.

### DIFF
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -472,6 +472,7 @@ _dvr_sub_scraper_friendly(const char *id, const char *fmt, const void *aux, char
    */
 
   size_t offset = 0;
+  const dvr_config_t *config = de->de_config;
 
   if (is_movie) {
     /* TV movies are probably best saved in one folder rather than
@@ -485,7 +486,12 @@ _dvr_sub_scraper_friendly(const char *id, const char *fmt, const void *aux, char
      *   "title (yyyy)"                     (without genre_subdir)
      *   "title"                            (without genre_subdir, no airdate)
      */
-    if (with_genre_subdir)   tvh_strlcatf(tmp, tmplen, offset, "tvmovies/");
+    if (with_genre_subdir) {
+      const char *subdir = config && config->dvr_format_tvmovies_subdir && *config->dvr_format_tvmovies_subdir ?
+        config->dvr_format_tvmovies_subdir : "tvmovies";
+      tvh_strlcatf(tmp, tmplen, offset, "%s/", subdir);
+    }
+
     if (*title_buf)          tvh_strlcatf(tmp, tmplen, offset, "%s", title_buf);
     /* Movies don't have anything relevant in sub-titles field so
      * anything there should be ignored. I think some channels store a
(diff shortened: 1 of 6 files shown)


src/dvr/dvr_rec.c
```diff
@@ -356,6 +357,179 @@ dvr_sub_episode(const char *id, const char *fmt, const void *aux, char *tmp, siz
...
+  const char *subtitle = lang_str_get(de->de_subtitle, NULL);
+  const char *desc     = lang_str_get(de->de_desc, NULL);
+
+  if (subtitle && desc && strcmp(subtitle, desc) == 0) {
+    /* Subtitle and description are identical so assume they are from
+     * bad OTA EIT.  Some OTA EIT often has a (long) summary which is
+     * put in to both subtitle and description. So we really don't
+     * want this to be used as the subtitle field.
+     */
+    subtitle = desc = NULL;
```
perexg: I think that here should be only 'subtitle = NULL'.

src/dvr/dvr_rec.c
```diff
@@ -356,6 +357,179 @@ dvr_sub_episode(const char *id, const char *fmt, const void *aux, char *tmp, siz
...
+     * artwork and information for them anyway and it makes deleting
+     * and moving them easier since they get tracked by inotify on
+     * just the one directory.
+     *
+     * Example format below:
+     *   "tvmovies/title (yyyy)"            (with genre_subdir)
+     *   "title (yyyy)"                     (without genre_subdir)
+     *   "title"                            (without genre_subdir, no airdate)
+     */
+    if (with_genre_subdir)   tvh_strlcatf(tmp, tmplen, offset, "tvmovies/");
```
perexg: 'tvmovies' should be probably configurable through 'DVR Profiles'.

src/dvr/dvr_rec.c
```diff
@@ -356,6 +357,179 @@ dvr_sub_episode(const char *id, const char *fmt, const void *aux, char *tmp, siz
...
+     * We put the episode number before the subtitle to make it easier
+     * to see if we are missing episodes when you do ls.
+     *
+     * Example formats below:
+     *   "tvshows/title/title - S01E02 - subtitle" (with genre_subdir)
+     *   "title - S01E02 - subtitle"               (without genre_subdir)
+     *   "title - subtitle_2001-05-04"             (without genre_subdir, long running show)
+     *   "title - subtitle"                        (without genre_subdir, no epg info on show)
+     */
+    if (with_genre_subdir) tvh_strlcatf(tmp, tmplen, offset, "tvshows/");
```
perexg: 'tvshows' should be probably configurable through 'DVR Profiles'.

azlm8t: Good idea. I've taken a quick look and it should be easy enough to do since we have access to the config via the aux variable. I'll make the change in a couple of days.

src/dvr/dvr_rec.c
azlm8t: Agreed. The desc isn't used after that point but it makes sense to leave it non-null.
I'll make the change.

azlm8t:
I've delivered the changes to make the tvshows/tvmovies directory names configurable.

The build failed for an unrelated reason since it could not fetch the build packages. I don't think I can restart the build to verify the merge.

perexg:
Merged. Thank you for your work. I added little changes to cdd35c8cdf670738e1c94ac158583caf99b283e7 .

azlm8t:
Thanks, the changes look good and looks more intuitive with a def.s.

PR #1015, opened 2017-10-14
dvr: add new dedup method by programid (#4652)

azlm8t:
This patch implements a new dedup method based on using the unique id from xmltv or the crid from OTA.

The idea is that if the ids are equal then the programmes must be equal. So an xmltvid of MV12345 will always refer to the same movie.

If the entries don't have equal ids then we try and compare various fields for equality to try and just work. So, if they have season and episodes then we can use that as a differentiator.

This dedup method only work for unique programmes, so sports and news typically have non-unique details and so cannot be deduped with this method and the user needs to use existing "once per day" methods.

I've also bumped the htsp_protocol_version so clients know the dedup method is supported.

This patch has mostly been tested with xmltv data.

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -203,6 +203,7 @@ typedef struct dvr_entry {
   char *de_owner;
   char *de_creator;
   char *de_comment;
+  char *de_uri;                 /* Programme unique ID */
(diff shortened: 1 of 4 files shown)


perexg:
Merged. Thanks.

PR #1016, opened 2017-10-15
WebUI: Split EPG details (broadcast details, DVR status) and actions …

ProfYaffle:
…(play) into separate columns

Purely selfish - I've never liked the wall of "play" buttons you get in the EPG, especially when I never use them unless I click them by accident. 

This PR simply splits the first column into two, then, so you have "Details" in one and "Actions" (currently only "Play", but there's room for future actions as well) in the second. You can then hide the "Actions" as required.

@perexg Some advice on the resizing of the second column, please - I'll add a comment to the code.

### DIFF
--- a/src/webui/static/app/epg.js
+++ b/src/webui/static/app/epg.js
@@ -417,7 +417,7 @@ tvheadend.epg = function() {
         new tvheadend.VideoPlayer(item.data.channelUuid);
     };
 
-    var actions = new Ext.ux.grid.RowActions({
+    var eventdetails = new Ext.ux.grid.RowActions({
         id: 'details',
         header: _('Details'),
         tooltip: _('Details'),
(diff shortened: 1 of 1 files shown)


src/webui/static/app/epg.js
```diff
@@ -1050,7 +1060,7 @@ tvheadend.epg = function() {
                    state.filters = {};
                    /* Otherwise this non-resizable column will not expand to newly set width */
                    if (state.columns)
-                       state.columns[0].width = actions.width;
+                       state.columns[0].width = eventdetails.width;
```
ProfYaffle: I wasn't 100% certain on this one - I've renamed the object, so all works, but I wasn't certain whether we should do the same with eventactions.width. I certainly couldn't get it to do anything, but my Javascript isn't up to working out why/how (or even whether...).

perexg:
Merged. Thanks.

PR #1018, opened 2017-10-20
Add dvr disk space function

dave-p:
At present there is no way to view the dvr disk utilisation through the HTTP API. This PR allows the free and used disk space to be queried.

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -508,6 +508,20 @@ api_dvr_entry_file_moved
   return 0;
 }
 
+static int
+api_dvr_diskspace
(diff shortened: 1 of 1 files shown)


perexg:
Actually, there's issue that there can be multiple directories for recordings on different mounts. TVH does not actually handle this so I don't like to add any other APIs until we support multiple mounts. But, please, keep this request open.

dave-p:
DVR disk utilisation can be obtained from the comet/poll endpoint (https://github.com/dave-p/TVH-API-docs/wiki/Other-Functions#cometpoll) so this PR is not necessary.

PR #1019, opened 2017-10-22
ui: Send episode image URL to UI for upcoming recordings. (#4681).

azlm8t:
Previously the image was only visible in the EPG tab. However it is useful to have it in the upcoming recordings tab too.

Issue: #4681


perexg:
Merged. Thanks.

PR #1020, opened 2017-10-22
dvr: Allow scheduled programmes to be stopped (never record). (#4506)

azlm8t:
Previously you could not stop an upcoming scheduled programme from
recording.

You could delete it, but it would only unschedule that showing and
restarting tvheadend would cause it to be scheduled again, and any
future showings would also be recorded.

So we now allow the user to "stop" an upcoming recording to mark
that the user does not want that programme or any future showings
of the programme to be recorded.

We do this by marking it as "file removed". This allows you to unschedule
a programme and not have it record. This is useful if an autorec
rule is matching programmes you don't want to record such as episodes
you've already seen.

Once in the file removed section you can re-record it if necessary
by deleting it from the file removed which then causes the autorec
rule to re-match the recording.

The extra check for de->de_s is because we don't have a subscription
stream for upcoming recordings so can't unsubscribe.

Issue: #4506

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -2098,7 +2098,7 @@ dvr_stop_recording(dvr_entry_t *de, int stopcode, int saveconf, int clone)
   dvr_rs_state_t rec_state = de->de_rec_state;
   dvr_autorec_entry_t *dae = de->de_autorec;
 
-  if (!clone)
+  if (!clone && de->de_s)
     dvr_rec_unsubscribe(de);
 
   de->de_dont_reschedule = 1;
@@ -3669,7 +3669,11 @@ dvr_entry_move(dvr_entry_t *de, int failed)
 dvr_entry_t *
 dvr_entry_stop(dvr_entry_t *de)
 {
-  if(de->de_sched_state == DVR_RECORDING) {
+  /* Allow stopping a scheduled item. This will then mark it as file missing.
+   * This is useful if you have auto_rec rule that matches programmes you
+   * don't want to record such as episodes you've already seen.
+   */
+  if(de->de_sched_state == DVR_RECORDING || de->de_sched_state == DVR_SCHEDULED) {
     dvr_stop_recording(de, SM_CODE_OK, 1, 0);
     return de;
   }
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -268,11 +268,14 @@ tvheadend.dvr_upcoming = function(panel, index) {
 
     function selected(s, abuttons) {
         var recording = 0;
+        var scheduled = 0;
         s.each(function(s) {
             if (s.data.sched_status.indexOf('recording') == 0)
                 recording++;
+            if (s.data.sched_status.indexOf('scheduled') == 0)
+                scheduled++;
         });
-        abuttons.stop.setDisabled(recording < 1);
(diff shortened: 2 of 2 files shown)


perexg:
There's "enabled" field for each DVR entry which causes "Service not enabled" error when the recording is going to be started. Perhaps, the real fix might be to ask the user when he clicks on "Stop" button, if he wants to turn off the future recordings.

azlm8t:
The problem is that I want my autorec recording rule to still be active, but want specific recordings that it matches to never be recorded.

For example, say I have a rule to record action film rated 80% or better. But it matches films I've already seen in the past and don't want to see again soon (such as Die Hard), or films I'm just not interested in. So I need to say "never record Die Hard since I don't want to watch it again", but I also need the autorec to continue to match "Die Hard 8".

I know it's an odd way of recording (rather than looking through guide and saying "record series", or "record this"), but I use it as a "you might also like..." feature so I have programmes recorded that might interest me and means I only need to look through a few upcoming tv entries rather than the whole tvguide.

I initially just hit "delete" on the entries, but I restart Tvheadend every day and the programmes would then get re-matched.

azlm8t:
Thinking about it, perhaps what we need in addition to this patch is an additional button (either on the toolbar or on the info dialog) for "disable autorec rule". The button would be enabled if autorec field is not empty. That would satisfy the requirement for disabling all upcoming recordings for a rule.

Or perhaps a new button on the upcoming info dialog could be "Autorec rule" and pop-up the autorec dialog so all settings can easily be modified such as autorec priority. I'm not sure if it's easy to do cross-panel calls in the webui for the upcoming panel to open the dialog from the autorec panel.

perexg:
There's logic in the autorec code which should preserve the 'enable' field for re-match (based on the EPG event check).

azlm8t:
I've just tried it but I don't think that works for my scenario.

I tried it on a programme which is scheduled at 00:40 and repeated on a +1 timeshift channel at 01:40.

On the upcoming tab, I unticked the enabled button and hit save and the dvr/log is updated with enabled: false for the 00:40 entry. However, when I reloaded the tab then the programme scheduled on the +1 channel is then shown as being scheduled (rather than having the line through it as being a duplicate of the 00:40 entry).

Looking at timestamps in dvr/log, the entry at 00:40 was updated, but the other entries that matched the recording rule aren't updated. I'll have to wait for the record time to be reached, but I think it will record the 01:40 showing since it's not a duplicate of an existing recording.

So, the enabled flag in the upcoming tab's left hand column I believe is permanently disabling a specific showing, but Tvheadend will still record some other match for that rule and the same episode but will exclude that one specific showing and that will persist over a restart.

This is in contrast to the upcoming tab's "delete" button which (IIRC) removed the entry completely, didn't disable it, but then on a Tvheadend restart it would then be rescheduled.

And is different to this patch which aims to stop both the showing at 00:40 and the showing at 01:40 and future repeats from recording.

perexg:
Merged. Thanks.

perexg:
@azlm8t : If you like, you can join to our IRC channel #hts . Let me know, if you are interested to co-maintain the tvh's code.

azlm8t:
Thanks. I don't think I can possibly keep up with the amount of spare time you dedicate, nor your fast turnaround of bugs. But yes, I'd be glad to help out if I can with maintaining the code.

perexg:
I believe that we resolved this issue with a bit different way - the 'previously recorded' state. Closing.

PR #1028, opened 2017-11-03
Persist image url so it is available for recorded programmes (#4685)

azlm8t:
The Kodi pvr.hts currently currently uses a channel icon for recorded programmes. Since some xmltv providers supply an image icon, we now persist this URL so it can be sent to the client for recorded programmes.

Example htsp message:
`'image': 'https://s3.amazonaws.com/schedulesdirect/assets/p8648012_d_v8_aa.jpg',                                                                                                                
'method': 'dvrEntryAdd',
`
I changed the image property to just be HIDDEN, whereas previously it had NOUI, NOSAVE and RDONLY. The other removals I think are correct, but I'm not sure about the removal of RDONLY. It could (presumably) in the future make it possible for the user at the UI to alter the image used such as using a better poster image, so I think removing RDONLY is the right thing.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -3295,16 +3295,23 @@ static const void *
 dvr_entry_class_image_url_get_as_property(void *o)
 {
   dvr_entry_t *de = (dvr_entry_t *)o;
+  static const char *s = "";
+  /* We prefer the image from the broadcast to the one currently
+   * persisted.  This is because a programme scheduled far in the
+   * future may have a generic image that will be updated nearer the
+   * broadcast date with a more specific image.
+   */
+  if (de->de_bcast && de->de_bcast->episode && de->de_bcast->episode->image) {
+    snprintf(prop_sbuf, PROP_SBUF_LEN, "%s", de->de_bcast->episode->image);
+    return &prop_sbuf_ptr;
+  }
+
   if (de->de_image) {
     prop_ptr = de->de_image;
     return &prop_ptr;
   }
 
-  static const char *s = "";
-  if (!de->de_bcast || !de->de_bcast->episode || !de->de_bcast->episode->image)
-      return &s;
-  snprintf(prop_sbuf, PROP_SBUF_LEN, "%s", de->de_bcast->episode->image);
-  return &prop_sbuf_ptr;
+  return &s;
 }
 
 const char *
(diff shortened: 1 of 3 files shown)


src/dvr/dvr_db.c
```diff
@@ -3479,8 +3499,9 @@ const idclass_t dvr_entry_class = {
       .id       = "image", /* Name chosen to be compatible with api_epg */
       .name     = N_("Episode image"),
       .desc     = N_("Episode image."),
-      .get      = dvr_entry_class_image_url_get,
-      .opts     = PO_HIDDEN | PO_RDONLY | PO_NOSAVE | PO_NOUI,
+      .get      = dvr_entry_class_image_url_get_as_property,
+      .off      = offsetof(dvr_entry_t, de_image),
+      .opts     = PO_HIDDEN,
```
perexg: If you remove PO_NOSAVE then de_image will be set / loaded when tvh restarts. I think that there should be a logic similar to de_channel_name (so if event is available - take info from event, otherwise use cached value). And note that de_channel_name is PO_RDONLY. I just like to keep things a bit consistent.

azlm8t: Thanks.

OK. I'll add back in the read-only to make it `PO_HIDDEN | PO_RDONLY` since we want it reloaded on tvh restart, and leave out the `PO_NOSAVE | PO_NOUI`.

There exists similar logic to de_channel_name in the `dvr_entry_class_image_url_get_as_property`, but it's the opposite. My thinking is that in the future the user may want to replace the (xmltv) image on some recorded shows that they want to keep with some other image url they prefer from some fanart site. I saw some bug saying "allow user to edit season/episode" so my thinking was this might be useful for the image too. (Though not implemented, but if implemented we'd then presumably have to remove `PO_HIDDEN` and `PO_RDONLY`).

Whereas for channel name you always want to use the current channel name instead of the cached and that makes sense to have the opposite logic.

Let me know if you want me to swap the logic over, otherwise I'll submit just the `PO_HIDDEN | PO_RDONLY`.

azlm8t: I've been thinking about this and I think we should use the de_channel_name logic. Otherwise an xmltv provider could say "in 14 days some episode of X will be shown" and have a generic image which would then be stored against the de_image, but then nearer the time the xmltv might replace the image with a more-specific episode image, but we would keep using the generic image since it is already stored in the dvr entry.

I guess this scenario could happen with chat shows that the broadcaster records a day-or-so before broadcast where perhaps the episode image might be updated to be a picture of the people on the show.

So, let's make it read-only and prefer the image from de_bcast.

I'll make the change in the next couple of days.

perexg: Thanks.

azlm8t:
Updated to prefer the image from the current broadcast and to add `PO_RDONLY` to the de_image.

perexg:
Merged. Thanks.

PR #1029, opened 2017-11-03
dvr: allow inotify to track file moves between watched directories (#4702)

azlm8t:
Currently the inotify rename correctly updates the dvr/log if the filename is in the same directory. 

However, if the file is moved to a different watched directory then we currently use from_wd instead of to_wd. This means the directory component isn't updated since it always believes the two directories to be identical.

With this patch, if the "to" directory is already watched, then the directory component in dvr/log is also updated.

I've also added support to use realpath when tracking files since inotify always returns a realpath even if the path in the dvr/log passes through symlinks. This meant that some compares did not correctly work since the dvr/log would be "/a/b/c.ts" but realpath might be "/x/y/c.ts".

I think my newly added "return" at line 309 is correct: we failed to find the "to" directory so nothing to do since the entry can't be updated, and it is a mirror of the existing functionality at line 249 where we return if the "from" directory could not be found.

### DIFF
--- a/src/dvr/dvr_inotify.c
+++ b/src/dvr/dvr_inotify.c
@@ -112,13 +112,13 @@ static void dvr_inotify_add_one ( dvr_entry_t *de, htsmsg_t *m )
   dvr_inotify_filename_t *dif;
   dvr_inotify_entry_t *e;
   const char *filename;
+  char path[PATH_MAX];
   int fd = atomic_get(&_inot_fd);
 
   filename = htsmsg_get_str(m, "filename");
   if (filename == NULL || fd < 0)
     return;
 
-  char path[PATH_MAX];
   /* Since filename might be inside a symlinked directory
    * we want to get the true name otherwise when we move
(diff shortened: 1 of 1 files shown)


src/dvr/dvr_inotify.c
```diff
@@ -249,8 +259,34 @@ _dvr_inotify_moved
...
+         * and compare that.
+         *
+         * This is made far more complicated since the
+         * file has already disappeared so we can't realpath
+         * on it, so we need to realpath on the directory
+         * it _was_ in, append the filename part and then
+         * compare against the realpath of the path we
+         * were given by inotify.
+         */
+        char *dir = tvh_strdupa(filename);
```
perexg: Could you move all declarations on top of the block (or ideally function) ?

azlm8t:
Not a problem. Moved the variables to top of function/block.

perexg:
Merged. Thanks.

PR #1030, opened 2017-11-05
channel: Allow optional fuzzy matching when merging services. (#4709)

azlm8t:
For historical reasons, our DVB-T and DVB-S have different names
for the same channels. They often differ in case and spacing.
So we have 'One' and 'ONE', '5+1HD' and '5 +1HD'.

So allow an optional fuzzy matching checkbox to ignore whitespace
and HD markers. This allows the channels to be merged. The
exact name chosen depends on the order of mapping, so if the
HD channel is mapped first then they would all merge in to this
name, but if a non-HD channel is the first one created then that
name is chosen. However the name could be subsequently modified
by the user if they desire.

This also ensures merging of channels on DVB-T or DVB-S will
allow automatic merging of HD and non-HD channels ("5" and "5HD")
to allow the "preferred service video type" to work without needing
manual merging by the user.

Issue: #4709

### DIFF
--- a/src/channels.c
+++ b/src/channels.c
@@ -579,6 +579,68 @@ channel_find_by_name ( const char *name )
   return ch;
 }
 
+
+/// Copy name without space and HD suffix, lowercase in to a new
+/// buffer
+static char *
+channel_make_fuzzy_name(const char *name)
+{
+  if (!name) return NULL;
+  const size_t len = strlen(name);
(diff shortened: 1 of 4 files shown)


perexg:
Merged. Thanks. Perhaps, you can also add 'UHD' check 👍

azlm8t:
Good idea about UHD. I'll add it in at the weekend.

PR #1032, opened 2017-11-16
configure: Fix libsystemd dependency for Stretch and later

samnazarko:
Debian Stretch now has libsystemd instead of libsystemd-daemon. 
This configure update will allow building for older Debian targets as well as newer ones. This has been tested on Debian Jessie and Debian Stretch.

Signed-off-by: Sam Nazarko <email@samnazarko.co.uk>

### DIFF
--- a/configure
+++ b/configure
@@ -730,7 +730,7 @@ fi
 # systemd
 #
 if enabled_or_auto libsystemd_daemon; then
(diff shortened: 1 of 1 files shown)


perexg:
Merged. Thanks.

PR #1033, opened 2017-11-16
channel: fuzzy merge channels in bouquet, tidy names, new user api call, fixes #4714, #4715

azlm8t:
The fuzzy merge of channels names was previously implemented on the map services. This adds the functionality to bouquets. It also adds an option to "tidy" the channel name for automatically stripping HD/UHD suffix so "History HD" will just be called "History".

As far as I could tell, the existing "merge bouquet channel by name" always explicitly created a new channel if bq, so never actually merged. So I added a new function to search for channels on a bouquet and allow merging with that channel on the same bouquet. I don't allow a channel to be merged across bouquets since we'd have to update to have a list of bouquets for channels, and each bouquet could have conflicting mapping options, so I'll wait to see if people need that functionality.

I added a new user-convenience API call to allow users to easily rename channels from an external script. This is based on the dvr filemoved code. This new call allows users to fix-up channels such as channel names that start with a lower-case letter so are sorted incorrectly in many UIs, or other fixups and avoids having large extra logic and numerous checkboxes in the core Tvheadend. I know they could use idnode/save but this allows them to rename without having to fetch a list of channels, parse ids, then form a valid message.

### DIFF
--- a/src/api/api_channel.c
+++ b/src/api/api_channel.c
@@ -96,6 +96,23 @@ api_channel_create
 }
 
 static int
+api_channel_rename
+  ( access_t *perm, void *opaque, const char *op, htsmsg_t *args, htsmsg_t **resp )
+{
+  const char *from, *to;
+  if (!(from = htsmsg_get_str(args, "from")))
+    return -EINVAL;
+  if (!(to = htsmsg_get_str(args, "to")))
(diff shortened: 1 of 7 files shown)


perexg:
Merged. Thanks.

PR #1036, opened 2017-11-22
debian: add a note to tvheadend.default explaining where to add/edit args

mpmc:
People seem to be getting confused by the systemd change, this should make it a bit more obvious :)

### DIFF
--- a/debian/tvheadend.default
+++ b/debian/tvheadend.default
@@ -6,6 +6,10 @@
 OPTIONS="-u hts -g video"
 
 # sysvinit
+# 
+# Editing the following variables has no effect when using systemd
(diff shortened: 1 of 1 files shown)


debian/tvheadend.default
```diff
@@ -6,6 +6,10 @@
 OPTIONS="-u hts -g video"
 
 # sysvinit
+# 
+# Editing the following variables has no affect when using systemd
```
ProfYaffle: * effect 😄

mpmc: Yep - I realised this mistake not long after committing & forgot to amend!

Thanks for the heads up!

perexg:
Mertegd. Thanks.

PR #1037, opened 2017-11-22
ui/dvr/epg: indicate new programmes in UI, with extra info in UI dialog (#4594)

azlm8t:
Adds a new column to indicate if a programme is new and to also contain icons for the programme type. For example, a new episode of Simpsons might display icons for "🆕 📺✏️😀", indicating new tv show, animated, comedy, which in my browsers are displayed as colourful pictures/emoji of the word new in a box, a tv, a pencil writing, and a laughing face. Some of the mappings are a bit arbitrary, but that's always unavoidable with mappings!

I had initially wanted these to float-right on the title, but it didn't seem compatible between different browsers since some floated it to the end of all columns and others just to the right of the title, so I added it all as a separate column before the title.

The mapping is done inside tvheadend.js. It's based on the top few category names in SD files I have. Other xmltv files I've seen for DK, NO, SE, FI all use English names for categories even though the descriptions are non-English, so it's possible that many grabbers use English for compatibility mapping to genres. So, I don't think those input strings should be subject to i18n, but perhaps we will need to add more categories in the future for other language mappings and maybe move it all to a separate file once it gets too big.

I send extra details to the UI for cast, credits, keywords, first_aired, copyright_year, genre; and renamed copyrightYear to copyright_year to be consistent between EPG and DVR.

I also fixed up the xmltv date parser since it was not correctly parsing the previously-shown dates since they were subset timestamps (YYYYMMDD) without HH:MM.

Titles for movies are now displayed as "Title (yyyy)" if the year is known.

A couple of UI fields I changed the CSS class since they were not wrapping in the dialog if the right-hand image was large. 

EPG and DVR Upcoming dialogs were updated to have a few more fields such as previously aired, and it tries to make it a nicer date as "today" or "yesterday."

Also updated the OTA scraper to have a new scraper for detecting new programmes and to populate the copyright_year instead of first_aired. I've left the first_aired variables there in case we populate it in the near future.


CvH:
@azlm8t could you add an screenshot ? :)

azlm8t:
DVR upcoming details dialog showing first aired and icons for comedy detective series
ss-dvr-details

EPG details dialog showing martial-arts action mystery and year in brackets:
ss-epg-details

EPG example with OTA genre rather than SD category and first aired previous day (since "yesterday" is a bit misleading when we talk about a programme in the future so I chose "previous day" instead):
ss-epg-genre

EPG for channels with categories also showing years in brackets and showing one film is "new". You can just about guess the categories such as cactus being for a Western.
ss-epg

azlm8t:
Sorry, I uploaded the wrong file for OTA genre. This is the one showing OTA genre instead of SD categories and showing "new" being parsed for the OTA channel:
ss-epg-ota-new

mpmc:
This looks awesome!

Trujulu:
@mpmc Yeah!!

perexg:
Merged. Really thanks for your work. Great extensions.

mpmc:
@azlm8t one thing I noticed (on mine) it adds "Parameters: New" even if the programme is an obvious repeat.

azlm8t:
@mpmc 
I don't see that on mine, I get "parameters: widescreen" for example on OTA.
I assume this is on EPG info dialog? Is it with the UK grabber? I assume the [new] icon isn't being displayed in the dialog or the grid, just the text at the bottom next to parameter.

The one change in that area is that we now request the 'new' column from the server, and of course we now regex for "New:" to set the flag, but if it's not displayed elsewhere then that logic should be ok. 

Perhaps we should rename the field 'new' to 'is_new' to avoid any potential conflict with keywords in case your browser is doing something different to mine I'll submit a patch later today if we don't find another cause.

If you have an example that fails then perhaps you could check to see if the server is sending "new" (it should only be sending it if an episode is new, and then it has the value 1).
`curl http://localhost:9981/api/epg/events/grid?limit=5000 | json_pp | less`

Thanks everyone for the comments.

azlm8t:
@mpmc 
Or if it's OTA if you let me know a programme that is on in a few days and fails and whether it's fv or fs then I can see if I have the same problem.

mpmc:
image

It's OTA only. I have all the UK grabbers enabled.

mpmc:
@azlm8t 

Thanks :)

azlm8t:
@mpmc 
OK, I think it may be the OpenTV sending different details. So I had this:
peston
Then after a few minutes more of scanning I had this and note how the "New:" for 22:45 disappeared.
peston-2
So, I think that may be related to why you see "new" on programmes that don't have "new".
However, I've not managed to get a run that looks like your dialog with S2Ep30. Mine all have "Robert Peston interviews the key political figures, newsmakers and cultural figures, while Allegra Stratton keeps viewers up to date with the big issues on social media. [S]"
I'll dig a bit more.

mpmc:
@azlm8t It's probably to do with my ordering as well?

azlm8t:
Quite likely. I'm not sure if you need OpenTV enabled. I'm still waiting for the epgscan to finish with a config with OpenTV but will probably have to retry tomorrow due to tuner conflict. I still haven't seen an instance with Peston with season and episode.

azlm8t:
@mpmc 
I've not managed to get any results with Peston with season/episode or the text you have, even with the same grabber priorities as you (eit:10, opentv:30, fs:20, fv:40; scrape episode/scrape subtitle enabled). Very odd.

But the following screenshot shows how if you un-merge the channels then you can see they have different ideas of subtitle. So if they are merged, you could get conflicting information with some thinking an episode is new and some thinking it is not-new.

Also it shows that some channels mark the 22:45 episode as new (although I assume it's a repeat of the 10am) which might explain why you see "new" for that episode in the info dialog.

peston-3
(Channel 4000+ I set up at T2 so I could see difference to S2)

So, I think it's probably just that you have merged channels and perhaps your region's DVB-T has the season/episode number but my DVB-T region doesn't include that information. I think the only way to know for sure would be if you unmerge your channels and, if that doesn't help, remove the is_new from the uk config file to see if programmes are still flagged with the new flag due to some browser difference.

azlm8t:
@mpmc 
I did some more testing of disabling grabbers. If you only have OpenTV enabled then I can get the Peston with season and episode. What happens is first it populates the titles and later it gets additional information such as description.

If I then enable dvb grabbers then the season/episode remains. But, if all are enabled at the start then I never get season/episode. If I restart with a clean database and have all the grabbers enabled with OpenTV higher priority that uk freesat then I don't get season/episode.

So there may be an issue with grabber priorities and merging/overwriting data since the OpenTV should be clearing the dvb eit information so I should see the same season/episode as you if I have the same priorities, whereas I imagine it's probably that you happen to grab OpenTV first.

This is grab with OpenTV only.
peston-4

mpmc:
Nope: this is my ordering.

Name | Priority 
---------|---------- 
EIT     | 1
UK Free**sat** | 2
OpenTV Sky UK | 3
UK Free**view** | 4

Has been for a while.

azlm8t:
@mpmc 
I noticed you have B&W icons. If you are running Ubuntu then you can install fonts-emojione to have colour icons.

I've submitted a new pull request to rename the 'new' to 'is_new' which I hope will get rid of the issue since I've tried two OSes and three browsers and can't reproduce it yet. The 'parameters new' logic was pre-existing (but not invoked since we didn't request the field) and it checked 'new' slightly differently to the way I check it to display the icon, so that might explain it.

The ordering I tested I believe is same as yours, but I used different numbers.

mpmc:
@azlm8t 

I'm using Chromium if that makes any difference (on Ubuntu Budgie). I also installed that font but it changed ALL numbers everywhere which were really hard for me to see, so had to remove it.

I'll be sure to update (when it's committed) and let you know how I get on :)

azlm8t:
@mpmc 
Just tried chromium on Ubuntu and it's ok here with Times New Roman; which fonts do you use? I've been testing Ubuntu a bit more and I noticed that Firefox displays some icons in colour if you have "serif" but if you choose "roboto" then it renders most in colour from the emojione font but a few in b&w such as heart and smiling face.

mpmc:
image

image

image

If I increase these too much I weird things happen (windows look off), even with these settings I have issues with seeing text sometimes.

azlm8t:
@mpmc 
Sorry, about the font. I noticed that the github page for Emojione has this warning "Note: This requires Bitstream Vera is installed and will change your systems default serif, sans-serif and monospace fonts.", something which the Debian package doesn't mention! So that may explain the numbers issue.

I've been chatting with ProfYaffle about replacing the font with icons in Tvheadend instead.

I tried chromium but, for me, it refuses to resize the fonts on Tvheadend. It resizes them for other websites such as BBC. It may be that it's obeying the "font-size: 12" that Tvheadend sends, or perhaps it's some weird interaction with something else. But, hopefully if we put images instead of a font then your B&W icons will disappear.

mpmc:
@azlm8t 

> Sorry, about the font.

Why are you apologising? It's not your fault I have crappy eyesight :). 

Please don't worry about fixing it, I can ignore it.

ProfYaffle:
My $0.02 - we can't go around changing system fonts, it really isn't in the spirit of user choice. It's also creating another dependency (both the emojione and Vera fonts), which perhaps isn't ideal either.

I'll have a look at some icons this week, see if there's anything obvious that starts to make sense across clients.

azlm8t:
I've got a proof-of-concept using png images from Google Noto fonts (Apache 2 licence). This is a grab using chromium and only the Google Noto pngs (no local font images):
chromium-noto

ProfYaffle:
With typical understatement, "yes, that looks like an improvement worth considering". I saw the Google icons as well, and their licensing seems to be most reasonable (c/f your comments on the other thread), so that's as good a place to start as any. 

There are a few notable famfamfam icons that could easily be incorporated if needed, but a quick glance suggests that Google has most things covered.

We should really shift the icon-specific conversation to that other PR, I think, as it's in danger of confusing your new/extra info thread here.

(PS back on topic - what scraper source are you using for the extra metadata? I'm not seeing anything in the OTA or xmltv ones I've tried)

azlm8t:
@ProfYaffle 
I use tv_grab_zz_sdjson_sqlite to grab Schedules Direct data. I have my own version that also populates the image xmltv tag with artwork for the programmes, but the original version grabs all the other extra metadata such as cast and rating. 

My version is here: [https://github.com/azlm8t/tv_grab_zz_sdjson_sqlite/tree/download-artwork](https://github.com/azlm8t/tv_grab_zz_sdjson_sqlite/tree/download-artwork)

Then in Tvheadend in the EPGGrabber Modules you click the "Scrape extra metadata" tickbox.

For OTA, the only scraping is minimal in the EPGGrabber Module for scraping subtitle, season/episode, and new episode indicator. There's no scraping yet for cast from OTA.

PR #1039, opened 2017-11-25
ui/epg: rename field 'new' to 'is_new', use genre if no category (#4594)

azlm8t:
I renamed the ui field 'new' to 'is_new' to avoid any potential confusion with keywords. This may fix the cause of the browser of @mpmc always showing 'parameter: new' in the dialog.

When displaying icons, if categories don't map to any icons then we try genre mappings to icons. Previously we would not do genre mappings if there was a category string even if the category string didn't match any icons.

### DIFF
--- a/src/webui/static/app/tvheadend.js
+++ b/src/webui/static/app/tvheadend.js
@@ -229,7 +229,7 @@ tvheadend.getContentTypeIcons = function(rec) {
   }
 
   var ret = "";
-  if (rec.new)
+  if ('new' in rec && rec.new)
     ret += "&#x1f195;";         // Squared New
   return ret + tvheadend.uniqueArray(ret_major).join("") + tvheadend.uniqueArray(ret_minor).join("");
 }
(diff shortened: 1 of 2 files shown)


src/webui/static/app/tvheadend.js
```diff
@@ -220,7 +229,7 @@ tvheadend.getContentTypeIcons = function(rec) {
   }
 
   var ret = "";
-  if (rec.new)
+  if (rec.is_new)
```
perexg: Please, keep 'new' to keep HTTP API compatibility and use _**if ('new' in rec)**_ here...

azlm8t:
OK, sounds good.

I've removed the rename 'new' to 'is_new' patch and force pushed a new one to check 'new' correctly in the tvheadend.js. I've also appended a second check of "&& x.new" since the epg store seems to force new to exist (so "in" check works) but it is an empty string and that would cause the icon to always be displayed for me.

perexg:
Merged. Thanks.

PR #1042, opened 2017-11-27
ui: replace category characters with png icons (#4594)

azlm8t:
Although I believed everyone would have fonts containing the characters for category icons, it seems several Ubuntu systems ship without font support.

So, this replaces the font characters with pre-rendered png images from the [Google Noto Color Emoji](https://www.google.com/get/noto/help/emoji) font. The images are Apache v2.0 licence and are apparently the ones Google uses for Gmail, etc. This means the images are consistent across the different browsers and OSes and should have updates going forward for new icons.

I have also increased the http expire timeout for jpg and png images from ten seconds to an hour since the images rarely change. The pre-rendered images are slightly large than the existing silk icons (~5k vs ~600b) but since they are cached it should make no practical difference and sending the image unscaled means they look better when displayed larger in the dialogs. I have not replaced any existing icons.

I've also added a hover to the icons to make them larger when you hover over them to make it clearer what the icons are.

I've put the few icons we use in vendor/noto (copied from the github repository for the font), symlinked them from our icon directory, and added the licence to the Tvheadend about dialog.

I have no problem at all if people want to replace the icons for specific categories with a different mapping, add/alter the mappings, or find better icons to use.

The EPG now looks like this:

epg-noto


Saentist:
Please add tool-tip on icons

perexg:
Merged. Thanks.

PR #1043, opened 2017-11-29
Fixes for FreeBSD and Darwin/macOS

macrule:
These are some fixes to make master compile on macOS (tested on 10.12) and FreeBSD (tested on 11.1).

The two major items:

1. A working implementation of wrappers.c for macOS that doesn't make the CPU spin needlessly due to a mixup of clocks in pthread_cond_wait().

2. Fix for a bug in the kqueue implementation that also caused needless busy-waiting on FreeBSD and macOS.

Taken those two together, tvheadend now idles at 0-5% CPU even when receiving a stream from SatIP.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -478,7 +478,7 @@ I18N-C += $(SRCS-AVAHI)
 # Bonjour
 SRCS-BONJOUR = \
 	src/bonjour.c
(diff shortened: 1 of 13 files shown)


perexg:
Merged all.

Trujulu:
Can you have a look? I think there is something weird with the descrambling in linux. (Errors on every key update, using oscam)

PR #1045, opened 2017-12-04
Help refresh (part 1)

mpmc:
Where to start..

* This hopefully cleans up the Help somewhat, & should make a nice base to add more information (some I've started)
* This took me a **long** time, I may have made some mistakes, please point them out, don't be afraid :)
* Usability: added a new toolbar and button - at the bottom of the Help window - to jump to the top of the doc (using a hidden div)
* Some screenshots still need updating
* Cleaned up the wizard pages
* Tweaked a few properties
* I've reordered the TOC a little so it isn't as long
* Moved all buttons info to one location - replaced the old button table in all docs (where it made sense) with an include pointing users to the button table's new location
* Tables.. tables.. EVERYWHERE. Because, why not?

Opinions, mistakes, anything are welcome!

Phew!


perexg:
Merged. Your work is really appreciated. Thank you.

PR #1047, opened 2017-12-08
Some improvements

pablozg:
I'm a newbie programmer and my programming level is very low, I'm sure that exist a better method to do this features, but I'm using them few days and they are working without issues.

Thanks for you great work with tvheadend.

New features:

- Now the autorec name is by default the epg title.
- A new button to show / hidde the skipped recordings in the webui.
- A new button to add as completed an upcoming recording to avoid record it again.

### DIFF
--- a/src/api/api_dvr.c
+++ b/src/api/api_dvr.c
@@ -82,7 +82,7 @@ api_dvr_entry_grid_upcoming
   dvr_entry_t *de;
 
   LIST_FOREACH(de, &dvrentries, de_global_link)
-    if (dvr_entry_is_upcoming(de))
+    if (dvr_entry_is_duplicate(de))
       idnode_set_add(ins, (idnode_t*)de, &conf->filter, perm->aa_lang_ui);
 }
 
@@ -267,6 +267,14 @@ api_dvr_entry_rerecord_toggle
   return api_idnode_handler(&dvr_entry_class, perm, args, resp, api_dvr_rerecord_toggle, "rerecord", 0);
 }
 
+static int
+api_dvr_entry_skipped_toggle
+  ( access_t *perm, void *opaque, const char *op, htsmsg_t *args, htsmsg_t **resp )
+{
+  dvr_entry_set_skipped();
+  return 0;
+}
+
 static void
 api_dvr_rerecord_deny(access_t *perm, idnode_t *self)
 {
@@ -332,6 +340,21 @@ api_dvr_entry_remove
   ( access_t *perm, void *opaque, const char *op, htsmsg_t *args, htsmsg_t **resp )
 {
   return api_idnode_handler(&dvr_entry_class, perm, args, resp, api_dvr_remove, "remove", 0);
+}
+
+static void
+api_dvr_previouslyrecorded(access_t *perm, idnode_t *self)
+{
+  dvr_entry_t *de = (dvr_entry_t *)self;
+  if (de->de_sched_state == DVR_SCHEDULED)
+    dvr_entry_previously_recorded(de);
+}
+
+static int
+api_dvr_entry_previouslyrecorded
+  ( access_t *perm, void *opaque, const char *op, htsmsg_t *args, htsmsg_t **resp )
+{
+  return api_idnode_handler(&dvr_entry_class, perm, args, resp, api_dvr_previouslyrecorded, "remove", 0);
 }
 
 static void
@@ -531,6 +554,8 @@ void api_dvr_init ( void )
     { "dvr/entry/rerecord/allow",  ACCESS_RECORDER, api_dvr_entry_rerecord_allow, NULL },
     { "dvr/entry/stop",            ACCESS_RECORDER, api_dvr_entry_stop, NULL },   /* Stop active recording gracefully */
     { "dvr/entry/cancel",          ACCESS_RECORDER, api_dvr_entry_cancel, NULL }, /* Cancel scheduled or active recording */
+    { "dvr/entry/previouslyrecorded", ACCESS_RECORDER, api_dvr_entry_previouslyrecorded, NULL }, /* Add selected schedules as previously recorded */
+    { "dvr/entry/showskipped/toggle", ACCESS_RECORDER, api_dvr_entry_skipped_toggle, NULL },
     { "dvr/entry/remove",          ACCESS_RECORDER, api_dvr_entry_remove, NULL }, /* Remove recorded files from storage */
     { "dvr/entry/filemoved",       ACCESS_ADMIN,    api_dvr_entry_file_moved, NULL },
     { "dvr/entry/move/finished",   ACCESS_RECORDER, api_dvr_entry_move_finished, NULL },
--- a/src/dvr/dvr_autorec.c
+++ b/src/dvr/dvr_autorec.c
@@ -384,7 +384,7 @@ dvr_autorec_add_series_link(const char *dvr_config_name,
   dvr_autorec_entry_t *dae;
   htsmsg_t *conf;
   const char *chname;
-  char *title;
+  char *title, *name;
   if (!event || !event->episode)
     return NULL;
   chname = channel_get_name(event->channel, NULL);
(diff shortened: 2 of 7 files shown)


src/api/api_dvr.c
```diff
@@ -531,6 +554,8 @@ void api_dvr_init ( void )
     { "dvr/entry/rerecord/allow",  ACCESS_RECORDER, api_dvr_entry_rerecord_allow, NULL },
     { "dvr/entry/stop",            ACCESS_RECORDER, api_dvr_entry_stop, NULL },   /* Stop active recording gracefully */
     { "dvr/entry/cancel",          ACCESS_RECORDER, api_dvr_entry_cancel, NULL }, /* Cancel scheduled or active recording */
+    { "dvr/entry/previouslyrecorded", ACCESS_RECORDER, api_dvr_entry_previouslyrecorded, NULL }, /* Add selected schedules as previously recorded */
```
azlm8t: I think this is similar to #1020 (where you press "stop" on upcoming recordings to stop them from recording)?

src/tvheadend.h
```diff
@@ -561,6 +561,8 @@ typedef enum {
 #define SM_CODE_NO_INPUT                  402
 #define SM_CODE_NO_SPACE                  403
 
+#define SM_CODE_PREVIOUSLY_RECORDED       500
```
azlm8t: I'm not sure this code is being used since it seems you use SM_CODE_OK?

But I _think_ using this might require changes to htsp (for Kodi frontend) to map it for older clients (who don't know what a code 500 is). It _might_ be worth using a code in the 300 range such as 301 (since 300 is ABORTED and this is a type of abort).

src/dvr/dvr_db.c
```diff
@@ -135,6 +139,24 @@ int dvr_entry_is_upcoming(dvr_entry_t *entry)
...
+	 	
+	 }else{
+	 	
+	 	if (_dvr_duplicate_event(entry) == 0){
+	 		return state == DVR_RECORDING || state == DVR_SCHEDULED || state == DVR_NOSTATE;
+	 	}else return 0;
+	 	
+	 }
+}
+
```
azlm8t: Would it be possible to re-indent/fix whitespace to match existing code?

src/dvr/dvr_db.c
```diff
@@ -4185,6 +4230,35 @@ dvr_entry_cancel(dvr_entry_t *de, int rerecord)
   return de;
 }
 
+static void
+dvr_entry_add_previously_recorded(dvr_entry_t *de)
+{
+  de->de_dont_reschedule = 1; // Set as not reschedule
+  de->de_file_removed = 1; // Set as file removed
+  de->de_start = gclk(); // Need in case you want to record it again after.
+  de->de_stop = gclk() + 1000;
```
azlm8t: I don't quite understand the changing of de_start/de_stop.

src/api/api_dvr.c
pablozg: I'm sorry I haven't seen that pull request before, I'm come from argustv server with a pc, and in argus I had the option to select one program and mark it as previously recorded and then the autorec rule will skip it in the time, so I tried to replicate it in tvheadend, in this case the button only is enabled if you select a upcoming recording and disable if the program is recording now.

When you press it popup a question to be sure to add it as previously recorded and then move it to "removed recordings". So the autorecord rule still active and will skip same episode or program in the time.

src/dvr/dvr_db.c
```diff
@@ -57,6 +57,10 @@ static void dvr_timer_stop_recording(void *aux);
 static int dvr_entry_rerecord(dvr_entry_t *de);
 static time_t dvr_entry_get_segment_stop_extra(dvr_entry_t *de);
 
+static dvr_entry_t *_dvr_duplicate_event(dvr_entry_t *de);
+
+int showskipped = 0;
```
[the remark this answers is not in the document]
pablozg: I had not thought about it, because I'm the only use tvheadend at home, you're right, but my skills is not enought for do it in other way. If you could give me some ideas I would try to solve. I would like to progress in programming, but I'm self learning programmer. Thanks.

src/dvr/dvr_db.c
pablozg: I will fix it.

src/dvr/dvr_db.c
pablozg: I do it because I don't like to have removed recordings with a date later than the day I have omitted them, is only a cosmetic fix. :-)

src/tvheadend.h
pablozg: This is a mistake, I had tried to have a self code, but I would have to do more changes, so I prefered to send as SM_CODE_OK, but I forgot to remove it. I'm sorry.

src/dvr/dvr_autorec.c
```diff
@@ -384,17 +384,20 @@ dvr_autorec_add_series_link(const char *dvr_config_name,
...
-  char *title;
+  char *title, *name;
   if (!event || !event->episode)
     return NULL;
   chname = channel_get_name(event->channel, NULL);
   if (!chname)
     return NULL;
   conf = htsmsg_create_map();
   title = regexp_escape(epg_broadcast_get_title(event, NULL));
+  name = (char *)epg_broadcast_get_title(event, NULL);
```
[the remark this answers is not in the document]
pablozg: Ok, I have change it. Thanks.

src/dvr/dvr_db.c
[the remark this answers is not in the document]
azlm8t: It seems to me: api_idnode_grid_conf creates a config with a field filter from the http request from api_idnode_grid. So I imagine api_dvr_entry_grid_upcoming (which takes a conf parameter) would apply such a filter. The filters are specified in the idnode.h.

There's some logic in idnode.js for building the filters[].

So perhaps a filter can (somehow) be applied directly to the grid_upcoming in the Javascript code and it would remove the need for any C code at all since you could possibly apply a "duplicate eq 0" filter and it would remove the duplicates.

Then, in the future, we can add a "title regex" search which is what I'd quite like for dvr panels.

But that's based on a quick look through idnode and I imagine the difficult part is forcing the filter in the JS. There is already filter logic in the epg.js that may help, but that uses a slightly different way of presenting information so it may not prove too helpful.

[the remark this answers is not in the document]
pablozg: Hi, now I able to filter the store data pressing the button on the client side, but I have a little problem, I can't update the "pagingtoolbar" after apply the filter, so it still showing the total count not the filtered count. Do you know how can I do it? Thanks. 

This is the button code:

var showSkippedButton = {
        name: 'showskipped',
        builder: function() {
            return new Ext.Toolbar.Button({
                tooltip: _('Show / hidde duplicate schedules'),
                iconCls: 'showskipped',
                text: _('Skipped'),
                disabled: false
            });
        },
        callback: function(conf, e, store, select) {
                  if (store.isFiltered()){
        	     store.clearFilter();
        	  }else{
        	     store.filter('duplicate', 0);
        	 }
        }
    };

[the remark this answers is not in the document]
azlm8t: Fantastic.

I don't know how to do update the total count. I just checked the EPG and that does update the counts. Its code in epg.js has "totalProperty: 'totalCount'" and elsewhere it has a "epgStore.getTotalCount" when creating an autorec rule which definitely returns the filtered count. So, I don't know if that's useful?

_Maybe_ setting "grid.getStore().totalProperty='totalCount'"?  It looks like it defaults to "total" in idnode.js, so maybe setting it after the create in dvr.js:351?

Let me know if that works, otherwise I probably won't have time to look more before the new year, but I'll take a look then. (And if I don't, please add a new comment to remind me).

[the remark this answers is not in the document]
pablozg: After some test, the filter only works in the actual page, so now the client request to the server at two differents url's, one filtered and the other one as it.

PR #1048, opened 2017-12-09
dvr: Add autorec for new-only. (#1167).

azlm8t:
Previously we had "all", "new/unknown", and "repeat", but
no ability to only record episodes marked as "new". So we
rename DVR_AUTOREC_BTYPE_NEW to DVR_AUTOREC_BTYPE_NEW_OR_UNKNOWN
to remain backward compatibility with existing autorec
rules and add new semantics for DVR_AUTOREC_BTYPE_NEW.

This helps with OTA "scrape new" where broadcasters mark shows as new,
but all other shows aren't marked as repeat so remain as "unknown".

We don't update htsp since it currently does not send the
broadcast type field.

Also alter DVR_AUTOREC_BTYPE_NEW_OR_UNKNOWN since
previously we never checked 'new' but instead checked 'repeat'.
However, SD has a previously-shown for all programmes (even first
showings) which causes us to mark programmes as repeat.

It is difficult to fix the repeat logic without breaking existing
behaviour since in the US a programme can be a premiere but have
a previously-shown of the previous day due to timezone differences
on the coasts. Similarly, programmes can be premiere outside the US
but have a previously shown date from the US or from a different channel.
For that reason we now check 'new' instead of 'repeat'.

Real example: Programme is shown on channel A at 9pm and on A+1 timeshift
channel at 10pm. Both are marked as "new" in the paper/OTA tv guide. However,
the programme was actually first shown three years ago on a premium
channel, so it's actually also a repeat since it has been shown before.
So the programme is both a new episode and a repeat episode.

Similarly, one of my tv channel insists Roger Moore Bond films from the
1970s are "new" even though most people would consider them a repeat,
but since it's the first time that particular channel has aired it
they use the "new" tag.

Issue: #1167.

### DIFF
--- a/src/api/api_epg.c
+++ b/src/api/api_epg.c
@@ -343,6 +343,7 @@ api_epg_grid
   if (str)
     eq.stitle = strdup(str);
   eq.fulltext = htsmsg_get_bool_or_default(args, "fulltext", 0);
+  eq.new_only = htsmsg_get_bool_or_default(args, "new", 0);
   str = htsmsg_get_str(args, "channel");
   if (str)
     eq.channel = strdup(str);
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -325,8 +325,9 @@ typedef enum {
 
 typedef enum {
   DVR_AUTOREC_BTYPE_ALL = 0,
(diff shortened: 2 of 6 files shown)


perexg:
Merged. Thanks.

PR #1050, opened 2017-12-09
ui: Allow filtering by categories (#4777)

azlm8t:
If the server has parsed xmltv categories then we create a second toolbar below the existing one to add category filters and pass them to the  rule.

We use a second toolbar since the primary toolbar is too cramped to add additional filters.

Also fix a small memory leak on the api filter and fix bug where an autorec with a category would match OTA programmes without categories.


azlm8t:
Moved the free in to epg_query to be consistent and force-pushed to remove the incorrect free() patch.

perexg:
Merged. Thanks.

PR #1051, opened 2017-12-09
api: 4.2: Fix small memory leak on filter. (#4773).

azlm8t:
This is backport from same fix in #1050 since I believe 4.2 would have the same problems based on a look of its code. (Not tested on 4.2, but tested on 4.3).

### DIFF
--- a/src/api/api_epg.c
+++ b/src/api/api_epg.c
@@ -475,6 +475,10 @@ api_epg_grid
 
   epg_query_free(&eq);
   free(lang);
+  if (eq.lang)        free(eq.lang);
+  if (eq.stitle)      free(eq.stitle);
+  if (eq.channel)     free(eq.channel);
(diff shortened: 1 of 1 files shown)


src/api/api_epg.c
```diff
@@ -475,6 +475,10 @@ api_epg_grid
 
   epg_query_free(&eq);
   free(lang);
+  if (eq.lang)        free(eq.lang);
```
perexg: This probably belogs to epg_query_free(), doesn't ?

Also, it seems that epg_query() does such free calls.

azlm8t: Yes, you're completely right.
I think it's because I saw the alloc and no free when I was adding my filter on the other PR so added them in.

Let's scrap this PR and I'll fix up my #1050 to move my free in to epg_query.

PR #1052, opened 2017-12-09
codacy: Fix couple of problems identified by codacy. (#4773).

azlm8t:
Fixed two issues identified by codacy in issue 4773.
- buffer overrun in memcpy (I believe 3 is correct value since buf is already memset 0);
- redundant condition in xmltv.c.

Codacy also spotted a local variable returned at src/htsp_server.c:3385. I thought of making it static, but then it doesn't look like it's actually returned since it's nulled at 3471. But since it's added to a LIST_INSERT_HEAD at 3400 (but then removed soon later) and used in another thread, I _think_ this variable should be malloc'd but haven't worked it out to be sure, so I've not made any changes to it.


perexg:
Merged. Thanks.

PR #1053, opened 2017-12-12
eit: Allow empty match subexpressions (#4787)

banburybill:
If a scrape regex includes a subexpression matching the null string (),
this match is treated as if the regex did not match.
Amend this to return an empty string as the match; this is plainly what
the regex author wanted.

As an example of why this might be wanted, consider the UK Freeview
extraction of a subtitle from the summary. A user might wish to specify
the subtitle is left blank if no obvious  subtitle is present in the
summary.

Issue: #4787.


perexg:
Merged. Thanks.

PR #1054, opened 2017-12-13
eit: Add optional 2nd match subexpression for subtitle (#4791)

banburybill:
If the regex for the subtitle contains a second subexpression, and a match is made, use the first subexpression for the subtitle and replace the summary with the second subexpression.
    
For example, a UK Freeview subtitle regex might choose, when matching a summary 'Subtitle: Text', to set the subtitle to 'Subtitle' and set the summary to 'Text' to avoid repetition of the subtitle.


perexg:
Merged. Thanks.

PR #1055, opened 2017-12-14
eit: extend generic regex handling for subpatterns and use in scraper regex (#4795)

banburybill:
Currently scraper regex usage is hardwired to Posix. Using PCRE/PCRE2 if available would give more flexibility and potentially save repetition in patterns, e.g.

    (?:[.][.][.][:.]*[.:]|[0-9]+/[0-9]+[.])? ([^:]*):

would require multiple Posix patterns, each duplicating the captured subpattern.

So, this PR extends the interface in tvhregex.h to handle subpatterns and uses it in EIT scraper regex matching.


CvH:
Could you add some docs about that usage ? If you don't follow github you have probably no idea that this works at all so this might end as feature that nobody use because nobody knows about it ;)

banburybill:
Yes, good point. As far as I can see, how scraper regexes are used isn't really documented anywhere at present. I guess that data/conf/epggrab/eit/scrape/README should contain something more, apart from how to test the regexes, and perhaps there should be a wiki page too.

(Although I've been using TVH for a long time, I'm new to TVH hacking, so I may have missed some docs.)

azlm8t:
Most of the documentation is minimal but currently resides in the READMEs and in the PR/bug report such as [4577](https://tvheadend.org/issues/4577#note-1) and [4287](https://tvheadend.org/issues/4287). The files are "system" files rather than "user" configuration files, so mostly edited by developers, though we have had a few contributions.

My one concern with using pcre is that currently it is not a requirement of Tvheadend. It is currently optional. So, a user can use pcre in EPG autorecs since they write them on their system (and not shipped with Tvheadend), but configuration files, such as scraper files, ship without currently requiring pcre. Also I believe Python's re (used in the test harness) isn't fully pcre compatible, though that's probably a minor problem since I doubt the scraper regex used will exercise the obscure cases.

So shipping scraper configuration that only works if you have pcre would be change in dependencies and would mean those scraper configurations will not work on non-pcre systems.

As developers, all of our systems have pcre, but I have seen bug reports of people trying to run it on obscure limited platforms such as their router. Whether we should restrict improvements due to potential edge case hardware is someone else's judgement call.

Unless anyone knows for sure that there are platforms that _don't_ support pcre or that all platforms _do_ support pcre, perhaps we should just make pcre a configuration requirement for a few weeks before merging and see if anyone raises a bug saying we broke their build. 

However, I note that the existing opentv scraper does things differently and has "cleanup_title" regex to extract the required title and doesn't require pcre.

I think the overall intention of this PR is to try and tidy-up UK Freeview/Freesat titles/summary. Having looked at Mythtv's tidy functions in the past, I believe that it's something that can't be done by regex operating on separate fields alone due to, for example, sub-titles being split between title and summary, with lots of edge cases (such as "Agents of S.H.I.E.L.D." causing problems) However I see nothing wrong with getting common case tidy-ups working to make things better.

banburybill:
You're quite right - I'm attempting to scratch an itch with UK Freeview titles/summary. I also agree that regex on single fields, or indeed any regex processing, is always going to have failing edge cases.

And yes, the Python re used in the test harness is not PCRE, and has differences to PCRE, though it's a lot closer to PCRE than Posix.

I'm not intending (myself) to alter the supplied uk regexes; in no small measure, these are a matter of personal taste. However, since I'm using the Debian packages, and those are built with PCRE, and PCRE is used in the autorec regex, and PCRE would make what I'm trying to do a bit easier, I'd like to use it. However, I do also get the longer-term danger of someone accidentally updating those patterns to something that doesn't work with Posix. 

And yes, I have been thinking of adding title cleanup too, but I've not figured out yet what/how. :-)

perexg:
Ok, we have just two threads (one in bug tracker - https://tvheadend.org/issues/4795 - and one here). To follow my objection: Perhaps, we can specify the syntax in the top-level of json config like:

<pre>
{
  "posix": {
     "season_num": [],
     "airdate": []
  },
  "pcre": {
     "season_num": [],
     "airdate: []
  }
}
</pre>

If "pcre" rules does not exist the "posix" should be fallback. I don't know the exact differences between the former pcre and newer pcre2 API, but it seems that mostly it's just about C API cleanups.

For python, I believe there's also pcre bindings to make serious tests.

azlm8t:
I don't use OTA EIT so the changes won't affect me.

But, I still don't really understand the benefit other than being initially a bit easier to write the regex. For the example regex in the bug report, isn't there already regex extractors in the UK scraper and test harness which extract those subtitles. Granted, it requires two separate regex and not one extended regex, but it currently works as-is and is portable across all systems.

The (UK) eit copies the exact same text in to sub-title and summary, and it appears to have existed since the sub-title field was created.

I'd have preferred to edit eit.c to just copy the text to the summary and then extract a sub-title if possible.

This would then be usable by everyone, not just people who can edit regex in system config files, or create personal files that they forget about and become out-of-date.

Then, add an array of regex to tidy the summary where each entry is run in turn with the results passed to the next entry.

So, you'd start with "subtitle: Summary here [HD,S]" and the first regex would strip the colon to leave "Summary here [HD,S]", and the second would strip the suffix to leave "Summary here". If a regex didn't match then we'd supply the previous matched string for the next regex match.

Most people can't understand regex and I really believe we should ship something where it just works, or requires a tickbox if there are alternative approaches or it affects pre-existing recording rules. I'm not convinced that tidying up EIT is personal taste: I think most people would like the HD extracted and scraped in to the is_hd flag, and the S in to is_subtitled.

But my only current interest in OTA EIT is Bulgarian genres so I don't mind either way.

banburybill:
I've been doing some more thinking on this.

I must admit that I don't quite get the concern over regex engine; at least, why this exists when the existing code already (for some uses) varies the library in use. However...

You're quite right that editing regex is a minority sport, and something sensible should happen out of the box. I also seen where you are going with the idea of successive regex performing edits, and I like the ability that gives. And when I talked about personal taste, I was simply thinking of why I started; text repeated in the subtitle and summary bothers me, but is not an issue for others. Of course there are items such as HD and Subtitle flags that should always be extracted. Right now I'm actually reasonably happy with the rules I'm using on my production system, which are all Posix-compliant:

  "scrape_subtitle": [
      "^([.]{2,3}[^:.]*)[.:] +(.*)",
      "^[0-9]+/[0-9]+[.]? +([^.?!:]+[!?]?): +(.*)",
      "^[0-9]+/[0-9]+[.]? +([A-Z][^.?!]{4,70}[.?!]) +(.*)",
      "^[0-9]+/[0-9]+[.]? +()",
      "^New[.:] +([^.?!:]+[!?]?): +(.*)",
      "^New[.:] +([A-Z][^.?!]{4,70}[.?!]) +(.*)",
      "^New[.:] +()",
      "^([^.?!:]+[!?]?): +(.*)",
      "^([A-Z][^.?!]{4,70}[.?!]) +(.*)",
      "()"
  ],


It's true that the existing uses I've given, and indeed the uses I had in mind when doing the work, was just to make writing the regexes easier, and in particular avoiding the combinatorial explosion in regexs required on every alternate selection because Posix doesn't have non-capturing groups, hence all the repetition in the above.

But there are cases where things I'd like to do which simply cannot, I think, be done with Posix regex. For example, the UK pattern currently extracts as a subtitle anything ending in a : using the pattern "([^:]*):". But this goes wrong if the summary happens to contain a : in a sentence past anything that might be a subtitle, so I'm restricting subtitle extraction to text where : occurs before the end of a sentence using "([^.?!:]+[!?]?):". This fails on the corner cases where . is use in a context where it doesn't end a sentence, so (as mentioned) 'Agents of S.H.I.E.L.D.' or 'The work of Dr. Phil Academic'. I could guard against these using a regex lookahead or lookbehind, e.g. "(?<!([.].|Dr)[.]",  but these aren't available in Posix.

I have also been thinking about how to edit the title to include text from subtitle or summary. It's common in UK EIT to have a title which I guess is too long for older devices and continue it in the summary, e.g. 

Scottish Parliament - Portfolio...
...Questions.

So, I have a suggestion. How about a scheme like this? First, we keep the existing scraper regexes, but ensure that they are **always** executed using a Posix engine.

Second, we have a set of new regex ("capture_title/subtitle/summary") which are only executed if a PCRE engine is available, but if it's available **and** capture patterns are specified they are executed instead of the Posix ones. capture_title is given EIT title/subtitle/summary all concatenated with a space separating them, capture_subtitle gets EIT subtitle and summary. The result is **all** group captures concatenated (assuming the pattern matches). This lets me pick out bits from the inputs; to take your example, I can omit [HD,S] with a non-collecting group.

I can also add Posix scrapers for subtitle, AD, HD, widescreen which don't currently exist, but that's for a different PR.

It would then be possible to ship rules which are more or less capable and will work automatically depending on the regex libraries available.

Would you like that any better that the current proposal?

perexg:
My opinion is that if we can specify the regex type in the json config file to make sure that the regex works as intended, any solution is acceptable to me.

On a side note: The regexes might slow the input processing, so I'm working to move parsing of OTA EPG data (EIT, opentv) to one thread which can drop the received data if the queue is too big.

banburybill:
Possibly the way to proceed right now would be to redo this PR to use regex engine specific regexes if specified. So the JSON would specify:

```
{
   "scrape_subtitle": [...],
   "pcre" {
      "scrape_subtitle": [...]
    }
}
```

The base regexes will always be executed by the Posix engine. Iff PCRE is available, AND a PCRE regex is specified, that regex is used instead.

I can then play with other ideas above separately.

perexg:
Looks good to me. But I found another major issue with the config files when I'm trying to move things outside global lock - they don't take care about the language. Actually, I believe that the general rules should handle multiple languages. The current situation where the user can choose the file with regexes is mostly fine, but it does not work correctly in the multi-language environment. I know guys, you probably don't care for your usage, but the project is aimed to be a general solution.

So, ideally, we should make rules as:

<pre>
{
  "eng": {
     {
        "scrape_subtitle": [...],
        "pcre" {
           "scrape_subtitle": [...]
        }
     }
  },
  "cze": {
    .....
  }
}
</pre>

The language names are in src/lang_codes.c .

The eit_pattern_list functions should be extended to pass the language string.

On the other side, we may probably specify a list of languages which will be "scraped" and/or filtered in the grabber configuration to save the computing time.

banburybill:
That seems a reasonable point, but perhaps a bit much to shovel into one PR. If I carry on, and add a regex selection hierarchy based on regex engine, I think that adding a further selection layer based on language would make a good followup PR.

perexg:
I agree that the language should be handled in a separate update.

My EIT rework (untested!): https://github.com/perexg/tvheadend/commits/epg . It should not clash with your changes much, but actually, my code applies scraping to all languages, because we don't have proper config data yet.

banburybill:
Thanks for the warning. Assuming yours are near ready to push, I'll carry on but not update this PR until your have appeared and I've merged them.

banburybill:
This PR now selects regexes from a "pcre" or "pcre2" hierarchy as appropriate or available as outlined above - changes pushed so I don't lose them. I'll wait for the other EIT changes and merge them.

perexg:
I pushed my changed to master.

banburybill:
OK, merged and pushed. So, to repeat, EIT scraper regexes are by default POSIX and executed by the POSIX engine. Optionally you can specify "pcre" or "pcre2" specific regexes. If one of PCRE or PCRE2 is built-in, its regex will be used (if specified) and executed by that engine. POSIX regexes will continue to be executed by the POSIX engine.
```
{
   "scrape_subtitle": [...],
   "pcre" {
      "scrape_subtitle": [...]
    }
}
```
I think this means everyone should be happy now?

perexg:
Please, rebase to the current master or make your change as one commit (diff against the current master). The merging is evil in my eyes.

azlm8t:
Looks good.

For language support, when we add it, I don't know if we also need "*" in the config file for fallback. For example, season/episode and certainly broadcast year might be in a common format on a satellite, whereas subtitle scraping is language-specific.

Part of me thinks we should move the posix regex in to a "posix" subsection so it's all tidy and consistent, but I think that might just cause unnecessary re-work of our existing configs, and I don't think it would add much benefit.

<pre>
{
   "posix" {
       "scrape_subtitle": [...],
    },
   "pcre" {
      "scrape_subtitle": [...]
    }
}
</pre>

banburybill:
Rebasing is evil in my eyes :-) But, OK, changes rebased and a couple merged :-)

The Posix regexes are different from the others, in that they are guaranteed to be always available, so while I understand the urge to file them under 'posix', part of me also likes having them not require a special qualifier because they are the defaults.

perexg:
Merged with one extension: b9ce472cfae6449357b18ca1bcb35fb9529aec7c . Thank you for your contribution.

PR #1057, opened 2017-12-15
xmltv: optional disable mapping xmltv category to genre (#3753)

azlm8t:
Some frontends (such as Kodi) can only use categories if we have no genre. Often genre mapping from xmltv is imprecise since xmltv programmes often have a number of categories and one happens to match a genre. For example opera was "non-music" due to matching a minor-category tag of "performing arts", but not matching the major category tags such as "opera".

Tested by edit4ever ! in comment 28 of bug report.

Also a minor change to cast to epggrab_module_int_t instead of epggrab_module_ext_t for existing scrape config options since these are in the base struct.


meyerchak:
thanks edit4ever

perexg:
Merged. Thanks.

PR #1059, opened 2017-12-18
Map genre from xmltv category (#4753)

azlm8t:
Previously we required a one-to-one mapping between xmltv categories and genres. So the xmltv category of "Documentary" is mapped to epg genre of "Documentary".

However, many mappings are not one-to-one. For example, "Opera" should be mapped to "Musical / Opera". So we now perform common mappings for SchedulesDirect and for the comments in the Kazer thread for French and Spanish, and from sample guide.xml for Polish webgrab in bug 4424.

Currently people are running perl scripts to map category names in to English genre names prior to import, and so have categories in a mixture of languages. This change will remove the need for external scripts.

Since we have only had a couple of category mapping contributions submitted in years, and since categories are a mixture of English and native languages, I've taken the easier/more performance route of just having one configuration file with all the mappings and auto-generating a C lookup array at compile time instead of having multiple configuration files per language. Autogenerating should allow us in the future to also auto-generate Javascript for displaying icons for these categories using their original language (instead of falling back to using genre icons).


perexg:
I think that my idea was to exactly identify the genre in the commit a7ee9411f86238b3905ca4949af812bd409071ca . It's fine for UI, but it does not work for the xmltv import. My fault. What about to add a new function epg_genre_list_add_by_str2() or just extend the '_epg_genre_find_by_name()' to search also in the separated strings when the 'joined/full genre string' is not found ?

The change is fine, but I just want to reduce the obvious English cases..

Also - we should also handle the 'lang' attribute in the category element ?

azlm8t:
Sounds good.

I've taken a look and perhaps in epg.c in _epg_genre_find_by_name the _genre_str_match fails then we could additionally compare the category string against just the tvh_gettext_lang(_epg_genre_names[i][j]) (instead of joining sub-genres with a / as in _genre_get_name). This will match common strings like "football" which are already translated in to multiple languages so will just work for them.

So let's:
- move _xmltv_map_category_to_epg_genre in to epg.c and rename.

In _epg_genre_find_by_name logic:
- _genre_str_match on genre_get_name; else
- _genre_str_match on tvh_gettext_lang(_epg_genre_names[a][b]) (i.e., each sub-genre string on its own).

If all the above fails then:
- fallback to doing _xmltv_map_category_to_epg_genre to map remaining categories like "Historical drama" to "Historical movie"

Re: language.

The one problem is that since those strings _are_ internationalized, it means that an xmltv category that uses an English name of "Movie / Drama" for compatibility would not match if the system was run in a different locale.

So, if we pass lang from xmltv then perhaps we should do two lookups, one with lang (for cases where xmltv has local language) and one without lang (where xmltv uses English names for categories).

I'll take a look after the holidays.

michaelarnauts:
This PR seems to have gone stale?

Leatherface75:
Same here i wonder what happened?

azlm8t:
Sorry, I had some hardware failures and then worked on getting the image support in ready for the Kodi release. I'll have to refresh my memory.

piotrasd:
+1 for add that ;)

azlm8t:
I've added a few more patches if you want to try it (since I don't have appropriate xmltv).

We continue to map known categories to genres in the generated csv as before using mappings I found in the forums.

Otherwise, we lookup each sub-category word, so "Opera" in the xmltv now matches against "Musical / Opera". We do this using both English and localized version of the words since many xmltv that I've seen use English for most categories or a mixture of English and non-English categories.

The htsp client tries to find the "best" genre for sending to clients such as Kodi (as discussed in the forum). This is because xmltv may have mutliple categories/genres such as "adventure, action, childrens" and we want to send the major genre of "children" to get colour in Kodi rather than a sub-genre of "adventure" which would get the generic "movie/drama" colour. The logic for the "best genre" seems reasonable for my setup and makes the guide colourful, and has a fallback of "first genre" if the logic fails, so for OTA (with one genre) it should work as before. It's difficult to be 100% for "best genre" since xmltv can have numerous categories, but it should work for most programmes.

I won't be around much over the holidays to do any fixups, but if it works/doesn't work then post a comment so it's known if it can be merged.

piotrasd:
here full polish xmltv with all categories if you want look or test
https://www.sendspace.com/file/sbkg1b

azlm8t:
@piotrasd 
The xmltvgenre.csv first column is from "_epg_genre_names" in src/epg.c. So "Movie / Drama" is generated from the line that says 'N_("Movie"), N_("Drama")'. That's what the kodi forum has, but that does not have the " / " that we need.

The remaining columns in the file are the non-English words that map to that genre.

So "Education / Science / Factual topics, Edukacyjny" means the Edukacyjny maps to Education / Science / Factual topics genre.

Hope that helps.

perexg:
Thinking more about this and it would be probably much better to have configuration stored in json in the configuration directory and not hardcoded in the binary. In this way, users may just modify the config file. The sorting might be done in the load time.

Kick4U2:
Any chance we can add the "user defined" (0xF) major genre to the code since Kodi has implemented this in 'xbmc/strings.po' (lines 11645-11700)?  I hate Comedy shows showing up as a movie.  SD provides a category for seperating movies from series to determine which field it should be.
I want to help, but am just learning how to use git so excuse my ignorance. Here's a link to a branch I made to assist.
[tvheadend EPG.c](https://github.com/Kick4U2/tvheadend/pull/1/files)

Leatherface75:
This is what this file does

"https://github.com/tvheadend/tvheadend/pull/1059/files#diff-de915aab61211d1842b9b6b1c35bcd71"

Kick4U2:
Hello,

I agree and like how this is implemented, but it puts all entries for Comedy under the major category of Movie/Drama (ETSI EN 0x1).  As you can see from the comments in the link you provided, a daytime quiz show is an example from SchedulesDirect where Comedy isn't always designated as a Movie.  The ETSI standard has Comedy solely for Movies.

Kodi has implemented the 'user defined' (ETSI EN 0xF) major category to properly represent genres that could be used for series and not only movies.

Since SchedulesDirect always has the first category either as "movie", "sports", "series", or "tvhsow" so I figured these should be used to make a decision on whether the where the following categories fall under.

Hope that makes sense...
________________________________

Leatherface75:
My point is that you can edit that file that fixes the issues you talk about.

Kick4U2:
Thanks again, but how do I change the major category?  It seems to me that if you have a category of Comedy, than there is no way in telling tvheadend that it is a show.  It automatically uses Movie/Drama as the major category (content-type).

What I’m saying is that there should be some additional logic looking for category “movie” or “series” in order to distinguish the difference for what category Comedy is meant to be under.


I don’t see how editing this file can change the major category for Comedy from 0x1 to 0xF.  Would you please elaborate?  Thank you


Friday, June 5, 2020, 10:14 PM -0400 from notifications@github.com  <notifications@github.com>:

Kick4U2:
Can anyone explain what I'm doing wrong?  It seems no one else is having this same issue... perhaps I'm making too big a deal about this???

To help explain, I've included an example of a kids TV Show that has a category of Comedy and is coming up with the Content Type of "[Movies / Drama] Comedy".  Ideally, TVHeadend would search for a major category within the list of categories (in the case below it would find "Children's / Youth programmes") and consider this as the Content Type.  If no major category is found than it would ideally fall back on the first category (SchedulesDirect provides the first category either as "movie", "sports", "series", or "tvshow").


image

Here's the listing within the XML file:

>   <programme start="20200808223000 +0000" stop="20200808230000 +0000" channel="Boomerang">

    <title lang="en">The Tom and Jerry Show</title>
    <sub-title lang="en">Tap Cat</sub-title>
    <desc lang="en">(2014-03-01). (TV-PG). An old pair of Ginger's tap shoes turns Tom into a dancing fanatic. Cast: Stephen Stanton, Rick Zieff, Grey Griffin, Gary Cole. Categories: Animated, Children, Comedy, Episode, Series, Show(n)</desc>
    <credits>
      <director>Darrell Van Citters</director>
      <actor>Stephen Stanton</actor>
      <actor>Rick Zieff</actor>
      <actor>Grey Griffin</actor>
      <producer>Sam Register</producer>
      <producer>Jay Bastian</producer>
      <presenter>Gary Cole</presenter>
    </credits>
    <date>2014</date>
    <category lang="en">series</category>
    <category lang="en">Animated</category>
    <category lang="en">Children's / Youth programmes</category>
    <category lang="en">Comedy</category>
    <category lang="en">Episode</category>
    <category lang="en">Series</category>
    <category lang="en">Show</category>
    <icon src="https://schedulesdirect-api20141201.s3.us-east-1.amazonaws.com/assets/p10478306_b_v8_ao.jpg" />
  </programme>

PR #1060, opened 2017-12-18
xmltv: Only log "new episode" if episode has been created. (#4803).

azlm8t:
The xmltv import always logs new and tot to be equal. Now we check the CREATE flag to determine if the entry was newly created to update either new or mod appropriately.


perexg:
Merged. Thanks.

PR #1061, opened 2017-12-18
ui: Rename "Category Type" icons column to "Category Icons" (#4594).

azlm8t:
In the forum, it was mentioned that we have two EPG columns called "Content Type", so rename the one with icons to be called "Category Icons" to avoid any confusion.


perexg:
Merged. Thanks.

PR #1066, opened 2017-12-21
eit: Fix undefined charset (fixes #4807)

ljalves:
Fixes #4807


perexg:
Fixed in 72fd5e25f881850090ab8e59a9548c53e153f6a7 (a bit differently). Your proposed patch used service_t outside the global_lock which is wrong.

PR #1067, opened 2017-12-21
eit: Add title and summary scrapers, and re-work how text is extracted

banburybill:
This PR adds title and summary scrapers to EIT grabbing. As with subtitles, the summary scraper is given just the summary field as input, but the title scraper is given "<title> <summary>" as input (note space between title and summary).

The output from a scraper is extended to become the concatenation of **all** matched sub-patterns. This allows patterns to cherry-pick text from different parts of their input. This breaks my previous (unused in all public patterns) hack of setting the summary from a second sub-pattern in the subtitle.

The new scrapers mean it is now possible to reconstruct long titles when split by the UK EPG e.g. a title "Lorum Ipsum..." and summary "...dolor sit amet. Consectetur adipiscing elit: Vestibulum aquilam sodales tellus." can now be scraped into title "Lorum Ipsum dolor sit amet.", subtitle "Consectetur adipiscing elit", summary "Vestibulum aquilam sodales tellus."

I have an updated set of patterns for UK Freeview/Freesat in development using these, but that's for another PR. Mechanism first. This PR does not alter the behaviour of the existing UK scrapers.

At present, title and summary scraping can be enable/disabled separately to subtitle scraping. I am undecided whether they should be combined into a single user setting. It's likely/possible that title/subtitle/summary patterns will work in concert, and having them separately enabled may give confusing results.

### DIFF
--- a/data/conf/epggrab/eit/scrape/README
+++ b/data/conf/epggrab/eit/scrape/README
@@ -11,32 +11,40 @@ object are:
 * episode_num
 * airdate
 * is_new
+* scrape_title
 * scrape_subtitle
+* scrape_summary
 
 Each member's value is a list of regular expressions. Each regular
 expression must contain at least one sub-pattern, i.e. a pattern
 enclosed in (). Input data is matched against the first regex in the
 list. If no match is found, the second regex is tried, and so on until
-a match is found or the list exhausted.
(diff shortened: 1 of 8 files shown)


perexg:
Merged. Thanks.

PR #1069, opened 2017-12-25
Little user friendly changes on webui

pablozg:
I have made some user friendly changes on the webui:

- When show / hide the duplicates recordings, now it show / hide the "rerun of" column. 

- Added the grouping option, is only  enabled by default at "Finished Recordings" (but it can be used in the other tabs) and the group are collapsed, so you can expand the needed group and show the recordings.

I wish it will usefull for others.

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -355,6 +355,8 @@ tvheadend.dvr_upcoming = function(panel, index) {
         },
         callback: function(conf, e, store, select) {
             duplicates ^= 1;
+            Ext.getCmp('idnode_grid').getColumnModel().setHidden(3, !duplicates);
+            Ext.getCmp('idnode_grid').getBottomToolbar().changePage(0);
             this.setText(duplicates ? _('Hide duplicates') : _('Show duplicates'));
             store.baseParams.duplicates = duplicates;
             store.reload();
@@ -570,7 +572,8 @@ tvheadend.dvr_finished = function(panel, index) {
                 }
             }],
         tbar: [removeButton, downloadButton, rerecordButton, moveButton],
-        selected: selected
+        selected: selected,
+        grouping: true
     });
 
     return panel;
@@ -668,7 +671,7 @@ tvheadend.dvr_failed = function(panel, index) {
         },
         sort: {
           field: 'start_real',
-          direction: 'ASC'
+          direction: 'DESC'
         },
         plugins: [actions],
         lcol: [
@@ -742,7 +745,7 @@ tvheadend.dvr_removed = function(panel, index) {
         },
         sort: {
           field: 'start_real',
-          direction: 'ASC'
+          direction: 'DESC'
         },
         plugins: [actions],
         lcol: [actions],
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -295,6 +295,7 @@ tvheadend.dvr_upcoming = function(panel, index) {
                 list + ',owner,creator' : list) + ',retention,removal';
     var duplicates = 0;
     var buttonFcn = tvheadend.dvrButtonFcn;
+    var columnId = null;
 
     var stopButton = {
         name: 'stop',
@@ -355,7 +356,6 @@ tvheadend.dvr_upcoming = function(panel, index) {
         },
         callback: function(conf, e, store, select) {
             duplicates ^= 1;
-            var columnId = select.grid.colModel.findColumnIndex('duplicate');
             select.grid.colModel.setHidden(columnId, !duplicates);
             select.grid.bottomToolbar.changePage(0);
             this.setText(duplicates ? _('Hide duplicates') : _('Show duplicates'));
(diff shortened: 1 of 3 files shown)


src/webui/static/app/dvr.js
```diff
@@ -355,6 +355,8 @@ tvheadend.dvr_upcoming = function(panel, index) {
         },
         callback: function(conf, e, store, select) {
             duplicates ^= 1;
+            Ext.getCmp('idnode_grid').getColumnModel().setHidden(3, !duplicates);
```
perexg: Find the column by name, don't use hard-coded indexes.

src/webui/static/app/idnode.js
```diff
@@ -1728,19 +1729,37 @@ tvheadend.idnode_grid = function(panel, conf)
...
             baseParams: params,
             autoLoad: true,
             id: 'uuid',
-            totalProperty: 'total',
-            fields: fields,
-            remoteSort: true,
+            remoteSort: true, //  We lost multi sort at server side :-(, maybe perexg has a better idea
+            reader: groupreader,
+            remoteGroup: true,
+            groupField: 'disp_title',
```
perexg: This code is shared among more callers. Use 'conf' variable to determine the grouping column.

src/webui/static/app/idnode.js
```diff
@@ -2086,16 +2105,21 @@ tvheadend.idnode_grid = function(panel, conf)
 
         plugins.push(filter);
         var gconf = {
+            id: 'idnode_grid',
```
perexg: Do not hardcode anything here - use conf.id or so.. Perhaps, this change is not required at all - 'select' might contain a pointer to the column model in the callback.

src/webui/static/app/idnode.js
```diff
@@ -2113,7 +2137,10 @@ tvheadend.idnode_grid = function(panel, conf)
             page.changePage(0);
         });
         if (conf.beforeedit)
-          grid.on('beforeedit', conf.beforeedit);
+            grid.on('beforeedit', conf.beforeedit);
+
+        if (fields.indexOf('duplicate') !== -1)
```
perexg: Hardcoded field name.

pablozg:
I think now all is fixed. Could you take a look? Thanks.

src/webui/static/app/idnode.js
```diff
@@ -1728,19 +1729,41 @@ tvheadend.idnode_grid = function(panel, conf)
...
+               sorters: [{field : 'disp_title', direction : 'ASC'},
+                         conf.sort ? conf.sort : null],
+               direction: 'ASC'
+            },*/
             pruneModifiedRecords: true,
             sortInfo: conf.sort ? conf.sort : null
         });
 
+        if (!conf.grouping)
+            store.clearGrouping();
```
perexg: This is not probably required when groupField is not set.

src/webui/static/app/dvr.js
```diff
@@ -570,7 +574,9 @@ tvheadend.dvr_finished = function(panel, index) {
                 }
             }],
         tbar: [removeButton, downloadButton, rerecordButton, moveButton],
-        selected: selected
+        selected: selected,
+        grouping: true,
+        groupField: 'disp_title'
```
perexg: I believe that grouping should be optional (another button?).

src/webui/static/app/idnode.js
```diff
@@ -1728,19 +1729,41 @@ tvheadend.idnode_grid = function(panel, conf)
         var params = {};
         if (conf.all) params['all'] = 1;
         if (conf.extraParams) conf.extraParams(params);
-        store = new Ext.data.JsonStore({
+        groupreader = new Ext.data.JsonReader({
```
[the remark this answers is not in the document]
pablozg: A JsonStore will be automatically configured with a JsonReader, but the GroupingStore not, so we need to create and pass it as config parameter.

src/webui/static/app/idnode.js
```diff
@@ -1728,19 +1729,41 @@ tvheadend.idnode_grid = function(panel, conf)
...
             baseParams: params,
             autoLoad: true,
             id: 'uuid',
-            totalProperty: 'total',
-            fields: fields,
-            remoteSort: true,
+            remoteSort: true, //  We lost multi sort at server side :-(, maybe perexg has a better idea
+            reader: groupreader,
+            remoteGroup: true,
+            groupField: groupField,
```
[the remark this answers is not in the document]
pablozg: It's true, I will change it.

src/webui/static/app/dvr.js
pablozg: In case of series recordings, it's better to have it grouped by title, as you can see in this capture, but you can choose make a group from other column and the next time you enter to the tab will continue using the last group used, so I think is not needed create a new button, because the users can configure the grid as they like. 

finished

pablozg:
Changes made. ;-)

mpmc:
I like these changes, makes the GUI look much nicer IMO. :) 

However I agree with @perexg, grouping should be optional & having a group button is more user friendly (IMO) rather than having it hidden in the drop-down.

With that said, @pablozg please do continue with these usability improvements =D, and if possible, please update the help screenshots to include your awesome features :)

---

My overall thoughts though is there's very little point in making huge changes to the webui as it's under discussion for replacement!

src/webui/static/app/dvr.js
perexg: Believe me, it won't work on my "production" tvh at all, because the titles are different. Broadcasters puts series and episode info to the title here. Yes, recent EIT scraping might help, but we should expect all situations.

pablozg:
Button added for finished recordings, when enable you can group by field clicking on the field header.

perexg:
Merged and added changes in d61d2006e8ba39a01c22dfe6c28cf36893092ff8 .

PR #1071, opened 2017-12-28
eit: allow scraper regexes to be marked as filters (#4818)

banburybill:
Introduce an extended scraper regex syntax; as well as a string, a regex
can be specified as a map. The map must have an entry "pattern" with the
regex pattern. It may also have an entry "filter", with a numeric value.
If this value is not 0, the regular expression is a filter.

If a filter regular expression matches the input text, the result of the
match replaces the input text, and processing continues from the next
regular expression with that new input text.

Issue: 4818

### DIFF
--- a/data/conf/epggrab/eit/scrape/README
+++ b/data/conf/epggrab/eit/scrape/README
@@ -24,7 +24,7 @@ of the match is the contents of all the sub-patterns in the regular
 expression concatenated together.
 
 For each EPG episode, the title, description and summary are matched
-in turn against the season_num, episode_num, airdate and is_new regexes.
+in turn against the season_num, episode_num, airdate and is_new regex lists.
 
 - season_num converts the contents of the match result to an integer,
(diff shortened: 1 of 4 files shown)


perexg:
Looks good. Perhaps, you can add the language selector using this data structure, too, like:

<pre>
     {
         "pattern": "^[.][.][.][^:.?!]*[.:?!] +(.*)",
         "lang": "cze"
     },
     {
         "pattern": "^[.][.][.][^:.?!]*[.:?!] +(.*)",
         "lang": ["cze","slo"]
     },
</pre>

banburybill:
That's an idea. Pass the text language into the pattern matcher, and if a pattern has one or more languages specified, ignore the pattern and skip to the next unless one of the pattern languages matches the text language.

I'll give it a go. Probably makes sense as a separate PR, though.

perexg:
Merged. Thanks.

PR #1072, opened 2017-12-28
eit: add language identifier for scraper regexes (#4820)

banburybill:
Allow scraper regexes to be tagged with a single language identifier or
a list of language identifiers.

Tagged regexes will only be used against input text in languages
matching the tag. Otherwise the regex is skipped.

I don't have access to EIT with multilingual text. I have tested the mechanism, but it would be nice to confirm it works under live firing conditions. :-)


perexg:
Merged. Thanks.

PR #1075, opened 2018-01-02
transcode: set AVFILTER_AUTO_CONVERT_ALL, issue #4797

dkonerma:
Hope this has no unwanted side effects, but this resolves issue #4797 for me. See also https://tvheadend.org/boards/5/topics/30402.


perexg:
Applied. Thanks.

PR #1076, opened 2018-01-03
HTSP: Expose recording files' start and stop time.

ksooo:
The start/stop times of the recording files are needed by Kodi's tvheadend PVR client addon (pvr.hts) to calculate proper recording play times.

It would be great if this PR could also be applied to the 4.2 branch as without this fix Kodi can only work with tvh master (4.3).


ksooo:
>  Needs HTSP version bump?

No. Clients can simply check whether the two new fields ("start" and "stop") are contained in the HTSP message and react accordingly.

Jalle19:
Okay

perexg:
Merged. Thanks.

PR #1078, opened 2018-01-15
Fix: Rerecord of an event added as "previously recorded", typo mismatch and missing "Episode" column in webui 

pablozg:
Since we added the "previously recorded" option, I have been observing that an event added as previously recorded is rescheduled for record if the start time change to an early time. An example:

Event added as previously recorded with a start time of "14/01/2018 15:15:00" (Spanish date format),
If the next time we reimport the xmltv file it has a start time change like: "14/01/2018 15:14:00" then the event will be rescheduled.

After commit #37db5d9 episode column name was changed but not updated in webui.


perexg:
Merged. Thanks.

PR #1080, opened 2018-01-22
src/epggrab/module.c: Presumed typo in xmltv unix socket listen() call

ewestbrook:
With "mod->sock", the xmltv epg module fails startup with invalid file descriptor, and connections fail with "connection refused".  With "sock", connections succeed and data is accepted.


perexg:
Merged. Thanks.

PR #1081, opened 2018-02-04
Add feature #4902, the option to edit extratext field (subtitle,summa…

pablozg:
…ry or description) and the episode field, because sometimes the recordings don't have the proper order (one missed, no free adapter, etc) and is good to have the option to add the "Season XX.Episode XX" text by hand if it not set previously.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -4005,7 +4005,7 @@ const idclass_t dvr_entry_class = {
       .desc     = N_("Episode number/ID."),
       .set      = dvr_entry_class_disp_episode_set,
       .get      = dvr_entry_class_disp_episode_get,
-      .opts     = PO_NOSAVE | PO_ADVANCED,
+      .opts     = PO_HIDDEN | PO_NOSAVE | PO_ADVANCED,
     },
     {
       .type     = PT_STR,
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -3235,36 +3235,20 @@ dvr_entry_class_disp_extratext_set(void *o, const void *v)
 {
   dvr_entry_t *de = (dvr_entry_t *)o;
   const char *lang = idnode_lang(o);
-  const char *s = "";
-  int field = 0;
-  v = tvh_str_default(v, "UnknownExtraText");
-  if (de->de_subtitle){
-    s = lang_str_get(de->de_subtitle, lang);
-    field = 1;
+  v = tvh_str_default(v, "?????");
+
+  if (de->de_subtitle && lang_str_get(de->de_subtitle, lang)[0] != '\0'){
+    return lang_str_set(&de->de_subtitle, v, lang);
   }
-  if (s == NULL || s[0] == '\0'){
-    s = lang_str_get(de->de_summary, lang);
-    field = 2;
+  if (de->de_summary && lang_str_get(de->de_summary, lang)[0] != '\0'){
+    return lang_str_set(&de->de_summary, v, lang);
   }
-  if (s == NULL || s[0] == '\0'){
-    s = lang_str_get(de->de_desc, lang);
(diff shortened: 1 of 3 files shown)


src/dvr/dvr_db.c
```diff
@@ -3230,6 +3230,43 @@ dvr_entry_class_disp_extratext_get(void *o)
   return &prop_ptr;
 }
 
+static int
+dvr_entry_class_disp_extratext_set(void *o, const void *v)
+{
+  dvr_entry_t *de = (dvr_entry_t *)o;
+  const char *lang = idnode_lang(o);
+  const char *s = "";
+  int field = 0;
```
perexg: It should be probably 1 as default (to edit the empty extra text).

src/dvr/dvr_db.c
```diff
@@ -3230,6 +3230,43 @@ dvr_entry_class_disp_extratext_get(void *o)
...
+  v = tvh_str_default(v, "UnknownExtraText");
+  if (de->de_subtitle){
+    s = lang_str_get(de->de_subtitle, lang);
```
perexg: The lang_str_set() already returns 0 (no string change) or 1 (string changed), so you can do 'return lang_str_set(de->de_substitle, v, lang);' here.

src/dvr/dvr_db.c
```diff
@@ -3230,6 +3230,43 @@ dvr_entry_class_disp_extratext_get(void *o)
...
 }
 
+static int
+dvr_entry_class_disp_extratext_set(void *o, const void *v)
+{
+  dvr_entry_t *de = (dvr_entry_t *)o;
+  const char *lang = idnode_lang(o);
+  const char *s = "";
+  int field = 0;
+  v = tvh_str_default(v, "UnknownExtraText");
```
perexg: Use a more universal (no international dependency) string here, like '?????'.

src/dvr/dvr_db.c
```diff
@@ -3967,7 +4005,7 @@ const idclass_t dvr_entry_class = {
       .desc     = N_("Episode number/ID."),
       .set      = dvr_entry_class_disp_episode_set,
       .get      = dvr_entry_class_disp_episode_get,
-      .opts     = PO_RDONLY | PO_HIDDEN | PO_NOSAVE,
+      .opts     = PO_HIDDEN | PO_NOSAVE | PO_ADVANCED,
```
perexg: Why this was changed to 'advanced'?

src/dvr/dvr_db.c
```diff
@@ -3715,7 +3752,8 @@ const idclass_t dvr_entry_class = {
       .name     = N_("Extra text"),
       .desc     = N_("Subtitle, summary or description of the program (if any) (display only)."),
       .get      = dvr_entry_class_disp_extratext_get,
-      .opts     = PO_RDONLY | PO_NOSAVE,
+      .set      = dvr_entry_class_disp_extratext_set,
+      .opts     = PO_NOSAVE | PO_ADVANCED,
```
perexg: I don't think that this should be advanced.

src/dvr/dvr_db.c
```diff
@@ -3230,6 +3230,27 @@ dvr_entry_class_disp_extratext_get(void *o)
...
 }
 
+static int
+dvr_entry_class_disp_extratext_set(void *o, const void *v)
+{
+  dvr_entry_t *de = (dvr_entry_t *)o;
+  const char *lang = idnode_lang(o);
+  v = tvh_str_default(v, "?????");
+
+  if (de->de_subtitle && lang_str_get(de->de_subtitle, lang)[0] != '\0'){
```
pablozg: I don't know if this is the best method to do the compare "lang_str_get(de->de_subtitle, lang)[0]" but it works.

perexg:
Merged. Thanks.

PR #1083, opened 2018-02-14
eit: combine title and subtitle when scraping title with ' % ' (#4873)

banburybill:
Currently title and subtitle are combined by joining with a single space. If the scrape is attempting to just modify the title, that's a problem because you don't know where the title ends.

So instead join with ' % '. % isn't a regex metachar, so this is convenient for use in regexes.

Update scraper test engine and UK regexes to match.

Issue: #4873


perexg:
Merged. Thanks.

PR #1085, opened 2018-02-16
Trace broken since 6ee756a829ef6ce2e89daa7dd1737164a6697e29

lambdadelivery:
Bug in bitops.h


perexg:
Fixed in ee368a02125006c8d2bea0c0cdfed44edc131c05 (a bit differently).

PR #1091, opened 2018-02-24
SAT>IP: refactoring of IPTV inputs

Trujulu:
http://www.tvheadend.org/issues/4784
By Mono Polimorph


perexg:
Declined, see the conversation in the issue - http://www.tvheadend.org/issues/4784 .

PR #1092, opened 2018-02-24
SAT>IP: Print list of PIDs used in the SUBSCRIPTION section of the UI

Trujulu:
http://www.tvheadend.org/issues/4934
By Mono Polimorph

### DIFF
--- a/src/subscriptions.c
+++ b/src/subscriptions.c
@@ -1015,6 +1015,7 @@ subscription_create_msg(th_subscription_t *s, const char *lang)
   profile_t *pro;
   char buf[284];
   const char *state;
+  htsmsg_t *l;
 
   htsmsg_add_u32(m, "id", s->ths_id);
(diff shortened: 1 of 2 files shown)


perexg:
The implementation is not correct. The active pid list per service is managed in the mpegts input - mm->mm_pids, the service pointer is in owner fields (mpegts_pid_t).

So, we need probably another callback in service_t which returns the htsmsg_t list and put this code to mpegts_service.c and/or mpegts_input.c.

Also, s->ths_service might be NULL (there is already check for this).

perexg:
I mean that the implementation is not correct for standard services - if we add such feature, the subscribed PIDs should be visible for all subscriptions (in my opinion - including the RAW services like epggrab).

perexg:
Implemented a bit differently in the master. Commit 0c50a763f630f91d5a0ee6687890b631c619a21f .

PR #1093, opened 2018-02-24
SAT>IP: Add Transport Protocol to Input Name

Trujulu:
http://www.tvheadend.org/issues/4938
By Mono Polimorph


perexg:
The implementation is incomplete - if you change TCP settings, the name is not changed. Also, user may freely edit this name, so I guess we don't need this.

PR #1094, opened 2018-02-24
fix: When a new dvr is created and the epg info have onscreen episode…

pablozg:
… is not passed to the new entry


perexg:
Merged. Thanks.

PR #1095, opened 2018-02-27
Fix potential nullptr access found by gcc 7.3

Logout22:


src/idnode.c
```diff
@@ -1595,7 +1595,7 @@ idnode_slist_set ( idnode_t *in, idnode_slist_t *options, const htsmsg_t *vals )
       HTSMSG_FOREACH(f, vals) {
         if ((s = htsmsg_field_get_str(f)) == NULL)
           continue;
-        if (strcmp(s, o->id))
+        if (s != NULL && strcmp(s, o->id))
```
perexg: It's really obvious false positive. 's' cannot be NULL - see the previous condition. You should report this to gcc .

Logout22: True. I did not think this through. Maybe I can open up a case with gcc.

PR #1097, opened 2018-03-05
Insert card into card list before setting EMM state (issue #4988)

LubosD:
This fixes the bug I described here: https://tvheadend.org/issues/4988

`cc_emm_set_allowed()` expects the card to be already present in `cc->cc_cards`.


perexg:
Merged. Thanks.

PR #1099, opened 2018-03-12
Update dvr_db.c

thromer:
Fix typo -- &&& should have been &&.

Caught by compiling with clang:
src/dvr/dvr_db.c:2099:16: error: address of 'e->subtitle' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]


perexg:
Merged. Thanks.

PR #1100, opened 2018-03-18
Add request feature #5009

pablozg:
I think it can useful for everyone.

### DIFF
--- a/filelist.txt
+++ b/filelist.txt
@@ -1,0 +1,6 @@
+/usr/src/tvheadend/../tvheadend_4.3-1160~ga50eccd-dirty_armhf.deb
+/usr/src/tvheadend/../tvheadend-dbg_4.3-1160~ga50eccd-dirty_armhf.deb
+/usr/src/tvheadend/../tvheadend_4.3-1160~ga50eccd-dirty_armhf.changes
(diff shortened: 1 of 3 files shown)


pablozg:
It's not correct, the field "channel" shows the number and the name, so don't worry.

channel

ProfYaffle:
Okay - I'll relax a little, so thanks! - although doesn't that screw up alpha sorting by channel name?

pablozg:
The field sorting is not affected. ;-)

ProfYaffle:
I'll shut up and go away, then :grin:

perexg:
It's incomplete. You should handle the empty case (when empty uuid is returned in the channel field - it's the case when channel is removed, but the recordings have saved the channel name in 'channelname' field which is used currently).  See 'renderer:' functions in src/webui/static/app/dvr.js for examples.

pablozg:
Before to make the changes over 'dvrdb.c' file, I tried to do the dvr.js aproach with the rendered option, but only was able to get the field but not the number, now I have to learn of your code for the next time. Thanks perexg.

P.D. Feel free to close this PR

perexg:
Merged your 222c8541b981e6db6fa9f2d59f2553235bbc1c4b and fixed missing removed channel names in 163322179186fcadbef18145ad10a7818e2b094a .

PR #1102, opened 2018-03-23
Add request feature #5022

pablozg:
I think this request is useful and make more easy to do changes from the grid without to have to enter in the edit form.

Perexg I wish you are agree with me.

Thanks.


perexg:
Merged.

PR #1103, opened 2018-03-25
Cosmetic fix for dvr entry info (webui)

pablozg:
Inside upcoming tab, the show info popup, shows in the autorec field the rule name + (null) instead only the name ("Blindspot (null)") when the autorecord comment field is empty and ("Blindspot ()") when you change it from a text to an empty field.


perexg:
Merged with the fix: 0b223e5a1d4e8d6619a163f8ae43eea17722e858

PR #1106, opened 2018-03-31
Fix server crash in satip_server_http_page

dave-p:
Accessing http://your.ip:9981/satip_server/ crashes TVHeadend. The fix also returns Bad Request for unexpected filenames.


perexg:
Merged. Thanks.

PR #1109, opened 2018-04-05
Webui: Correctly sort numbers (status)

Trujulu:


src/webui/static/app/status.js
```diff
@@ -42,16 +42,16 @@ tvheadend.status_subs = function(panel, index)
             totalProperty: 'totalCount',
             fields: [
                 { name: 'id', sortType: stype },
-                { name: 'hostname', sortType: stype },
+                { name: 'hostname', sortType: stypei },
```
Trujulu: I'm not completely sure if hostname should be an integer or not, since it could be a domain name in some cases (or not? NOT tested)

perexg:
Merged although the 'hostname' sorting is not really an integer. It may be IPv4 or IPv6 address - we need probably a special sorting handler for this.

PR #1111, opened 2018-04-10
 descrambler: fix build with ddci disabled 

alanswanson:
Per the commit message, the  build fails in src/input/mpegts/mpegts_service.c with an undefined reference to dvbcam_is_ddci in mpegts_service_update_slave_pids() when ddci is disabled (as no descrambling needed for my usage).


perexg:
Thanks for the hint. The correct fix is in 55f9d626365057d723695de0b890d015d9749ddf .

PR #1113, opened 2018-04-19
webui: Added partial support for Kodi label formatting

pablozg:
1- Added support for:

[B]bold[/B] - bold text.
[I]italics[/I] - italic text.
[COLOR red]red text[/COLOR] - colored text.
[UPPERCASE]uppercase text[/UPPERCASE] - all text is made uppercase e.g. UPPERCASE TEXT
[LOWERCASE]lowercase text[/LOWERCASE] - all text is made lowercase e.g. lowercase text
[CAPITALIZE]capitalized text[/CAPITALIZE] - first letter of every word is capitalized (Isengard onward) e.g. Capitalized Text
[CR] - carriage return (line break).

2- Added an auto height function in order to maximize the info size at screen and vertical scroll bar in DVR info.


ksooo:
Does this make sense? Isn't it up to Kodi to format the strings to fit it's needs? What about other, non-Kodi clients? What do they make out of this "Kodi-formatting"?

pablozg:
Non-kodi clients will not be affected, the code will only replace the kodi tags to html tags if they exist in the description field and only on the web client side, making it easier to use color texts if you are using a xmltv file for kodi or you can add these tag to your actual file and get the color improvement. 

That is just my opinion.

ksooo:
tvheadend knowing anything about its clients ("kodi tags" ?!) feels conceptionally wrong.

pablozg:
It's my fault, when I say 'tags' I want to say 'keywords' included in [https://kodi.wiki/view/Label_Formatting](`https://kodi.wiki/view/Label_Formatting`), so anyone can use it to make a more aesthetic view of the info included in the description field.

For me is usefull because I can implement it in my own grabber and still be compatible with others xmltv files made for be used with Kodi.

ksooo:
No idea what @perexg thinks about this, my opinion is that it is wrong to implement kodi-specific frontend stuff in the tvheadend backend.

mpmc:
@ksooo even I don't like it :rofl:

pablozg:
I respect your opinion, but I think that the EPG and DVR info is a part of the tvheadend frontend, and with this code we offer a new option to the frontend users that use the webui to manage the recordings, epg, etc. and Kodi to see the content as me. 

And as I said before, this code don't affect to the normal use of tvheadend if you are not using this feature. I think that is a good idea have the choice to use colored text in the EPG.

Then how can we add a color function to the epg? maybe with a new set of keywords?

I will respect the decision made by @perexg . :smile:

guiniguada:
I´m using EPG guides from .xml file with label formatting from some time ago... I´m also interested to see this feature included because I use many times the EPG section in the tvheadend itself.

nickiman:
@ksooo Could you please explain why is wrong implement this feature in tvheadend?

As an end user, I think it´s more clear see the EPG with the tag functionality **not only in kodi** **also in tvheadend** 

And this commits tries to avoid this situation... **look the aspect of the EPG´s broadcast details in tvheadend**:
imagen

And now, take a look with label formatting...:
imagen

From my personal point of view it look much better and attractive.

Best regards

ProfYaffle:
Forgive me for being dumb here - it's a habit of mine, sadly :)

Are these tags in a widely-available XML stream that's being fed to tvheadend and then on to Kodi? So they're formatted by the source and arrive fully-formed? In that case, it's nothing to do with Kodi, and merely a case of whether tvheadend can understand the tags, and either use them or strip them out.

Or are these tags generated by Kodi or some other post-processing script acting on the XML feed? In which case, you shouldn't have to be changing the tvh web "client" to deal with what is effectively a malformed feed: whatever is inserting those tags should have an option to produce a "raw" equivalent that gets used by default, as those formatting commands are very highly client-specific.

Either way, formatting shouldn't be mandatory. As I read the code, it's a simple search-and-replace, so wouldn't affect any raw text that doesn't have those tags, which is fine, but it's still a question as to where the tags are coming from.

nickiman:
@ProfYaffle look the thread where is written the original request. 
https://tvheadend.org/issues/5077
You will find also posted the .xml with the label formatting used.

Best regards

Diablobait:
Im very interesting on this feature pls

ProfYaffle:
@nickiman Thanks for that. 

So, my simple view... if it is common practice to have an EPG that universally contains formatting information, then supporting that in any client that uses that data is acceptable. However, if this is actually a Kodi-specific EPG that has the formatting inserted especially for Kodi, perhaps as an option when fetching it, then it's not the job of another client to interpret that. I suspect the same could be achieved by simply sed'ing the file before import to tvheadend, after all...

nickiman:
@ProfYaffle I see your point... but let me give you more details about this.

There´s a big amount of users using this type of EPG...  some with Tvheadend as PVR client but also another ones feeding the EPG in other PVR clients like IPTV Simple client... **all of them using Kodi as the main frontend.**

We are aware that the "Label Formatting" capability it´s only available in Kodi but, hey! thanks to @pablozg now it´s possible to enjoy it in Tvheadend also.

Our intention was to have the chance to make up the "Broadcast Details" section in Tvheadend in the same way as Kodi does... just to have a readable EPG in both of them Kodi and Tvheadend itself.

Last, as you mention... yes for sure we could feed the .xml file directly in kodi but, in scenarios where we have two or more clients... the best option is leave to Tvheadend in charge of the EPG and feed it to all clients connected on it.

Best regards

mpmc:
Here's an idea, make it optional?

I suppose it could be added as an option in the base config? "Use Kodi label formating (broadcast details dialog)" under the "Web interface settings".

nickiman:
@mpmc Yes, could be...
But, why not leave it always enabled in background? I mean, without any option to enable/disable the label formatting?
It´s just to avoid any other thing that I´m missing?

I tell you that because, at the end, there will be two types of EPG: 
normal (with plain text) and this one (with **keywords** for label formatting)

Since to activate the label formatting is needed an EPG file with that keywords to be shown... I think could be better leaving this option let me say, always enabled. 
In this way users will don´t have to care of what kind of EPG are feeding in Tvheaded or to activate any option to show the label formatting. 

Keep in touch!

alebupal:
I find it a very useful option, I look forward to it!
a greeting!

perexg:
Pablo, please, make this parser optional (through the tvh's configuration).

PR #1114, opened 2018-04-24
webui: Added custom date format field in config tab

pablozg:
Now is possible to write a custom mask for the dates and use the select language to make it locale depend.


Trujulu:
Awesome work.

P.S.: Is code protected against wrong introduced values?

pablozg:
I added a syntax check, but now I don't have internet connection in my computer, so I will update the PR later.

Trujulu:
The change is not working for the Read Only Info dialog, for services 

image

PR #1117, opened 2018-04-28
Virtual channel tuning support (ATSC and CableCARD networks)

rpcameron:
This adds the ability for muxes to have virtual channels defined to give an alternative means of tuning information; it is presently only exposed for ATSC-T and ATSC-C networks.

Additionally, if virtual channel tuning information is present in a mux, HDHomeRun tuners will use that information as the preferred tuning method, ignoring any frequency and other tuning information present in the mux. This makes it possible to natively use the HDHomeRun Prime CableCARD tuner as an ATSC-C device using the tvhdhomerun drivers, as opposed to IPTV sources.

Channel number and service/provider information is also retrieved from the mux and passed on to any service created from a mux tuned by virtual channel number if the service stream does not have that information in it.

For more information, please see: http://tvheadend.org/boards/5/topics/32553

### DIFF
--- a/src/input/mpegts/dvb.h
+++ b/src/input/mpegts/dvb.h
@@ -611,7 +611,7 @@ typedef struct dvb_mux_conf
   dvb_fe_delivery_system_t    dmc_fe_delsys;
   dvb_fe_modulation_t         dmc_fe_modulation;
   uint32_t                    dmc_fe_freq;
-  dvb_fe_vchant_t	          dmc_fe_vchan;
+  dvb_fe_vchan_t	          dmc_fe_vchan;
   dvb_fe_spectral_inversion_t dmc_fe_inversion;
   dvb_fe_rolloff_t            dmc_fe_rolloff;
   dvb_fe_pilot_t              dmc_fe_pilot;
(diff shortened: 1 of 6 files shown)


perexg:
I don't like this change for the generic ATSC muxes. It's too much hdhomerun specific and the "real" tuning parameters are unused for those virtual muxes. It would be better to define new class like dvb_network_atsc_t_virthdh_class to describe the virtual muxes for UI (invalidate the tuning parameters - use PT_NONE and define the virtual parameters only).

perexg:
Also, the developent is in the master branch (the changes might be cherry-picked to the 4.2 relase, too).

rpcameron:
While the "real" tuning parameters are unused when tuning via virtual channels, setting a virtual channel number to 0 will then allow the mux's "real" tuning configuration to be used. So, while you can only tune one way or the other, the mux itself can contain the information on how to tune both.

While this may be HDHomeRun specific, it is not necessarily so. If there are other tuners that allow tuning by virtual numbers, there is nothing stopping the tuning routines from those other drivers from using the dmc_fe_vchan.

While I understand that development occurs against the master branch, in general I no longer run master versions of TVH because of past instabilities. I may pull one tuner off my main server to move this codebase to master, and investigate setting it up as a new network type.

rpcameron:
As per @perexg 's requests, I have re-implemented this as a new CableCARD network type done against master: #1120 

Therefore, I'm closing this PR.

PR #1120, opened 2018-05-14
CableCARD and HDHomeRun Prime support

rpcameron:
As previously requested, I have changed the generic virtual channel support to instead be a new CableCARD network. Also, this development has been done against the master branch (although I am maintaining a branch against release/4.2 as well)


src/input/mpegts/dvb.h
```diff
@@ -633,6 +640,7 @@ typedef struct dvb_mux_conf
     dvb_ofdm_config_t         dmc_fe_ofdm;
     dvb_isdbt_config_t        dmc_fe_isdbt;
   } u;
+  dvb_fe_vchan_t              dmc_fe_vchan;
```
perexg: Why those new data are outside the union?

rpcameron: Originally because my code allowed for the use of virtual channel tuning across arbitrary network types. However, with the introduction of dvb_network_cablecard*, there's no real reason why dvb_fe_vchan cannot become dvb_cablecard_config and be moved inside the union.

PR #1128, opened 2018-06-08
webui: dvr/epg popup window autoresize to actual resolution

pablozg:
I forgot add the auto resize to adapt popup window to actual resolution.


f1vefour:
Seems like a reasonable change, why not merge?

perexg:
Merged.

PR #1130, opened 2018-06-18
Fix DVR crash (issue #5087)

irgendsontyp:
I could reproduce the crash by explicitely setting subtitle_buf to NULL. Under normal circumstances, that pointer (and the others I fixed) can also be NULL. The problem here was that those pointers were dereferenced, even if they were NULL. This leaded to undefined behaviour (a crash with SIGSEGV in this case).


perexg:
Fixed in ea9159ecbdc695a4e717c420c2dd7b4d926cb2b9 .

PR #1131, opened 2018-06-20
Fix creation of CA Client type 'AES128 ECB Constant Code Word' (Issue #5119)

irgendsontyp:
Fixes https://tvheadend.org/issues/5119

### DIFF
--- a/src/descrambler/caclient.c
+++ b/src/descrambler/caclient.c
@@ -120,7 +120,8 @@ caclient_create
 #if ENABLE_CONSTCW
   if (c == &caclient_ccw_csa_cbc_class ||
       c == &caclient_ccw_des_ncb_class ||
-      c == &caclient_ccw_aes_ecb_class)
+      c == &caclient_ccw_aes_ecb_class ||
+      c == &caclient_ccw_aes128_ecb_class)
     cac = constcw_create();
 #endif
   if (cac == NULL) {


irgendsontyp:
Btw., what happens to the corresponding issue in the tvheadend bug tracker if you accept the PR? https://tvheadend.org/issues/5119 in this case.

perexg:
BTW: Write 'fixes #5119' or 'issue #5119' to the subject of the commit to close or connect to the issue in the bug tracker next time. Thanks.

irgendsontyp:
@perexg Does the bug tracker automatically close the issue when I add "fixes #..." or "issue #..." to the commit? 

The corresponding bug for my other PR was not closed, too (https://tvheadend.org/issues/5126). 

Also, your commit "ea9159ecbdc695a4e717c420c2dd7b4d926cb2b9" fixes https://tvheadend.org/issues/5087 but 5087 is still in "New" state.

perexg:
"fixes #" closes the ticket as fixed and "issue #" just assigns the commit to the bug

irgendsontyp:
Is it better to use "issue #" then, as we as non-commiter contributors can never be sure for the PR to be accepted? Thus, linking the commit to the bug is always right but closing the issue before the commit is merged doesn't make sense to me.

perexg:
Note that the bug linking is proceed when the commit is merged (validated), thus you can mark the commit as you like. The "fixes" word saves my time when it's an obvious code fix.

PR #1133, opened 2018-06-23
configure: check for strlcat & strlcpy

bkuhls:
Building tvheadend with uclibc and musl fails:

src/tvh_string.h:50:22: error: static declaration of 'strlcpy' follows non-static declaration
 static inline size_t strlcpy(char *dst, const char *src, size_t size)

src/tvh_string.h:61:22: error: static declaration of 'strlcat' follows non-static declaration
 static inline size_t strlcat(char *dst, const char *src, size_t count)

because they provide strlcat & strlcpy:
https://sourceware.org/glibc/wiki/strlcpy

This patch adds configure checks and makes the implementation in
tvh_string.h optional, the configure log looks like this:

glibc
  checking for cc strlcat ...                       fail
  checking for cc strlcpy ...                       fail

musl
  checking for cc strlcat ...                       ok
  checking for cc strlcpy ...                       ok

uclibc
  checking for cc strlcat ...                       ok
  checking for cc strlcpy ...                       ok

### DIFF
--- a/configure
+++ b/configure
@@ -159,6 +159,20 @@ else
   COMPILER=gcc
 fi
 
+check_cc_snippet strlcat '#include <string.h>
+int test(int argc, char **argv) {
+  char dst[10];
+  strlcat("test", dst, sizeof(dst));
+  return 0;
+}'
+
+check_cc_snippet strlcpy '#include <string.h>
+int test(int argc, char **argv) {
+  char dst[10];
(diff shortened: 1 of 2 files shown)


bkuhls:
Patch was committed as https://github.com/tvheadend/tvheadend/commit/e26ce3390a5ca88f3fcee8a7fa0e4bb1b078e538

PR #1134, opened 2018-06-24
webui: Feature #5136, Add Filmaffinity search in broadcast details

pablozg:
I added the request feature #5136, if you think is usefull, please merge it.

Thanks.

### DIFF
--- a/src/epggrab.h
+++ b/src/epggrab.h
@@ -246,8 +246,10 @@ struct epggrab_ota_mux
   uint8_t                            om_complete;     ///< Has completed a scan
   uint8_t                            om_requeue;      ///< Requeue when stolen
   uint8_t                            om_save;         ///< something changed
+  uint8_t                            om_detected;     ///< detected some activity
   mtimer_t                           om_timer;        ///< Per mux active timer
   mtimer_t                           om_data_timer;   ///< Any EPG data seen?
+  mtimer_t                           om_handlers_timer; ///< Run handlers callback
   int64_t                            om_retry_time;   ///< Next time to retry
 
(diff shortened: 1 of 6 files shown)


src/webui/static/app/dvr.js
```diff
@@ -119,27 +119,41 @@ tvheadend.dvrDetails = function(uuid) {
...
         }
 
-        function searchTheTVDB(){
+        function searchTheTVDB() {
             window.open('http://thetvdb.com/?string='+
                         encodeURIComponent(title)+'&searchseriesid=&tab=listseries&function=Search','_blank');
         }
 
+        function searchFilmAffinity() {
+            window.open('https://www.filmaffinity.com/es/search.php?stext=' +
```
CvH: is it a good choice to hard coding the spanish version ?

pablozg: My fault, I will add a language check to use the English or Spanish version.

mpmc: IMO, I would just use en & not bother with any checks. The reasons for this are..

- People will ask  "why isn't <insert lang here> supported in tvheadend, it's supported on the site."
- The available languages (on the site) may change at any point.

pablozg: Filmaffinity it's a spanish page, it's support two languages (english and spanish), by default now it's locale dependent, so will use spanish version only if you set it as language in user config.

perexg:
Could we use a combobox to select the search type? It seems that we end with a long bar of search icons...

pablozg:
OK. I'll take a look.

PR #1136, opened 2018-07-25
debian: always use gzip compression for deb files (bintray issue) - fixes #5091

mpmc:
This fixes Bintray barf about malformed deb files (it seems to have issues with tar.xz files).

@perexg This needs to be backported to release/4.2 as well :)


mpmc:
This has been merged :+1:

PR #1139, opened 2018-08-03
implement netstream 4c

mpietruschka:
This pull request implements the geniatech netstream 4c

### DIFF
--- a/src/input/mpegts/satip/satip_frontend.c
+++ b/src/input/mpegts/satip/satip_frontend.c
@@ -2065,6 +2065,7 @@ satip_frontend_hacks( satip_frontend_t *lfe )
     lfe->sf_play2 = 1;
   } else if (strstr(sd->sd_info.modelname, "EyeTV Netstream 4C")) {
     lfe->sf_specinv = 1;
+    lfe->sf_pass_weight = 0;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/input/mpegts/satip/satip_frontend.c
```diff
@@ -216,6 +216,14 @@ const idclass_t satip_frontend_class =
       .opts     = PO_ADVANCED,
       .off      = offsetof(satip_frontend_t, sf_pass_weight),
     },
+    {
+      .type     = PT_BOOL,
```
perexg: This should be tri-state: do-not-use, 0 or 1

mpietruschka: What do you mean with tri-state? pdids_deladd is set with 0/1, too.

"sd->sd_pids_deladd = 0;"

F4FNlR: tri-state have the following possible value: not set, disable, enable
I'am not a tvheadend-developer but found this via Google.
I'am very interested in this change because I have two netstream 4c, currently running via minisatip.

mpietruschka: Of course I understand the tri-state :) What I didn't understand is how to implement the tri-state. I couldn't find a tri-state configuraton type (PT_BOOL, ...). All that I found stored their boolean states as 1 or 0.

mpietruschka: @F4FNlR  if you want to use the netstream 4c instantly, then you can use my fork...

https://github.com/mpietruschka/tvheadend/tree/release/4.2
https://github.com/mpietruschka/tvheadend

perexg:
Merged.

PR #1140, opened 2018-08-23
Tvshows/tvmovies should use separate config variables for $q

azlm8t:
During tidyup (ea9159), the tvmovies incorrectly used tvshows as its directory name when using the "$q" tag for recordings. So now _dvr_get_tvmovies_subdir returns tvmovies and new function _dvr_get_tvshows_subdir returns tvshows config variables.


perexg:
Merged.

PR #1141, opened 2018-08-23
FreeBSD: Fix recv problem if no data received.

azlm8t:
If using satip then we would frequently fail to read the data and then disconnect with errno 0.

So, we now make the FreeBSD socket read consistent with the Linux version and return EIO on non-error (timeout).

We're returning -errno on error, which is what the FreeBSD path used to return.

The github view below of "show changed file" cuts out just before the end of the function where 0 is returned if neither error nor timeout.


perexg:
Merged.

PR #1142, opened 2018-09-01
WebUI: fix channel tag lookup for autorec (local, not global)

ProfYaffle:
This has been bugging me for some time, and I finally got round to looking at it. It looks like the code references a global store and method (tvheadend.tags.find) when the store is actually local to this function. As such, any "Create Autorec" from the EPG will fail if a channel tag is set, because the lookup of the channel tag won't work.


perexg:
Merged.

PR #1144, opened 2018-09-07
BugFix: SATIP_SETUP_SPECINV0 and SATIP_SETUP_SPECINV1 has the same definition

mwhesotech:
SATIP_SETUP_SPECINV0 and SATIP_SETUP_SPECINV1 has the same definition.
Because of that specinv=0 is always set in the RTSP header. 
This causes problems with the AVM FritzBox 6590.


perexg:
Merged thanks.

PR #1148, opened 2018-09-17
FreeBSD: Check for fdatasync system call

azlm8t:
It's supported on FreeBSD 11.1 and later, so add check to configure.


perexg:
Merged. Thanks.

PR #1149, opened 2018-09-17
Clang remove no-incompatible-pointers Makefile option

azlm8t:
Remove the option "no-incompatible-pointer-types" for clang since gcc does warn for such types and the official build is done with gcc, making clang builds sometimes not compile with gcc.

Fixed up the couple of locations with such incompatible pointers (mainly const). Also fixed up a few places that gcc can warn about out style declaration (mainly "static" not being first and "__thread" not being before type).


perexg:
Merged. Thanks.

PR #1150, opened 2018-09-18
ui: Fix createToolbar2 issues on some recent browsers.

azlm8t:
Recent versions of Firefox had an exception saying the createToolbar2
function does not exist when it was called to create the secondary
toolbar that is used for category searching in the EPG.

I believe this was due to scoping issues, so rescoped the variables.
This worked, but although the toolbar was added it was not displayed
on Firefox.  So changed the way the toolbar is added and it now works
on Firefox and Safari.


perexg:
Merged. Thanks.

PR #1151, opened 2018-09-18
dvr: Add minyear/maxyear to autorec.

azlm8t:
My listings provider thinks every old film is worth 5/5, so we add
minyear to allow an autorec of "good films" to filter out old
films. Also add corresponding maxyear for people who think the
opposite and only want old programmes recorded.


perexg:
Merged. Thanks.

PR #1152, opened 2018-09-19
dvr: Async reschedule autorecs whenever a log entry is destroyed.

azlm8t:
Destroying a log entry can cause another autorec rule to be valid and
need scheduling.

For example, if you have two autorec rules that match the same
programme then only one autorec rule will schedule the programme
and the other autorec will see the programme is scheduled and do
nothing.

However, if you then disable the first autorec rule, then we need the
second autorec rule to re-arm the timer for that programme since it
still matches it.

This resheduling is done async after a delay. This avoids large
changes causing constant rescheduling. So, if user deletes a thousand
log files, we do not want a thousand reschedules to occur. Instead,
we dispatch a single timer after the last update has occurred.

Performance:
I've left debug log messages around the cb for timing reasons, but they 
could be removed in the future. On my box, the cb time for reschedule
is tiny, and I'd expect it to be minimal on low spec boxes, but thought
it best to leave the debug in just in case the schedule causes issues.

I had to drop an info to debug for "max recordings" since it would log
every time a reschedule occurred.

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -698,6 +698,8 @@ void dvr_autorec_init(void);
 void dvr_autorec_done(void);
 
 void dvr_autorec_update(void);
+/// Check autorec timers after a short delay.
+void dvr_autorec_async_reschedule(void);
 
 static inline int
   dvr_autorec_entry_verify(dvr_autorec_entry_t *dae, access_t *a, int readonly)
--- a/src/dvr/dvr_autorec.c
+++ b/src/dvr/dvr_autorec.c
@@ -1360,6 +1360,29 @@ dvr_autorec_update(void)
   }
 }
 
+static void
+dvr_autorec_async_reschedule_cb(void *ignored)
+{
+  tvhdebug(LS_DVR, "dvr_autorec_async_reschedule_cb - begin");
+  dvr_autorec_update();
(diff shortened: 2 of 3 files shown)


perexg:
Merged. Thanks.

azlm8t:
I've spotted a problem where it seems we can create an dvr_entry in the autorec code, but then it later gets marked as a duplicate and destroyed, which then causes us to do an async reschedule after the minute delay. It only seems to affect one of my recordings so I'll see if I can track it down properly tomorrow.

Impact is an additional "scheduling" info line once every minute.

=-=

Seems the problem may be that when a programme has been recorded and an autorec matches that programme, then we create the dvr entry regardless (since broadcast=0 on the existing entry so we bypass all the new replace logic), then we create the dvr_entry. We realize the start time has elapsed so we trigger the recording to start immediately in an async timer. That timer then realizes the dvr entry is a semantic dup of the original dvr entry so destroys it. Hence, the messages only occur during the duration of the programme when it's repeated on tv. So, at least that explains why I missed it.

I'll have to do a bit more digging. Perhaps we will have to create the entry and do a dup check earlier before the logging. I'll have to do a bit more debugging. Certainly it appears our dvr list has entries that will get cancelled once the start time is reached (e.g., in dvr/log I have one scheduled for Friday that I know is a repeat of the film that was on the previous Thursday).

PR #1154, opened 2018-09-19
FreeBSD makefile fixups

azlm8t:
Couple of minor fixups for FreeBSD. The "stat(1)" command on FreeBSD takes different arguments. It also requires libinotify for inotify support, otherwise the compile works but the link fails.


mpmc:
Thanks for your commits @azlm8t ! You're on a roll! :+1:

azlm8t:
*laugh* Thanks, yes, I had a bit of spare time for a change.

perexg:
Merged. Thanks.

PR #1155, opened 2018-09-20
ui: Add next button for epg and dvr

azlm8t:


azlm8t:
It was mentioned in the forum that dialogs could use a next/prev button. There did not seem to be enough room to add both without making the dialogs bigger, so I only added a next. If you get to the end of the list, then the button disables. This button is in the epg window and the dvr windows.

It was also mentioned that epg broadcast details dialog does not have channel name.  I think that's only a problem if user does not have channel icons, so I put the details as part of the dialog title bar to avoid having extra text.

These changes appear to work for me, even with two dialogs open at the same time. I had to re-scope some variables so things like "imdb" used the details for the current dialog, but it seems to work fine.

perexg:
Please, rebase to master.

azlm8t:
Sorry, I hadn't realized the other UI change would conflict. I've rebased the code and also added channel details to the dvr dialog titlebar.

perexg:
Merged. Thanks.

PR #1156, opened 2018-09-20
dvr: Add minseason and maxseason to autorec (#5176)

azlm8t:
This allows users to specify specific seasons to record, season has to be >= min and <= max.

This isn't quite what was requested in bug: it wanted a concat of title,subtitle,summary,description to allow users to regex seasons to record. But that method I think would be slow and a bit error prone with the OTA scrapers. So, instead I added another min/max field to autorec.

For example min=5,max=5 will record only season 5.

Setting min=5,max=0 will record everything on season 5 or
later, which is useful if you've already seen seasons 1--4.

This can then be combined with the normal fulltext regex
such as "Simpsons"+minseason=12.

Fixes: #5176

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -404,6 +404,8 @@ typedef struct dvr_autorec_entry {
   int dae_maxduration;
   int dae_minyear;
   int dae_maxyear;
+  int dae_minseason;
+  int dae_maxseason;
   uint32_t dae_retention;
   uint32_t dae_removal;
   uint32_t dae_btype;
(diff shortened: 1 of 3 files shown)


Trujulu:
Woow, what a good idea.

Even not using DVR from my side.

azlm8t:
I've spotted a problem where it seems we can create an dvr_entry in the autorec code, but then it later gets marked as a duplicate and destroyed, which then causes us to do an async reschedule after the minute delay. It only seems to affect one of my recordings so I'll see if I can track it down properly tomorrow. 

Impact is an additional "scheduling" info line once every minute.

azlm8t:
(Sorry, above comment was for wrong PR and is for the async reschedule PR).
Thanks!

PR #1157, opened 2018-09-21
webui: Add Previous button to epg and dvr broadcast popup, fix minor issues (If you s…

pablozg:
…elect the first / last row the "previous / next" buttons are actives when the popup window is opened first time)

I think the new "next" button is a great idea (thanks @azlm8t), but I think is needed a "previous" button too, because if you are reading  the broadcast details and you are going faster, you can miss a previous program of your interest, so if you don't have a previous button you must close the popup and find the broadcast you are looking for. ;-)

@perexg, if you think that is not needed, feel free to close this PR.

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -180,14 +180,12 @@ tvheadend.dvrDetails = function(grid, index) {
               handler: previousEvent,
               iconCls: 'previous',
               tooltip: _('Go to previous event'),
-              text: _("Previous"),
         }));
         buttons.push(new Ext.Button({
             id: nextButtonId,
             handler: nextEvent,
             iconCls: 'next',
             tooltip: _('Go to next event'),
-            text: _("Next"),
         }));
 
     return buttons;
@@ -202,7 +200,7 @@ tvheadend.dvrDetails = function(grid, index) {
             title: dialogTitle,
             iconCls: 'info',
             layout: 'fit',
-            width: 650,
+            width: 760,
             height: windowHeight,
             constrainHeader: true,
             buttonAlign: 'center',
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -26,6 +26,7 @@ tvheadend.dvrDetails = function(grid, index) {
     var win;
     // We need a unique DOM id in case user opens two dialogs.
     var nextButtonId = Ext.id();
+    var previousButtonId = Ext.id();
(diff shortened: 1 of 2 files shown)


azlm8t:
I merged it locally and seems good to me, but I had to increase the epg dialog width from 850 to 900 otherwise sometimes the "next" button would be partially off screen on series that have both record episode and record series buttons.

pablozg:
@perexg @azlm8t, What do you think about delete the button text and only left the icon? With this will avoid issues with lower resolutions and we can reduce a bit the popup window width.

I have added a check to disable the record button if the event is recording or scheduled.

An example:
epg

ProfYaffle:
@pablozg My personal view (I know you didn't ask :) ) is that icons should be icons without accompanying text. If you need text, the icon is probably unclear - plus, there are always the tooltips.

Removing the text from buttons in general would perhaps make the interface look less cluttered, and allow dialogs to be smaller at the same time. At the moment, the icon bar is determining the size of the window, which typically doesn't then contain that much text.

mpmc:
> @pablozg My personal view (I know you didn't ask :) ) is that icons should be icons without accompanying text. If you need text, the icon is probably unclear - plus, there are always the tooltips.
> 
> Removing the text from buttons in general would perhaps make the interface look less cluttered, and allow dialogs to be smaller at the same time. At the moment, the icon bar is determining the size of the window, which typically doesn't then contain that much text.

This is my thoughts too! :+1:

azlm8t:
Sounds good to me. I was thinking something similar last night with the epg in that it might be good to have a "next" icon on the epg right next to "filter channel". That could make the epg more usable for some people? The idea being that if you hit "next" it would populate the filter with the first channel, hit "next" again and it would go to the next channel. This would then filter the epg to just that channel and be like a "now and next epg" that some tvs have.

mpmc:
> Sounds good to me. I was thinking something similar last night with the epg in that it might be good to have a "next" icon on the epg right next to "filter channel". That could make the epg more usable for some people? The idea being that if you hit "next" it would populate the filter with the first channel, hit "next" again and it would go to the next channel. This would then filter the epg to just that channel and be like a "now and next epg" that some tvs have.

Yes! As would my comment here, http://tvheadend.org/issues/1133#note-30 =D

azlm8t:
Great minds... :-)

pablozg:
I have upload the changes. ;-)

azlm8t:
I've also added a PR for the next/prev buttons for the EPG.

PR #1162, opened 2018-09-24
epg.js - add Duration-Filter

Grabber66:
add Duration-Filter for 00:01:00 - 03:00:00


ProfYaffle:
Dare I ask why this is needed? Do you have some zero-length events you're trying to filter out, or some reaaaaaally long ones (24-hour news or shopping channels)?

When I wrote this originally, I did wonder about two separate boxes - "min..." and "max..." - or even a slider, like you see when searching web sites based on min/max price (... RAM, HDD space ...) sometimes. I wonder whether we're heading back that way, rather than adding "out-of-sequence" filters...

Grabber66:
@ProfYaffle
I used this time Filter, because in epg overview i have on the 1 Site only Sportevents and Shopping-
channels. Those are dropped out with an time Filter -> 3h. And 01min i must use because you can't use
two start-times with the same number.

ProfYaffle:
I would have thought, then, that a channel group would allow you to simply filter out those channels on your local installation when you don't want to see them - or simply disable them if you never want to see them.

Having an out-of-sequence filter for a specific use case feels odd to me otherwise; if it's really needed, I would lean towards making it fully-configurable instead. Just my thoughts, though.

xstreamon:
> I would have thought, then, that a channel group would allow you to simply filter out those channels on your local installation when you don't want to see them - or simply disable them if you never want to see them.
> 
> Having an out-of-sequence filter for a specific use case feels odd to me otherwise; if it's really needed, I would lean towards making it fully-configurable instead. Just my thoughts, though.

This would be perfect. But i can't do this. It is out of my knowlage.

perexg:
Rejected (and I think that we have a bug for this similar issue somewhere).

PR #1166, opened 2018-09-26
dvr: add infrastructure for fanart support, bump htsp version

azlm8t:
Add fanart_image to dvr entry and send it via htsp. I've a patch (for pvr.hts) that will then pass it to Kodi where it is displayed as the background image when browsing recordings. Kodi doesn't support fanart in EPG so we don't add support for that.

Added new format specifier (for post-proc) of "%U" for passing the UUID of the recording.

Added basic test program of support/tvhmeta for setting fanart/image artwork for a particular UUID using api/idnode/{save,load}. This is mostly a demo program to show how fanart could be set by the user so people could write scripts to do it. In the future, it could set other metadata from some data lookup (such as broadcast year, or full description from a tmdb lookup for people with poor OTA data).

I have not added any tmdb/tvdb/imdb lookups for fetching URLs.

The main reason to get the support in now is to try and get the Kodi patch in before the Kodi 18 release.

Tested via the support/tvhmeta program.

### DIFF
--- a/docs/property/postprocessor.md
+++ b/docs/property/postprocessor.md
@@ -26,6 +26,7 @@ Format | Description                               | Example value
 `%r`   | Number of errors during recording         |  0
 `%R`   | Number of data errors during recording    |  6
 `%i`   | Streams (comma separated)                 |  H264,AC3,TELETEXT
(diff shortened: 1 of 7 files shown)


perexg:
Merged, but without the htsp version update. The change is not critical and the clients can detect the fanart using the new item name.

PR #1168, opened 2018-09-27
ui: Add fanart background to dvr details dialog.

azlm8t:
We now display fanart (if available) on the background of the dvr
dialog. This fanart image is also displayed every ten seconds where
the existing image is displayed.

Tested with safari/firefox and it seems ok, but not great.  I could not get the images scaled nicely based on the center, but hopefully a bit more css will fix that and how to scale/zoom images will become more obvious once we have a proper fanart grabber and more fanart images to play with.

The main complication is that css does not directly support making images less visible, so I had to create three div elements, and make the image in its own layer, and then make that layer transparent.

I also have to fiddle with the z-index on the (existing) image otherwise hover stops working (for enlarging the image), presumably because it's now inside a layer. And has to put the image inside a fixed-width box in case the fanart and image are different dimensions, otherwise the text would reflow.
ss
ss-st

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -24,6 +24,7 @@ tvheadend.labelFormattingParser = function(description) {
 tvheadend.dvrDetails = function(grid, index) {
     var current_index = index;
     var win;
+    var updateTimer;
     // We need a unique DOM id in case user opens two dialogs.
     var nextButtonId = Ext.id();
     var previousButtonId = Ext.id();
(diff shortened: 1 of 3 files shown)


Trujulu:
:o looks pretty good

azlm8t:
Just discovered a problem (I had not realized the css was shared between dvr and epg which means the image is too big in the epg dialog). I'll fix later today.

azlm8t:
Fixed the epg dialog image problem caused by change to the css.

PR #1169, opened 2018-09-27
python: add very basic support for tmdb artwork lookup

azlm8t:
This requires the third party library tmdb3, which (apparently) can be installed with "pip install tmdb3" if not in the package manager.

The script tvhmeta can be added as a pre-processor command for recordings with arguments of "--uuid %U --tmdb-key abcdef". 
Such as:
`/usr/local/bin/tvhmeta --uuid %U --tmdb-key abcdef`

When a recording starts, it will get the uuid, query the tvheadend server for the title+year, then do a tmdb lookup. If successful, it will then set the artwork and fanart in tvheadend, and this is viewable on latest Kodi build.

This only works with python2.7 which is a limitation of the underlying tmdb library, which I assume will get fixed upstream soon, otherwise we can rewrite that library since the api seems simple enough. The tmdb3 library uses a cache of recent lookups which is stored in /tmp.

The program also has a "--debug" option to see the data sent and received.

Keys: Other PVRs seem to ship with a system-wide key. So that could be set as the default in the code, but still allow the user to override if they want to.

The tv_meta_tmdb.py is a library and an executable, allowing lookups on command line without tvheadend interaction. The intention with "tv_meta" prefix is that in the future we could perhaps plugin other libraries for other services like tv episodes.

I've done testing with a number of films and it seems ok, though I only have English data.  The library suggests it gets locale data and does lookups accordingly.

Future: I imagine we might have to bundle tmdb3.py. We should add a hard-coded default tmdb key.

Packaging: I believe the new binaries install ok (they do for me).

Auth: I'm not sure if username/password option is working (since I don't have passwords configured). Running the code against `nc -l 1234` shows python is sending a `Authorization: Basic YWJjOmRlZg==`, so I'm assuming that's correct.

Once we get tmdb lookups working then we can think about tvdb and other services.


CvH:
>We should add a hard-coded default tmdb key.

no too good idea, better add an default one and offer the possibility to add one via the gui

the default shipped one is likely to get nuked sooner or later as someone abuses it

azlm8t:
Do you mean put a default key in the code and keep the option for the user to change it (--tmdb-key)? In which case we agree. (My use of "hard coded default" was not a good phrase). We could always store the key rot13 to prevent casual abuse :-)

I like the idea of allowing the user to set the key in the GUI (rather than mess with preproc/postproc commands).

Eventually, I'd think I'd like a tick box "lookup artwork", some way of adding different modules and priorities (user A: "I want imdb then tmdb", user B: "no, I want tmdb then imdb"), and entering keys/extra config for each.

CvH:
>We could always store the key rot13 to prevent casual abuse :-)

afaik it is rather easy to get the key from the url requests, so crypting the key at the code is rather useless :)
Everything cli based around Tvh has MUCH less users, so if you want that ppl use it include it at the gui ;)

azlm8t:
I agree, I'd like it in the UI. The user should be able to just tick a box and have it work without having to do any actual config.

I've not got basic code to do "grabber plugins". So a command line option gives a list of grabbers to use (defaulting to tmdb) and it will load the class, create an object, and use that to do the fetch; if it fails then it uses the next one in the list.

I'll do some more tidy-up today before delivering.

azlm8t:
Added the ability to find and load grabbers dynamically (based on them having the prefix tv_meta, similar to tv_grab), so new grabbers could be implemented (for example ofdb, omdb, etc.)

PR #1171, opened 2018-09-30
channel ui: Return unique list of services

azlm8t:
Previously if multiple services were mapped to same channel then
we would get duplicate entries in the csv displayed at the UI in the 
drop down channel list such as a string of 'DVB-T,DVB-T,DVB-S,DVB-S'.

Now we return a unique list of 'DVB-T,DVB-S'.

### DIFF
--- a/src/channels.c
+++ b/src/channels.c
@@ -961,12 +961,26 @@ channel_get_source ( channel_t *ch, char *dst, size_t dstlen )
 {
   const char *s;
   idnode_list_mapping_t *ilm;
-  size_t l = 0;
+  /* Unique sorted string list since many services with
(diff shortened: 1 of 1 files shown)


Trujulu:
This is strange, when there were duplicates of the same network in services, normally one worked and the other did not (to give an example).

azlm8t:
This change is related to the UI rather than switching services. If you raise it in the forum with a screenshot then it may be clearer. Ensure you tick the boxes in Configuration->Stream->pass/htsp for "restart on error" and "switch to other service". Also if you rescan the mux and go to the status tab then you should see all adapters able to scan for services.

PR #1180, opened 2018-10-04
fanart: Add tvdb, update tmdb to avoid external dependencies

azlm8t:
Add a tvdb parser. This requires a "--tvdb-key XYZ" option.

Rewrite tmdb parser to no longer rely on external modules other than python-requests; update Debian control to recommend python + python-requests.

The tmdb parser now handles tv as well as movies. All code now works under Python3.

And new checkbox to dvr_config to only lookup artwork for "good broadcasts", i.e., ones with year and/or season/episode.

Added new loop in dvr to keep removing entries from "pending fanart lookup" list until one needs artwork. This could happen if user disables fanart lookup or edits fanart externally. So entries can be quickly (but lazily) purged once the timer occurs.

I think two final future improvement are: 
1. Parse Tvheadend's three character language ("eng") in to two character code ("en") so programme lookups automatically know the correct language (since tmdb/tvdb allow two character language codes).
2. Add some nicer UI. The tvhmeta has a "--list-grabber-capabilities" to return a json of modules and their required arguments, but nothing parses this yet.

Sample image from OTA broadcast below:

simp


azlm8t:
Added support in fanart to map the language tuple in the title provided by Tvheadend to the locale code needed by tvdb/tmdb (so eng become en) and pass this to the tmdb/tvdb for searches.

perexg:
Merged. Thanks.

PR #1182, opened 2018-10-07
htsp: Allow updateDvrEntry to set previously recorded flag. (htsp 34).

azlm8t:
This will allow Kodi to offer a client actions menu to mark timers as "previously recorded."

We add an additional check for a new "prevrec" field in the htsp updateDvrEntry. It takes a single integer similar to UI API of 1=>previously recorded, 0=>not, -1=>toggle, and just passes it to the existing handler used by the UI API.

My current proposed pvr.hts changes would just send [id, prevrec=1] to updateDvrEntry, so Tvheadend does an update with no changes to start/stop, etc., then updates the prevrec field.

I have the pvr.hts code working locally, but I'm not 100% sure if this change to the htsp message is the right one to use for this, or if we should add a new message such as a prevrecDvrEntry, similar to the api/dvr/entry/prerec/{toggle,set,unset}.

So, reject if you think we should do it differently.

Setting a flag in the message that then alters multiple flags internally seems slightly inconsistent with the existing API; but it is an "update" so makes sense and keeps the api small.

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -65,7 +65,7 @@
 
 static void *htsp_server, *htsp_server_2;
 
-#define HTSP_PROTO_VERSION 33
+#define HTSP_PROTO_VERSION 34
 
 #define HTSP_ASYNC_OFF  0x00
 #define HTSP_ASYNC_ON   0x01
@@ -2042,7 +2042,7 @@ htsp_method_updateDvrEntry(htsp_connection_t *htsp, htsmsg_t *in)
   time_t start, stop, start_extra, stop_extra, priority;
   const char *dvr_config_name, *title, *subtitle, *summary, *desc, *lang;
   channel_t *channel = NULL;
-  int enabled, retention, removal, playcount = -1, playposition = -1;
+  int enabled, retention, removal, playcount = -1, playposition = -1, prevrec;
 
   de = htsp_findDvrEntry(htsp, in, &out, 0);
   if (de == NULL)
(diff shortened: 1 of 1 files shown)


azlm8t:
It's mainly for autorec rules that match something you never want to record, rather than recordings you've manually scheduled. The (built-in) UI has the functionality.

So, say you have an autorec "record all Tom Cruise movies, handle duplicates by title". Then it will match Mission Impossible when it's on. But, you've already seen that film and don't want to waste a tuner recording it, so you click "disable timer". But, the next time the show is on, Tvheadend will then schedule it again (since it was not recorded, just that specific timer disabled).

Whereas the "previously recorded" button in the UI marks the programme as already recorded so then that showing and future showings match as duplicates. So, whenever Mission Impossible is aired, it matches against the previously recorded entry and does not record it. MythTV has something similar with its "never record" option.

This was mainly where I was hoping to go with the pvr.hts change since Kodi allows me to disable a specific timer, but not flag something as "never record"; and 99% of my timers are autorecs for things that might be interesting, such as "action film with 70% rating".

Leatherface75:
I can see this useful in cases with AutoRecs.
As it is now it often ends with disabling the whole AutoRec timer instead.

azlm8t:
OK, I accept sometimes adding more features may make pvr.hts less user-friendly for the majority. A bit like why the Kodi Estuary skin has X feature but not Y. Not a problem.

@pablozg You may want to try the free trial of SchedulesDirect.org to see if it helps with your stations to get better guide data.

If I get spare time next month I might see if I can add an additional "upcoming" tab to the Tvh UI to make it less grid-like and more touch friendly, perhaps have a split-panel of upcoming on the left and details on the right for the highlighted row, leaving the existing upcoming tab for more detailed/complicated changes.

PR #1186, opened 2018-10-16
FreeBSD: fix read write kevent

azlm8t:
On FreeBSD, the kevent EVFILT_READ=-1 and EVFILT_WRITE=-2 (rather than 1 and 2), so when they are ORed together you get -1, so we only end up registering READ in tvhpoll.

This means that on an async tcp connect we eventually timeout since we are not registered for WRITE.

So we now delete the special logic and use the existing separate read/write registration logic.

Also no longer return EIO for FreeBSD on tcp_socket_dead for zero bytes available to read, otherwise streaming fails to my tablet since we get no bytes to read so immediately disconnect the client, but no bytes on async socket is not same as eof.

### DIFF
--- a/src/tcp.c
+++ b/src/tcp.c
@@ -458,7 +458,7 @@ tcp_socket_dead(int fd)
   if (err < 0)
     return -errno;
   else if (err == 0)
-      return -EIO;
+      return 0;
 #else
   if (recv(fd, NULL, 0, MSG_PEEK | MSG_DONTWAIT) == 0)
     return -EIO;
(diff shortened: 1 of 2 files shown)


src/tcp.c
```diff
@@ -458,7 +458,7 @@ tcp_socket_dead(int fd)
   if (err < 0)
     return -errno;
   else if (err == 0)
-      return -EIO;
```
perexg: It looks like that 3895c923a3a959da05080831b8146c09ed143b00 should be reverted ?

perexg:
I applied kevent fix and reverted 3895c923a3a959da05080831b8146c09ed143b00 .

azlm8t:
Thanks. You're right, that earlier change should be reverted.

PR #1188, opened 2018-10-18
doozer: [4.2] add raspbian and cosmic

mpmc:
This is so that 4.2 stable is built for Raspbian and Cosmic, and is available from apt.tvheadend.org. Cosmic is currently failing, however @andoma is looking into that for us :)


perexg:
Merged. Thanks.

PR #1195, opened 2018-10-25
webui: fix maximum ticket expiration time, previously 34min, now 24hrs.

pablozg:
@perexg maybe we must set it to the maximum value of a uint32. What do you think?


Trujulu:
Why closed?

PR #1196, opened 2018-10-25
webui: deleted upper limit in ticket expiration field, now maximum value is 2047 seconds (given by INTEXTRA_RANGE function)

pablozg:


Trujulu:
That means that we can use unlimited time tickets then?

pablozg:
> That means that we can use unlimited time tickets then?

After a fast review of the code, the maximum value is given by de sec2mono(sec) function and the int64 variable, and the maximun value of the int64 is 2147483647 seconds that I think is time enought. 

@perexg can you confirm this aproach?

Trujulu:
> 
> 
> > That means that we can use unlimited time tickets then?
> 
> After a fast review of the code, the maximum value is given by de sec2mono(sec) function and the int64 variable, and the maximun value of the int64 is 2147483647 seconds that I think is time enought.
> 
> @perexg can you confirm this aproach?

It is abour 70 years !!! 👍

perexg:
I am not a fan of those changes. I added the code to allow the range 30-3600 seconds for the temporary tickets: a7eacfe7a75da4c64498c2a7be30b734de71236d . Note that each temporary ticket require some memory and if you set this value too high, the user can easily eat all memory on the server.

pablozg:
@perexg, If you see my last closed PR I have set the INTEXTRA_RANGE to a maximum of 3600 too, but to avoid change the function code I closed it and opened this PR, so for me its time enought too. Thanks.

pablozg:
It's time enought for me too. Thanks.

PR #1199, opened 2018-10-30
tags: fix exclude (from all channels)

ljalves:
Currently the exclude flag gives the user access to all channels that ARE TAGGED except the ones that match the selected tag(s).
This fix will give the user access to ALL channels (tagged or not) except the ones that match the selected tag(s).

Fixes #5288

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -98,7 +98,7 @@ access_ticket_find(const char *id)
     /* assume that newer tickets are hit more probably */
     TAILQ_FOREACH_REVERSE(at, &access_tickets, access_ticket_queue, at_link)
       if(!strcmp(at->at_id, id))
-	return at;
+        return at;
   }
   
   return NULL;
@@ -299,7 +299,9 @@ access_copy(access_t *src)
       memcpy(dst->aa_chrange, src->aa_chrange, l);
   }
   if (src->aa_chtags)
-    dst->aa_chtags  = htsmsg_copy(src->aa_chtags);
+    dst->aa_chtags = htsmsg_copy(src->aa_chtags);
+  if (src->aa_chtags_exclude)
+    dst->aa_chtags_exclude = htsmsg_copy(src->aa_chtags_exclude);
   if (src->aa_auth)
     dst->aa_auth = strdup(src->aa_auth);
   return dst;
@@ -354,6 +356,7 @@ access_destroy(access_t *a)
   htsmsg_destroy(a->aa_profiles);
   htsmsg_destroy(a->aa_dvrcfgs);
   htsmsg_destroy(a->aa_chtags);
+  htsmsg_destroy(a->aa_chtags_exclude);
   free(a);
 }
 
@@ -510,6 +513,20 @@ access_dump_a(access_t *a)
                    (long long)a->aa_chrange[first+1]);
   }
 
+  if (a->aa_chtags_exclude) {
+    first = 1;
+    HTSMSG_FOREACH(f, a->aa_chtags_exclude) {
+      channel_tag_t *ct = channel_tag_find_by_uuid(htsmsg_field_get_str(f) ?: "");
+      if (ct) {
+        tvh_strlcatf(buf, sizeof(buf), l, "%s'%s'",
+                 first ? ", exclude tags=" : ",", ct->ct_name ?: "");
+        first = 0;
+      }
+    }
+  } else {
+    tvh_strlcatf(buf, sizeof(buf), l, ", exclude tag=ANY");
+  }
+
   if (a->aa_chtags) {
     first = 1;
     HTSMSG_FOREACH(f, a->aa_chtags) {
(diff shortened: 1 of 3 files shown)


perexg:
It would be probably much better to create automatic tag which will maintain the untagged channels. This type of change will affect the rule merging scheme.

ljalves:
Sorry but I don't know what you mean with rule merging.
About your suggestion, I think that creating an automatic tag will not fix the second case reported on the opened issue (channels with 2 tags, where 1 is marked as exclude will not exclude the channel).

mpmc:
> Sorry but I don't know what you mean with rule merging.
More than likely the access entries, multiple entries of the same username with varying rights based on other factors such as IP address.

ljalves:
Got it!
I don't see a reason to cause issues on those situations... I'll try to do some testing!

ljalves:
@perexg 
You're right, I wasn't fully aware of how the access chain worked.
Let me rework this PR

ljalves:
Please check the new submit.
It builds 2 lists:
1) tag access list
2) tag exclude list

If down the chain the user adds access to some tag, it gets removed from the exclude list.

Tested with chained access and seems to work ok.

ljalves:
@perexg,
I was wondering if I should add the following behavior:

Down the access chain, if the user access has "change parameters: tags" turned on but the channel tag list is empty, then it should do the following, depending on the "exclude tag" switch:
1) exclude off: clear the tag list - user has access to all channels except previously exclude-tagged ones.
2) exclude on: clear the tag exclude list -> user has access to all channels previously tagged.

Right now, when the list is empty, it does nothing.

ljalves:
Any feedback on this?

src/access.c
```diff
@@ -622,32 +639,30 @@ access_update(access_t *a, access_entry_t *ae)
...
+        if(ct && ct->ct_name[0] != '\0') {
+          const char *ct_uuid = idnode_uuid_as_str(&ct->ct_id, ubuf);
+          if (ae->ae_chtags_exclude) {
+            lst = &a->aa_chtags_exclude;
+          } else {
+            lst = &a->aa_chtags;
+            /* adding a tag down the chain removes it from exclude list */
+            htsmsg_field_t *f;
+            if (htsmsg_is_string_in_list(a->aa_chtags_exclude, ct_uuid))
+              HTSMSG_FOREACH(f, a->aa_chtags_exclude)
```
perexg: Note to me: Add htsmsg_remove_string_from_list() fcn to htsmsc.c

src/access.c
```diff
@@ -622,32 +639,30 @@ access_update(access_t *a, access_entry_t *ae)
...
-              a->aa_chtags = htsmsg_create_list();
-            htsmsg_add_str_exclusive(a->aa_chtags, idnode_uuid_as_str(&ct->ct_id, ubuf));
+      htsmsg_t **lst;
+      LIST_FOREACH(ilm, &ae->ae_chtags, ilm_in1_link) {
+        channel_tag_t *ct = (channel_tag_t *)ilm->ilm_in2;
+        if(ct && ct->ct_name[0] != '\0') {
+          const char *ct_uuid = idnode_uuid_as_str(&ct->ct_id, ubuf);
+          if (ae->ae_chtags_exclude) {
+            lst = &a->aa_chtags_exclude;
+          } else {
```
perexg: missing the duplicate uuid check in aa_chtags (if it's excluded, it should not be allowed - overwrite previous settings)

src/channels.c
```diff
@@ -743,20 +746,21 @@ channel_access(channel_t *ch, access_t *a, int disabled)
   }
 
   /* Channel tag check */
+  if (a->aa_chtags_exclude) {
+    HTSMSG_FOREACH(f, a->aa_chtags_exclude)
+      LIST_FOREACH(ilm, &ch->ch_ctms, ilm_in2_link)
+        if (!strcmp(htsmsg_field_get_str(f) ?: "",
```
perexg: Here seems to be a clash, because the channel can be in both excluded / allowed tag lists (if the channel belongs to the multiple lists). But it's probably OK to check the excluded conditions as first.

src/access.c
ljalves: Modified to use "htsmsg_remove_string_from_list"

src/access.c
ljalves: The logic here (and on the comment below) is that the exclude list has priority over the "include" list.
So, it doesn't matter if the tag is already on the include list (no need to check for duplicates).

Only if down the access chain a user has a tag added to the "include" list, it will check for duplicates on the exclude and remove it from there (so that it gets access to that tag).

src/access.c
```diff
@@ -622,32 +639,30 @@ access_update(access_t *a, access_entry_t *ae)
...
-          }
-          if (ilm == NULL) {
-            if (a->aa_chtags == NULL)
-              a->aa_chtags = htsmsg_create_list();
-            htsmsg_add_str_exclusive(a->aa_chtags, idnode_uuid_as_str(&ct->ct_id, ubuf));
-          }
-        }
-      }
+    if (LIST_EMPTY(&ae->ae_chtags)) {
+      idnode_list_destroy(&ae->ae_chtags, ae);
```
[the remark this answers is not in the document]
ljalves: Not sure what you mean here...
If the list is empty it already does nothing - no matter if "including" or "excluding".

Not sure if I got this right, but line 643 (idnode_list_destroy) only makes sure that the webui list is clear if the list in the config is also clear, isn't it?

perexg:

ljalves:
Thank you!

PR #1200, opened 2018-10-30
 webui: access theme - color correction for EPG count info

Trujulu:
EPG events count is dark-gray so we can't see it with access theme.


Trujulu:
imagen

Just tested, this commit solves the problem.

PR #1202, opened 2018-10-30
webui: status, replace 'title' to 'user agent' (cosmetic)

Trujulu:
Probably 'Title' is not as exact meaning as it is User Agent, also the expression 'HTTP User Agent' is used in the configuration of the server, therefore, this one seems a more accurate description.


perexg:
It's not correct. The column might also contain other subscription identification. Perhaps, it would be better to split the client info and the subscription title/name.

perexg:
Fixed in b7b9ef8b8608b2c915d90d860a421221488f620e .

PR #1204, opened 2018-11-03
Rename watch descriptors to "wd" from "fd", improve traces.

joeusercz:
"fd" is used as the file descriptor for the inotify instance, so it is better to more correctly label the watch descriptors as "wd" to make it more clear.

### DIFF
--- a/src/satip/rtp.c
+++ b/src/satip/rtp.c
@@ -177,9 +177,24 @@ satip_rtp_send(satip_rtp_session_t *rtp)
       packets++;
       copy = 0;
     }
-    r = udp_multisend_send(&rtp->um, rtp->fd_rtp, packets);
-    if (r < 0)
-      return r;
+    while (1) {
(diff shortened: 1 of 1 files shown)


perexg:
I would prefer another name. 'wd' means for me write descriptor. Perhaps 'watchfd' or so..

joeusercz:
I used wd because that is what is used in inotify structure:

>            struct inotify_event {
               int      wd;       /* Watch descriptor */
               uint32_t mask;     /* Mask describing event */
               uint32_t cookie;   /* Unique cookie associating related
                                     events (for rename(2)) */
               uint32_t len;      /* Size of name field */
               char     name[];   /* Optional null-terminated name */
           };

Also, "watchfd" makes "fromfd" into something very long...
I will think of another name tomorrow...

mpmc:
How about "wed" for watch event descriptor or ewd? :)

PR #1210, opened 2018-11-11
Add ability to schedule a service to the mux scheduler.

joeusercz:


src/input/mpegts/mpegts_mux_sched.c
```diff
@@ -125,7 +163,7 @@ PROP_DOC(cron)
 const idclass_t mpegts_mux_sched_class =
```
perexg: Please, separate mux and service entries (create class hierarchy):

<pre>
  mpegts_sched_class +--- mpegts_mux_sched_class
                                  \--- mpegts_service_sched_class
</pre>

The mpegts_sched_class will have R/O entry 'Source' which will show the mux name or the service name. The source will be editable only through the 'edit' dialog. Makes sense?

joeusercz: Conceptually I understand, but I am not sure exactly how to do it. I guess I need to modify mpegts.js also. I probably will not have time to do it this week, but I will look into it.

joeusercz: Thanks for the hints.

PR #1212, opened 2018-11-13
Fix IPTV glitches

mcaptur:
Some channels used to glitch quite badly on TVH but play/recover fine on vlc(directly).. this patch makes TVH behave like vlc

### DIFF
--- a/src/descrambler/cclient.c
+++ b/src/descrambler/cclient.c
@@ -251,6 +251,7 @@ cc_remove_card(cclient_t *cc, cc_card_data_t *pcard)
     if (changed) {
       ct->cs_capid = 0xffff;
       ct->ecm_state = ECM_INIT;
+      changed = 0;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Trujulu:
Do you mean with glitches VLC cant recover properly from missing packets?

mcaptur:
sorry message was not clear
VLC opening stream directly short glitch but full recovers quickly
When using tvheadend gliches are much more severe and sometime does not even recover (when using htsp - did not test extensively with passthrough)

perexg:
How did you this analysis? The PCR clocks should be regulary updated, so the diff should not be big.

The only thing to consider is to increase the kernel socket UDP buffer settings (udp_bind() calls (setsockopt)). It may be worth to set internal kernel buffers to 512kB or so.

mcaptur:
piping via ffmpeg is actually better

EItanMao:
How do I apply these patches?

PR #1215, opened 2018-11-18
webui: epg: fix compatibility issue for FreeBSD

nomatter2k:
The EPG screen serviced by FreeBSD port of tvheadend has two abnormal control
behaviors:

1. Reset All button makes the list empty
2. Typing and then deleting any search string also makes the list empty.

This patch fixes the symptoms above.

Patch tested with the latest FreeBSD port of tvheadend-4.2.7.

### DIFF
--- a/src/webui/static/app/epg.js
+++ b/src/webui/static/app/epg.js
@@ -1166,13 +1166,11 @@ tvheadend.epg = function() {
     epgFilterTitle.on('valid', function(c) {
         var value = c.getValue();
 
-        if (value.length < 1)
-            value = null;
-
-        if (epgStore.baseParams.title !== value) {
+        if (value.length < 1) {
+            delete epgStore.baseParams.title;
+        else if (epgStore.baseParams.title !== value)
             epgStore.baseParams.title = value;
-            epgView.reset();
-        }
+        epgView.reset();
     });
 
     epgFilterFulltext.on('check', function(c, value) {


src/webui/static/app/epg.js
```diff
@@ -1166,13 +1166,11 @@ tvheadend.epg = function() {
...
-            value = null;
-
-        if (epgStore.baseParams.title !== value) {
+        if (value.length < 1) {
+            delete epgStore.baseParams.title;
+        else if (epgStore.baseParams.title !== value)
             epgStore.baseParams.title = value;
-            epgView.reset();
-        }
+        epgView.reset();
```
perexg: It would be probably better to call reset() when title changes, thus do 'delete title' in the second condition.

nomatter2k: Thanks for your comment, perexg. Then I need to modify the patch to call reset() in both 1st and 2nd condition. Calling reset() only in the 2nd condition does not refresh the list when user type some search string and then delete it to be empty.

src/webui/static/app/epg.js
```diff
@@ -1166,13 +1166,11 @@ tvheadend.epg = function() {
     epgFilterTitle.on('valid', function(c) {
         var value = c.getValue();
 
-        if (value.length < 1)
-            value = null;
-
-        if (epgStore.baseParams.title !== value) {
+        if (value.length < 1) {
```
nomatter2k: Sorry for undeleted open-brace at the end of this line.

PR #1217, opened 2018-11-19
OpenSSL Fixes

neheb:


src/main.c
```diff
@@ -1368,8 +1378,11 @@ main(int argc, char **argv)
   if(opt_fork)
     unlink(opt_pidpath);
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
perexg: What's the correct cleanup procedure for newer SSL? We need to free all allocated memory for the leak checkers.

neheb: There is none

PR #1218, opened 2018-11-20
webui: epg: fix compatibility issue for FreeBSD

nomatter2k:
The EPG screen serviced by FreeBSD port of tvheadend has two abnormal control
behaviors:

    Reset All button makes the list empty
    Typing and then deleting any search string also makes the list empty.

This patch fixes the symptoms above.

Patch tested with the latest FreeBSD port of tvheadend-4.2.7.

Changes since v1:
- call reset only if required


nomatter2k:
Thank you, perexg.

PR #1222, opened 2018-11-25
ui: show similar and alternative broadcasts (#5335, #5336)

azlm8t:
This add a similar/alternative button to DVR and EPG info dialog boxes.

Repurposed the "show duplicated" button from DVR to display this "show alternatives" dialog box. This then allows the user to select episodes to record.

Main internal changes was splitting some logic in epg.js so it could be shared with the new dialog, create new file epgevents.js to display the dialog box.

In tvh, we return the sorted lists using shared code since the dialog uses a livegrid that needs remote sorting. Also some minor const fixups and fixup core dump where NULL ptr passed through.

### DIFF
--- a/Makefile.webui
+++ b/Makefile.webui
@@ -150,6 +150,7 @@ JAVASCRIPT += $(ROOTPATH)/app/timeshift.js
 endif
 JAVASCRIPT += $(ROOTPATH)/app/chconf.js
 JAVASCRIPT += $(ROOTPATH)/app/epg.js
+JAVASCRIPT += $(ROOTPATH)/app/epgevent.js
 JAVASCRIPT += $(ROOTPATH)/app/dvr.js
 JAVASCRIPT += $(ROOTPATH)/app/epggrab.js
 JAVASCRIPT += $(ROOTPATH)/app/config.js
--- a/src/access.h
+++ b/src/access.h
@@ -261,7 +261,7 @@ access_get_theme(access_t *a);
  *
  * Return 0 if access is granted, -1 otherwise
  */
-static inline int access_verify2(access_t *a, uint32_t mask)
+static inline int access_verify2(const access_t *a, uint32_t mask)
   { return (mask & ACCESS_OR) ?
       ((a->aa_rights & mask) ? 0 : -1) :
       ((a->aa_rights & mask) == mask ? 0 : -1); }
--- a/src/api/api_epg.c
+++ b/src/api/api_epg.c
@@ -74,7 +74,7 @@ api_epg_add_channel ( htsmsg_t *m, channel_t *ch, const char *blank )
 static htsmsg_t *
 api_epg_entry ( epg_broadcast_t *eb, const char *lang, access_t *perm, const char **blank )
 {
-  const char *s;
+  const char *s, *blank2 = NULL;
   char buf[64];
   channel_t     *ch = eb->channel;
   htsmsg_t *m, *m2;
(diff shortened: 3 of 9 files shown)


azlm8t:
capture

mpmc:
😍

Trujulu:
@azlm8t Is it possible to adapt windows size also on translated UI? For English works fine, but look what happens in Spanish (text of buttons is bigger than expected and windows cut the buttons).
!(https://user-images.githubusercontent.com/7295293/49255254-c030e100-f42b-11e8-908c-6ee974251aa1.png)

Trujulu:
imagen

I just realized that it happens in more windows, it does not seem something specific about this commit.

azlm8t:
@Trujulu That's a good question.
.
I can see some examples for height, so perhaps it would work for width?
https://stackoverflow.com/questions/6602043/auto-size-ext-js-window-based-on-content-up-to-maxheight
```
listeners: {afterlayout: function() {
    var height = Ext.getBody().getViewSize().height;
    if (this.getHeight() > height) {
        this.setHeight(height);
    }
    this.center();
}}
```
Could you try that in whichever dialog fails? (with height replaced with width)?
If it works, we could put it in tvheadend.js and use on all the dialogs.

azlm8t:
@Trujulu In the future, I think it might be nice to move next/prev to a title menu bar (tbar) since the EPG "record"/"record series" means the buttons keep moving a few centimeters around. Or, perhaps move them (and the "find related" buttons) to the beginning of the bbar?

Trujulu:
> @Trujulu That's a good question.
> Could you try that in whichever dialog fails? (with height replaced with width)?
> If it works, we could put it in tvheadend.js and use on all the dialogs.

```
diff --git a/src/webui/static/app/epgevent.js b/src/webui/static/app/epgevent.js
index c79db65a6..196b56f35 100644
--- a/src/webui/static/app/epgevent.js
+++ b/src/webui/static/app/epgevent.js
@@ -154,8 +154,15 @@ function epgAlternativeShowingsDialog(eventId, alternative) {
         items: grid,
         bbar: new Ext.ux.grid.livegrid.Toolbar(
         tvheadend.PagingToolbarConf({view: epgView},_('Events'),0,0)
-        ),
-
+       ),
+       listeners: {
+       afterlayout: function() {
+           var width = Ext.getBody().getwidthSize().width;
+           if (this.getWidth() > width) {
+               this.setWidth(width);
+           }
+           this.center();
+       }}
     });

     // Handle comet updates until user closes dialog.
```

Just tried following and seems to work for epgevent dialog.

azlm8t:
@Trujulu It doesn't seem to work for me. The getwidthSize().width doesn't exist.

So I tried getWidth(), and also getSize().width.  I then also reversed your ">" to be "<" and this worked, but didn't display right.

So, I tried it on the EPG info window (not epgevent), I had set "width: 20" so I can tell that resize happens.

However, with my changes, it is resizing to 1760 (the value returned from both variants I tried). I don't know why since my screen isn't 1760 wide, but the dialog ends up looking full-width of the screen so perhaps 1760 allow space for window decoration.

Perhaps this is due to layout managers inside the epg window?

Trujulu:
Maybe I applied code Wrong. I dont know about c++, css... :/

azlm8t:
@Trujulu Not a problem, perhaps you didn't "make install"?

I'll also try and have more of a look tomorrow since I think we've almost got it working and there must be a good example online.

PR #1227, opened 2018-11-30
freebsd: Fixup header files for socket definitions. 

azlm8t:
Header files needed reordering for typedef ordering, and add missing socket headers to fix compilation on FreeBSD.

Also make thread owner debug conditional on Linux since not supported on FreeBSD/OSX.


perexg:
Merged. Thanks.

PR #1237, opened 2018-12-09
ui: Enable scrollbar for dialog info, fixes #5405

azlm8t:
When the dvr info dialog has a lot of text it was overflowing on to the buttons in the bbar.
By changing the css, the text now generates a scrollbar so no longer overflows in to the bbar. The image does still overflow, but its opacity makes it less of an issue.

Image below shows the scrollbar on the right when the window was resized to make the text flow.

scrollbar


azlm8t:
In the forum, the above patch causes a problem where the combo box can be used but the buttons are not clickable, so needs more investigation.

azlm8t:
It is now confirmed in the bug report that the above fix works.

PR #1239, opened 2018-12-17
webui, htsbuf: Content-Disposition escape chars are not correct.

### LINKED ISSUE #2086: _timeshift_flush_to_live() delivers all buffered packets unpaced, overwhelming clients
## Summary

When a client sends `subscriptionSkip` (or `subscriptionLive`) to return to the live edge during timeshift, the server calls `_timeshift_flush_to_live()` which delivers **all remaining buffered packets in a tight `while` loop** with no pacing or flow control. For long timeshift buffers (minutes to hours of content), this can:

1. **Overflow the TCP send buffer** → kernel sends RST → connection drops
2. **Overwhelm the client's demuxer/decoder pipeline** with a burst of stale data that will never be displayed
3. **Delay actual live playback** because the client must receive/process/discard the entire burst before real-time packets resume

## Affected Code

**File:** `src/timeshift/timeshift_reader.c`

In `_timeshift_flush_to_live()` (around line 140–170, depending on version):

```c
static void _timeshift_flush_to_live(timeshift_t *ts, timeshift_file_t **cur_file)
{
    streaming_message_t *sm;
    // ...
    while ((sm = _timeshift_read(ts, cur_file, 0))) {
        streaming_target_deliver2(ts->output, sm);  // ← delivers every packet, no pacing
    }
}
```

This `while` loop reads every remaining packet from the timeshift buffer and pushes it downstream immediately. There is no:
- Rate limiting or sleep between packets
- Check for client TCP backpressure
- Option to skip/discard packets that won't be needed

## Reproduction

1. Start watching a live channel (HTSP client)
2. Pause for 2+ minutes (builds timeshift buffer)
3. Send `subscriptionSkip` with a large forward offset to return to live
4. Observe: server dumps the entire remaining buffer at wire speed before resuming live delivery

With longer timeshift buffers, the problem is more severe. On a local network, the burst can be hundreds of MB delivered in seconds.

## Proposed Fix

Since the client is skipping to live, the buffered packets between the current position and the live edge are **unwanted** — the client will never display them. Instead of delivering them, they should be freed:

```c
static void _timeshift_flush_to_live(timeshift_t *ts, timeshift_file_t **cur_file)
{
    streaming_message_t *sm;
    // ...
    while ((sm = _timeshift_read(ts, cur_file, 0))) {
        streaming_msg_free(sm);  // ← discard instead of delivering
    }
}
```

This is a one-line change: replace `streaming_target_deliver2(ts->output, sm)` with `streaming_msg_free(sm)`.

After the loop completes, the reader thread naturally resumes reading from the live ringbuffer position, so the client receives fresh live packets without any burst.

### Alternative (less invasive)

If there's a reason some clients expect to receive the flushed packets (e.g., for their own re-sync logic), a compromise would be to only deliver the **last few seconds** of buffered data (enough for the client to find a keyframe and start decoding) and discard the rest.

## Client-Side Workaround

As a workaround, HTSP clients can monitor `timeshiftStatus` messages and **drop all incoming `muxpkt` packets** after sending the skip request until `timeshiftStatus.shift` confirms the stream is back at the live edge (`shift <= 1`). This is what we currently implement, but it still wastes bandwidth receiving packets that are immediately discarded.

## Environment

- TVHeadend 4.3-2634 (verified the code path is unchanged in current master)
- HTSP client on Apple TV (tvOS)
- Timeshift enabled with on-disk storage

## Related

- Bug [#2744](https://tvheadend.org/issues/2744) — crash when skipping to end of timeshift buffer (fixed in 4.1-1333, but the `_timeshift_flush_to_live` design was not changed)

oopscentral:
When attempting to download a recording with a comma Google Chrome will
fail with ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION. This is
because the comma ',' in the filename*=UTF-8'' field was not escaped.

This commit implements the defined list of non-escape characters from
RFC8187 based on htsbuf_append_and_escape_url.

The same problem occurs in issue #2086. Fixed in 2fdfe4836 "webui: fix the
attachment; filename encoding, fixes #2086" and broken again in ab9fc249a
"fix htsbuf_append_and_escape_url() - don't escape more allowed characters,
fixes #3721".

Only tested in release/4.2.

https://bugs.chromium.org/p/chromium/issues/detail?id=454165

### DIFF
--- a/src/htsbuf.c
+++ b/src/htsbuf.c
@@ -447,6 +447,61 @@ htsbuf_append_and_escape_url(htsbuf_queue_t *hq, const char *s)
   }
 }
 
+/**
+ * RFC8187 (RFC5987) HTTP Header non-ASCII field encoding
+ */
+void
+htsbuf_append_and_escape_rfc8187(htsbuf_queue_t *hq, const char *s)
+{
+  const char *c = s;
+  const char *e = s + strlen(s);
+  char C;
+  if(e == s)
+    return;
+
+  while(1) {
+    const char *esc;
+    char buf[4];
+    C = *c++;
+
+    /* RFC 8187, section 3.2.1, attr-char */
+    if((C >= '0' && C <= '9') ||
+       (C >= 'a' && C <= 'z') ||
+       (C >= 'A' && C <= 'Z') ||
+       C == '!'  ||
+       C == '#'  ||
+       C == '$'  ||
+       C == '&'  ||
+       C == '+'  ||
+       C == '-'  ||
+       C == '.'  ||
+       C == '^'  ||
+       C == '_'  ||
+       C == '`'  ||
+       C == '|'  ||
+       C == '~') {
+      esc = NULL;
+    } else {
+      static const char hexchars[16] = "0123456789ABCDEF";
+      buf[0] = '%';
+      buf[1] = hexchars[(C >> 4) & 0xf];
+      buf[2] = hexchars[C & 0xf];
+      buf[3] = 0;
+      esc = buf;
+    }
+
+    if(esc != NULL) {
+      htsbuf_append(hq, s, c - s - 1);
+      htsbuf_append_str(hq, esc);
+      s = c;
+    }
+
+    if(c == e) {
+      htsbuf_append(hq, s, c - s);
+      break;
+    }
+  }
+}
 
 /**
  *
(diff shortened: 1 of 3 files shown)


perexg:
Merged. Thanks.

PR #1245, opened 2019-01-02
CSS: More fixes

Trujulu:
2234 	.x-grid3-header-title 	Value Error : padding auto is not a padding value : auto 1px


Trujulu:
Before: 
 


After:

PR #1250, opened 2019-01-20
Changes to make tvheadend work in a container while talking to HDHomerun

kwkoo:
tvheadend relies on multicasting to communicate with the HDHomerun.

This presents problems when running tvheadend in a Docker container without enabling host networking in the container.

To solve this, I added 3 fields to the config -
1. hdhomerun_ip
2. local_ip
3. local_port

hdhomerun_ip is used during the discovery process. The IP of the HDHomerun device is passed to libhdhomerun, bypassing the multicast process.

local_ip and local_port are used for communications with the HDHomerun device. This lets you forward traffic from the Docker host to the container.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1707,6 +1707,9 @@ config_boot
   config.iptv_tpool_count = 2;
   config.date_mask = strdup("");
   config.label_formatting = 0;
+  config.hdhomerun_ip = strdup("");
+  config.local_ip = strdup("");
+  config.local_port = 0;
 
   idclass_register(&config_class);
 
@@ -1843,6 +1846,8 @@ void config_done ( void )
   free(config.picon_path);
   free(config.cors_origin);
   free(config.date_mask);
+  free(config.hdhomerun_ip);
+  free(config.local_ip);
   file_unlock(config_lock, config_lock_fd);
 }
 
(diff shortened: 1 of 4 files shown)


src/config.c
```diff
@@ -2432,14 +2441,57 @@ const idclass_t config_class = {
...
+                   "the container."),
+      .off    = offsetof(config_t, hdhomerun_ip),
+      .opts   = PO_HIDDEN | PO_EXPERT,
+      .group  = 6
+    },
+    {
+      .type   = PT_STR,
+      .id     = "local_ip",
+      .name   = N_("Local IP Address"),
+      .desc   = N_("IP address of the HDHomerun device. This is needed if you "
```
lwis: Is this description correct?

kwkoo: Thanks for catching that, @lwis . I got carried away with the copy and pasting. I have corrected the description.

oliroe:
Hi

I'm confused how these 3 settings are used, could anyone give me an example of their set up that works so I can try and replicate? TVH runs perfectly when not in a Docker but struggling with this

Thanks

kwkoo:
> I'm confused how these 3 settings are used, could anyone give me an example of their set up that works so I can try and replicate? TVH runs perfectly when not in a Docker but struggling with this

Just enter the web UI, go into Settings, and switch into Expert mode. You should see the 3 config fields there. Set the HDHomeRun IP field to the IP address of your HDHomeRun device, the local IP field to the IP address of your docker host, and the local port field to an unassigned port range on your docker host.

And don't forget to map those ports to the docker container when you start it up.

oliroe:
Appreciate your help with this but i still can't tune them. I've have tried with the Synology package and all is fine so definitely an issue with my docker setup.

I have hdhomerun_ip as 192.168.2.2 (the HDHomerun), local_ip as 192.168.2.20 (my NAS) and 9983 for the port with 9983-9986 as ports in my docker settings

From the TVH docker package it mentions the following......Host vs. Bridge
If you use IPTV, SAT>IP or HDHomeRun, you need to create the container with --net=host and remove the -p flags. This is because of a limitation in docker and multicast.

Is that still required?

Thanks

kwkoo:
> From the TVH docker package it mentions the following......Host vs. Bridge
> If you use IPTV, SAT>IP or HDHomeRun, you need to create the container with --net=host and remove the -p flags. This is because of a limitation in docker and multicast.
> 
> Is that still required?

If you create the container with --net=host, you don't need to set anything in the 3 config fields.

However, if you entered those details into the 3 config fields, you should create the container with the -p flag and leave out --net=host.

PR #1252, opened 2019-01-22
Kathrein EXIP: Add default config (SATIP)

Trujulu:
Share config w/ Triax TSS 400 (similar board and fw).

Tested with Kathrein EXIP 4124 and EXIP 418.


Trujulu:
[rootDescr.zip](https://github.com/tvheadend/tvheadend/files/2804787/rootDescr.zip)

PR #1255, opened 2019-02-03
Fix strncpy errors

olesalscheider:
We want to copy as many bytes as the destination buffer can hold.
Without this patch, tvheadend does not compile with GCC 8.

### DIFF
--- a/src/plumbing/transcoding.c
+++ b/src/plumbing/transcoding.c
@@ -1804,7 +1804,7 @@ transcoder_init_video(transcoder_t *t, streaming_start_component_t *ssc)
   char *str, *token, *saveptr, codec_list[sizeof(tp->tp_src_vcodec)];
   int codec_match=0;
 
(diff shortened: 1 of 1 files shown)


perexg:
Wrong patch. The behaviour was fine (the structure was filled with zeros at the creation time). If you like to propose the proper fix - use 'strlcpy', 'strncpy' does not guarantee that the destination string is zero terminated.

olesalscheider:
I thought `strlcpy` is not available in glibc? How about we set the last byte of the buffer to '\0' explicitly?

perexg:
We have own implementation in tvh_string.h .

olesalscheider:
Ah, ok. I have pushed an updated copy of the patch.

perexg:
Merged 8de98367c223d9e2899acd787a8e2ce55b1a1c37 . Thanks.

PR #1256, opened 2019-02-03
Fix no Signal Strength shown for FritzBox 6490/6590

Flole998:
When using FritzBox 6490/6590 as tuner we do not see a signal strength as it transmits an unexpected src=1 parameter.

### DIFF
--- a/src/input/mpegts/satip/satip_frontend.c
+++ b/src/input/mpegts/satip/satip_frontend.c
@@ -1047,7 +1047,8 @@ satip_frontend_decode_rtcp( satip_frontend_t *lfe, const char *name,
             mmi->tii_stats.snr = 12 * 0xffff / 15;
           }
           goto ok;          
-        } else if (strncmp(s, "ver=1.0;", 8) == 0) {
+        } else if (strncmp(s, "ver=1.0;", 8) == 0 ||
+                   strncmp(s, "ver=1.2;", 8) == 0) {
           if ((s = strstr(s + 8, ";tuner=")) == NULL)
             goto fail;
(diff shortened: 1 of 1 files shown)


perexg:
Could you show me the status string? Thanks.

Flole998:
Sure:  

````Status string: 'ver=1.2;src=1;tuner=1,172,1,0,346.00,8,dvbc,256qam,6900,,,,1;pids=0,18,104,6640,6641,6642,6644,1,16,17,6269'````

Flole998:
Please note that quality is inverted on the Fritzbox, I have already contacted the manufacturer about that and therefor this is not honored in this patch.

Trujulu:
What do you mean with inverted?

89% is 98%
Or 89% is 11%

??

Flole998:
No, 0 is lowest BER and 15 is highest, the standard wants it to be the exact opposite (0 is highest BER and 15 is lowest, see Page 38 of http://www.satip.info/sites/satip/files/resource/satip_specification_version_1_2_2.pdf).

perexg:
Rewritten 5caf8b8a445797a176376c9b28ce9f12cd28cf46 .

9000h:
btw for ver=1.2 the src=1 is a bug and should be fixed by AVM and they should report the signal quality which is 0 since the beginning.
but we may expect to much as they have no SAT>IP certification

Flole998:
The signal quality is the BER in this case, it's just inverted, it should be 15 but they are sending 0 instead. With the highest BER you will be getting 15. I reported this already.

PR #1260, opened 2019-02-14
Allow TVCT on ATSC-C muxes

k2fc:


src/input/mpegts/dvb_psi.c
```diff
@@ -2287,6 +2287,9 @@ psi_tables_atsc_c ( mpegts_mux_t *mm )
   mpegts_table_add(mm, DVB_VCT_C_BASE, DVB_VCT_MASK, atsc_vct_callback,
                    NULL, "vct", LS_TBL_ATSC, MT_QUICKREQ | MT_CRC | MT_RECORD,
                    DVB_VCT_PID, MPS_WEIGHT_VCT);
+    mpegts_table_add(mm, DVB_VCT_T_BASE, DVB_VCT_MASK, atsc_vct_callback,
```
chewitt, line 2290: I'll let someone else comment on whether the change can be accepted, but the indentation here is incorrect.

DeltaMikeCharlie: This PR has been outstanding for almost 7 years.  It is hoped that the following research will enable a decision to be made regarding either merging or closing this PR.

'TVCT' stands for 'Terrestrial Virtual Channel Table'.  It is part of the ATSC standard and according to Wikipedia it: "Defines each virtual channel and enables EITs to be associated with the channel."

The EIT is the 'Event Information Table', or in other words, the EPG.

The PR seems to have been raised as a result of the following forum discussion:

https://tvheadend.org/d/6466-tvct-on-an-atsc-c-system

The purpose of this PR is to allow TVCT data to be processed when received from ATSC-C broadcasts.

TVCT seems to be specifically for ATSC-T broadcasts with CVCT being used for ATSC-C broadcasts.  However, according to the forum discussion, some ATSC-C systems use TVCT when rebroadcasting services originally intended for ATSC-T delivery.

The result of this is that some ATSC services lacking channel numbers and EPG.

One user participating in the forum discussion reports that applying this PR has resolved their issue.

The PR duplicates the block of code that appears to setup a callback for a specific table ID, DVB_VCT_T_BASE (0xC8) enabling a callback for DVB_VCT_C_BASE (0xC9).

https://github.com/tvheadend/tvheadend/blob/67f98d620c03402b89f07cd0d96e843024a8ab86/src/input/mpegts/dvb.h#L91C1-L93C43

```
static void
psi_tables_atsc_c ( mpegts_mux_t *mm )
{
  mpegts_table_add(mm, DVB_VCT_C_BASE, DVB_VCT_MASK, atsc_vct_callback,
                   NULL, "vct", LS_TBL_ATSC, MT_QUICKREQ | MT_CRC | MT_RECORD,
                   DVB_VCT_PID, MPS_WEIGHT_VCT);
<NEW_CODE>
    mpegts_table_add(mm, DVB_VCT_T_BASE, DVB_VCT_MASK, atsc_vct_callback,
                   NULL, "vct", LS_TBL_ATSC, MT_QUICKREQ | MT_CRC | MT_RECORD,
                   DVB_VCT_PID, MPS_WEIGHT_VCT);
</NEW_CODE>
  mpegts_table_add(mm, DVB_ATSC_STT_BASE, DVB_ATSC_STT_MASK, atsc_stt_callback,
                   NULL, "stt", LS_TBL_ATSC, MT_QUICKREQ | MT_CRC | MT_RECORD,
                   DVB_ATSC_STT_PID, MPS_WEIGHT_STT);
}
```

Function `atsc_vct_callback` validates the table ID (0xc8 or 0xc9) but does not appear to treat the two tables differently.

Referring to the standard document ATSC A/65:2013 (7 August 2013):

Section 6.3.2 (Cable Virtual Channel Table) states: "The Cable Virtual Channel Table may be present in a terrestrial broadcast multiplex when a broadcaster has coordinated consistent channel labeling/numbering with all local cable operators carrying that multiplex, and different channel labeling and/or numbering between cable and terrestrial broadcast is desired. When both CVCT and TVCT are present in the multiplex, receiving devices are expected to use the TVCT to navigate services received via terrestrial broadcast and the CVCT to navigate services received via cable."  Section 6.3.1 (Terrestrial Virtual Channel Table), however, makes no reciprocal allowance regarding CVCT vs TVCT.

Section G.5.1 (Terrestrial Virtual Channel Maps on Cable) of the standard also states: "Although the TS on such 8-VSB may not strictly conform to SCTE standards for digital cable, cable-ready receivers should nonetheless be designed to handle the case where a Terrestrial VCT is found where a Cable VCT is expected."

Section 5.2 (Requirements for Cable), 'Requirement 6' states: "The required tables for a cable system are: the STT, either the CVCT or the TVCT, and the MGT."

It would appear that this PR is suitable for merging because:

- This feature seems to be technically non-standard but encouraged in the standards document as best practise.
- The TVH `atsc_vct_callback` function already handles the TVCT and CVCT transparently.
- The indentation issue, although aesthetically distracting, has no impact on functionality.

Flole998: Reformatting will happen soon-ish, so the formatting mistake will be auto-fixed then.

PR #1261, opened 2019-02-15
Triax & Kathrein: Increase pid length, issue #5544

Trujulu:


perexg:
Merged, but set the limit to 255 (should be fine now). 7ff49818e6ac5d0c46995f60f33248e1b2e172b0

PR #1263, opened 2019-02-25
Add option --jsondump to dump a configuration file/directory as json

olivluca:
With the new binary and gzip compressed configuration file format, it's impossible to easily view (and eventually modify) the content of configuration files.
The need to modify directly the files is because there are some fields that cannot be edited from the user interface, causing problems for some channels, e.g.

[Services cannot be mapped for channels that do not broadcast regularly](https://tvheadend.org/issues/5043)

[Biss at 27.5W](https://tvheadend.org/boards/13/topics/13421)

Once the file has been edited, there's no need to convert it to the binary gzipped format since tvheadend can read it back as json.


perexg:
No, no and no... You should not touch the config files without the tvheadend assistance (proper locking). Use the live json API in your scripts:

https://github.com/tvheadend/tvheadend/tree/master/lib/api/python#examples

olivluca:
Thank you for the link, before patching I searched for "tvheadend json api" and "tvheadend rest api" and it didn't give me that result.
But I cannot make work the export command: I got some uuids with exportcls, but any attempt to perform an export gives an "Unknown data" reply.

PR #1264, opened 2019-02-26
Change expected result type from export

olivluca:
It appears that now the result type for "export" is a list and not a dictionary

### DIFF
--- a/lib/api/python/tvh-json.py
+++ b/lib/api/python/tvh-json.py
@@ -124,7 +124,7 @@ def do_get(*args):
 def do_export(*args):
     if len(args) < 1: error(1, 'get [uuid]')
     body = do_get0('raw/export', {'uuid':args[0]})
-    if type(body) != type({}):
+    if type(body) != type([]):
         error(11, 'Unknown data')
     if 'entries' in body:
         body = body['entries']


perexg:
For which path? The result should be a dictionary.

olivluca:
I tried several uuids that I got with exportcls, the original version (expecting {}) was always failing. The result is an array.
e.g.

```
TVH_USER=admin TVH_PASS=admin python tvh-json.py exportcls mpegts_service | grep uuid | awk '{ print $2 }' | sed -e 's/"//g' -e 's/,//'  > uidslist
for i in $(cat uidslist) ; do echo $i ; TVH_USER=admin TVH_PASS=admin python tvh-json.py export "$i" ; done

```
fails with the original version, works with my modification. Maybe the result should be treated like the exportcls case (which accepts both an array and a dictionary)?

olivluca:
I now tried to export every available uuid (see below how I got them) and none of them failed, so the result is always a list (at least with my setup)

```
TVH_USER=admin TVH_PASS=admin python tvh-json.py classes | grep ':' | awk '{ print $1 }' | awk -F'"' '{ print $2 }' > classes
for i in $(cat classes) ; do echo $i >&2 ; TVH_USER=admin TVH_PASS=admin python tvh-json.py  exportcls $i ; done | grep uuid |  awk '{ print $2 }' | awk -F'"' '{ print $2 }' > uuids
for i in $(cat uuids) ; do TVH_USER=admin TVH_PASS=admin python tvh-json.py export "$i" > /dev/null ; done
```
(though several exportcls in the second step gave me the "Unknown data" error, but I suppose it's because I have no elements defined for those classes)
(there was one failure, but that's because the "grep uuid" also got the device_uuid from satip_client)

perexg:
Fixed in 0690af983e540a453e6685041e47b82bb5fa8bbd, thanks.

PR #1267, opened 2019-03-09
Added patch to HDHomerun library to allow cross-compilation

virtualdj:
Allows to static build HDHomerun library when cross-compiling for other platforms (such as armel), see feature #5130.

The HDHomerun [Makefile ignores the `CC` variable](https://github.com/Silicondust/libhdhomerun/blob/9719cbc7a562dfc3ae638ba1f3769f9451db5389/Makefile#L14-L15) set by the TVHeadend makefile after configuration and this builds the `libhdhomerun.a` library with the host gcc (x86_64 for example). This gives errors when linking at the end:
```
Markdown: docs/wizard/network.md
Markdown: docs/wizard/status.md
CC              src/docs.o
CC              build.o
CC              timestamp.o
CC              tvheadend
/root/tvheadend/tvheadend/build.linux/hdhomerun/libhdhomerun/libhdhomerun.a: error adding symbols: File format not recognized
collect2: error: ld returned 1 exit status
Makefile:699: recipe for target '/root/tvheadend/tvheadend/build.linux/tvheadend' failed
make: *** [/root/tvheadend/tvheadend/build.linux/tvheadend] Error 1
```

### DIFF
--- a/Makefile.hdhomerun
+++ b/Makefile.hdhomerun
@@ -48,6 +48,7 @@ LIB_FILES      := \
 			$(LIB_ROOT)/$(LIBHDHR)/.tvh_build \
 			$(LIB_ROOT)/libhdhomerun/libhdhomerun.a \
 			$(LIB_ROOT)/libhdhomerun/*.h
+LIB_DIFFS      := hdhomerun.diff
 
 include $(DIR)/Makefile.static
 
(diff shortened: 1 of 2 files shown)


virtualdj:
@perexg It would be great to have a similar fix for cross-compiling with **ffmpeg-static**, because it has the same issue (it doesn't use the correct `CC`).

PR #1273, opened 2019-05-01
bouquet: Allow merging of services across network bouquet, fixes #5617

azlm8t:
In a mixed network environment (such as DVB-T and DVB-S) it is common
to receive many of the same channels. However, with "auto map to
channels" on a bouquet, you could not merge the channels across the
network bouquets.

Now we create a separate "Tvheadend Network" in the bouquets and allow
"auto map to channels" on it. Any other bouquets that are enabled but
do _not_ have "auto map to channels" enabled will be added to the
enabled "Tvheadend Network".

So, if "DVB-T Network" and "DVB-S Network" in the bouquet tab are both
enabled _and_ their tickbox for "auto map to channels" is deselected,
then their services will be added to the "Tvheadend Network". That can
then be enabled with "merge channels" to produce the merged channels.

We call it "Tvheadend Network" instead of "Tvheaded Virtual Bouquet"
to be consistent with the other auto-generated names.

The new bouquet is lazily updated to ensure the CPU overhead is
minimized. This means that when other bouquets are scanned, we wait
until the system is quiet before updating the new bouquet with
the new channels. We do a full scan for the bouquet to ensure all
the counts are correctly updated in the bouquet tab.

Fixes: #5617

### DIFF
--- a/src/bouquet.c
+++ b/src/bouquet.c
@@ -34,6 +34,8 @@ typedef struct bouquet_download {
 
 bouquet_tree_t bouquets;
 
+static int bouquet_init_completed = 0;
+
 static void bouquet_remove_service(bouquet_t *bq, service_t *s, int delconf);
 static void bouquet_download_trigger(bouquet_t *bq);
 static void bouquet_download_stop(void *aux);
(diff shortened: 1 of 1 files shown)


tenfire:
I have been testing this merge request for a couple of weeks now, and it seems to work very well.
I lay my vote (if I had one) for this to be merged.

j0nnymoe:
Any news on when this will be merged or what needs to be done? Been looking for this feature for a while!
Thanks

PR #1275, opened 2019-05-14
xmltv: limit epg details for low memory devices, fixes #5630

azlm8t:
Requested in the forums.  https://tvheadend.org/boards/12/topics/37170

Some TVs have limited xmltv support, have very limited memory and don't support description or other large text fields. So, add an enum to the user to choose a basic format to reduce memory usage on clients. This will strip many large text fields.

I also changed the xmltv to never output 'lang="xyz"' fields if there is only one language. This reduces the file size.

I also added the "basic fields" support to htsp. However I'm happy to re-deliver without that commit if it's not that useful.

Many of the changes are just whitespace or splitting the basic/normal output in to a separate function.

I don't use the features, but Saen has been testing the xmltv for a couple of weeks.

### DIFF
--- a/docs/property/htsp_output_format.md
+++ b/docs/property/htsp_output_format.md
@@ -1,0 +1,8 @@
+:
+
+Option                           | Description
+---------------------------------|------------
+**All**                          | Include all information.
+**Basic**                        | Limited information for low memory devices.
+
+This setting can be overridden on a per-user basis, see [Access Entries](class/access).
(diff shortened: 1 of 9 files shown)


Saentist:
azim8t there is some missing in code it allow lines with series numbers

```<episode-num system="xmltv-ns"> . 1 .   </episode-num></programme>```

azlm8t:
I think the episode-num is correct in the xmltv output.

If your guide provides season or episode then it will still be passed through even with "basic". This should help other xmltv clients to do lookup (even though TV's IPTV currently does not use it).

PR #1277, opened 2019-05-31
bugfix for autorecs duplicate episode number detection in autorecs

HvB:
"record if different episode number" doesn't work as expected in the master branch:  https://tvheadend.org/issues/5632

With this modification, the season will be taken into account, as it was before, in order to determine if two episodes have different episode number or not.


HvB:
Not sure what's better to use between `epg_episode_number_cmp` function or `epg_episode_number_cmpfull `

PR #1278, opened 2019-06-08
Fix bogus ONID and TSID remapping

premultiply:
ONID and TSID values of 65535 (0xFFFF) are used and valid values.
This fix allows the proper reception of streams with this value.


perexg:
Your patch is not correct. Ok, I though that we need 32-bit tsid/onid values, but I was lazy like you to do the proper change. Fixed in bf7532d2c8548ae2b1519a014d619547a81508c5 and yours modified dcc50db45b322da22241c01807643160c16ccfc2.

premultiply:
Thanks! ;-)

PR #1280, opened 2019-06-18
fix #5661 Compilation Error with latest Master

caveman99:
don't assign const char to char pointer


perexg:
It would be better to use only capmt_peek_str(), it always returns a non-NULL string. Fixed in 771dfd6bea7bd4035ed991eccbe735dc00d3f800 .

caveman99:
Thanks for the quick fix. I was under the impression the 'unknown' was kind of important ;-)

PR #1281, opened 2019-06-26
Fix compilation with libhdhomerun 20190621

mamarley:
They renamed the symbol that was used to determine whether the
"hdhomerun_discover_find_devices_custom" needed to be aliased,
causing an FTBFS.  Instead, recognize both the old and new symbols.


CvH:
manually backported to 4.2 and it works there too 👍

PR #1288, opened 2019-07-22
Fix detect AC3+ audio track, fixes#5684

SqTER-PL:


src/input/mpegts/dvb_psi_pmt.c
```diff
@@ -359,6 +359,10 @@ dvb_psi_parse_pmt
     case 0x24:
       hts_stream_type = SCT_HEVC;
       break;
+          
+    case 0x87:
```
Trujulu: What about putting it after 0x81 to follow order?

Trujulu:
P.S.: Does we need add 0x87 here too? Or is not needed? https://github.com/tvheadend/tvheadend/blob/771dfd6bea7bd4035ed991eccbe735dc00d3f800/src/descrambler/capmt.c#L2093

src/input/mpegts/dvb_psi_pmt.c
SqTER-PL: You're right.
It can be after 0x81

SqTER-PL: I updated this file.
Now it's OK.

SqTER-PL:
I corrected the file "capmt.c" according to your suggestion.
AC3+(EAC3) works without this change but the source code must be correct.
This will prevent an unexpected failure.

Look at the code now. I think it will be OK.

Trujulu:
> I corrected the file "capmt.c" according to your suggestion.
> AC3+(EAC3) works without this change but the source code must be correct.
> This will prevent an unexpected failure.
> 
> Look at the code now. I think it will be OK.

I am not 100% sure of the capmt changes. They may be necessary, maybe not.

SqTER-PL:
There was a mistake in the source code.
AC3 has estype = 0x81, and EAC3 estype = 0x87 (checked in the standard documentation).
This is certainly bad, but maybe not bother.
https://github.com/tvheadend/tvheadend/blob/771dfd6bea7bd4035ed991eccbe735dc00d3f800/src/descrambler/capmt.c#L2093

SqTER-PL:
I will check the last change and let you know.

SqTER-PL:
Check other estype. In all entries, the values for a given type are the same. It certainly was a mistake that needs to be corrected.

I have just checked and all changes work well.
Tvheadent now correctly detects EAC3 audio.
I searched the entire source code and found nothing to fix.

Flole998:
Can you squash this into one commit please? Other than that: looks good to me

PR #1293, opened 2019-07-26
SATIP Kathrein & Triax: Force rolloff on DVB-S2, fixes #5517

Trujulu:
Set permanent 0.35 rolloff for Kathrein and Triax devices, to avoid failed tunning on DVB-S2 frequencies.


perexg:
It looks like that this implementation might be more universal - create a list 'force rolloff' with values 'auto', '0.25', '0.35' etc.. Also the description should be 'force rolloff' not 'avoid rolloff'.

PR #1294, opened 2019-08-06
Fixed bad quality for vaapi transcoding h264 and hevc with bitrate #5607

ovaron76github:
After this change picture quality is very good from the first second on with av-lib and vaapi transcoding (h.264 / hevc). 

See Tickets:

https://tvheadend.org/issues/5607
https://tvheadend.org/issues/5701


ovaron76github:
No reason for this. You can skip this line. For me it's more readable if the last return of a funktion is a little bit separated from the other code.

ljalves:
This PR should be merged.
Just tested this and it does fix the bad quality on the first minutes of a transcoded stream.

Thanks @ovaron76github

PR #1295, opened 2019-08-06
Corrected logic related to limit connections and connection types

niemert:
Sort issues related to user limits and dvr. 
Before fix user with "connection limit type" set to ALL and "limmit connections" set to 1, was not able to record. 
With fixes in place, limits are working for dvr and streaming as it's supposed to

This fixes Bug #5692
https://tvheadend.org/issues/5692

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -570,6 +570,7 @@ access_update(access_t *a, access_entry_t *ae)
     switch (ae->ae_conn_limit_type) {
     case ACCESS_CONN_LIMIT_TYPE_ALL:
       a->aa_conn_limit = ae->ae_conn_limit;
+      break;
     case ACCESS_CONN_LIMIT_TYPE_STREAMING:
       a->aa_conn_limit_streaming = ae->ae_conn_limit;
       break;
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -111,8 +111,8 @@ dvr_rec_subscribe(dvr_entry_t *de)
     rec_count = dvr_usage_count(aa);
     net_count = aa->aa_conn_limit ? tcp_connection_count(aa) : 0;
     /* the rule is: allow if one condition is OK */
-    c1 = aa->aa_conn_limit ? rec_count + net_count >= aa->aa_conn_limit : -1;
-    c2 = aa->aa_conn_limit_dvr ? rec_count >= aa->aa_conn_limit_dvr : -1;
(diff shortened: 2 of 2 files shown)


ljalves:
This logic seems wrong.
If you have connection limit = 1 then the limit of (stream + dvr) = 1
You are assuming that connection limit = 1 means that you can stream and record, but that mean 2 connections not 1.

ljalves:
Also with this change, if you set connection limit = dvr = 1, user will be able to start 2 recordings which is wrong.

niemert:
Hi

The difference is that when you look at connections for dvr, the connection for dvr is already started and is set to 1, therefor the difference with +1.
Today with limit all set to 1 no recording is possible at all, without watching any steam.

You can try it out yourself and see, it's an issue


On 19 Oct 2019 11:32, Luis Alves <notifications@github.com> wrote:

Also with this change, if you set connection limit = dvr = 1, user will be able to start 2 recordings which is wrong.

ljalves:
Yes, I agree with you that there is a bug but this fix doesn't seem to be correct.
I also reported a similar issue in the past (maybe related) where the http webui connection was also counting as a connection for the limits.

I would even change the config options:
- remove the limit type (the combo box to select all, dvr, stream)
- have 2 entries: one for dvr limit another for streaming limit

ljalves:
On a second look, the bug seems to be the **dvr_usage_count** function already accounting the recording that is about to start.
If this is true, then the fix should as simple as:

```
-rec_count = dvr_usage_count(aa);
+rec_count = dvr_usage_count(aa) - 1;
```

Flole998:
@ljalves I agree with the "original fix" in here, we are starting the recording, then we check if we are above the limit (and not above or equal the limit) and if we are we exit again. The rec_count should be dvr_usage_count(aa) and not dvr_usage_count(aa) -1 as we have already started the recording at that point and as its rec_count and not rec_count_minus_one we should try to do it the other way which seems cleaner to me.

ljalves:
@Flole998 - I don't think that's true - at the point where **dvr_usage_count(aa)** is called, the recording did **not** start yet - the state was already set to "recording" but it's didn't actually start yet.

In my opinion, the real bug here is the state variable **de_sched_state** that is set to **DVR_RECORDING** before the actual record starts.

Anyway, @perexg should know better...

perexg:
Fixed in fb23c42a9e398d83a76ad49d07553ddaf4c6e8d5 and 729651ce96cfd181fac127024267dbe8abedc924 .

PR #1297, opened 2019-08-23
Release/4.2 (4.3-1804~gebb096804) docker image

ralfi:
4.3-1804~gebb096804

Please fix detecting audio LATM AAC Audio Code
(example German TV tagesschau24, HDTV DVB-T2)

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -560,9 +560,6 @@
         "sudo"
       ],
       "buildcmd": [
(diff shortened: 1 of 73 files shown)


ralfi:
Sorry, my mistake. I do not know how these happens. What i mean is that the tvheadend Live-TV Player does not recognized LATM AAC Code. Everything else works fine.

PR #1299, opened 2019-09-10
man page: Correct default values for http and htsp port

thierer:
The man page incorrectly states the default values for the `--http_port` and `--htsp_port` options are 9**8**81 and 9**8**82, when in fact they are [9**9**81](https://github.com/tvheadend/tvheadend/blob/8f1de1621d78c91431238176bf4f6290870a031a/src/main.c#L822) and [9**9**82](https://github.com/tvheadend/tvheadend/blob/8f1de1621d78c91431238176bf4f6290870a031a/src/main.c#L824).

### DIFF
--- a/man/tvheadend.1
+++ b/man/tvheadend.1
@@ -82,13 +82,13 @@ Specify an interface IP address on which incoming HTTP and HTSP connections
 will be accepted. By default, connections are accepted on all interfaces.
 .TP
 \fB\-\-http_port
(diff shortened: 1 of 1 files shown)


ljalves:
This was already merged here: cb0a61e959065b321d91244d5558968a6cdcb4ad
Why is the PR still open?

perexg:
Merged in cb0a61e959065b321d91244d5558968a6cdcb4ad .

PR #1301, opened 2019-09-18
WebUI: Fix some tooltips

Trujulu:
Note: " is not shown while ' yes. Some text was truncated, so I replaced it.

image

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -2207,7 +2207,7 @@ const idclass_t passwd_entry_class = {
       .id       = "username",
       .name     = N_("Username"),
       .desc     = N_("Username of the entry (this should match a "
-                     "username from within the \"Access Entries\" tab)."),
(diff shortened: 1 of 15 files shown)


Flole998:
I dont think we shold have a period after single words like "Play" or "Fullscreen", it looks completely odd to have a "Fullscreen." tooltip.

Trujulu:
> I dont think we shold have a period after single words like "Play" or "Fullscreen", it looks completely odd to have a "Fullscreen." tooltip.

Yes, it would be better to delete those tooltips.

ljalves:
This seems more like personal preferences but honestly I don't get why the replacement of the " with ' ?
In my opinion, the correct symbol to quote words one should be " and the ' should be used on words like "don't" or "wouldn't".

I also don't like the dots in the end :)

Trujulu:
I will check it when I have time

PR #1303, opened 2019-09-24
bouquet: fix overflow when building for 32-bit system

vpeter4:
On 32-bit system hash value from service can be truncated.

For example with #SERVICE 1:0:1:835:3EA:2174:EEEE0000:0:0:0
hash value EEEE0000 become 7FFFFFFF and there is no match in function mpegts_service_find_e2().

Tested only on CoreELEC which is 32bit. Should be fine on 64bit system too but someone should check.


perexg:
Thanks. Applied to e372db0667a0072e51eb21a0b933d3b3bb8e095d .

PR #1304, opened 2019-09-25
mux grid: enable 'hide: parent disabled'

ljalves:
Lines were added commented waiting for "network enable/disable" to be implemented.
When it got implemented they were left behind.

PS: Should also make the "Services" tab rendering faster for "hide: parent disabled" since it will exit the loop as soon as it sees that the network is disabled (instead of unnecessary cycling through all the network muxes).


ljalves:
(this fixes #5732)

PR #1305, opened 2019-10-01
property texts/descriptions cleanup.

mpmc:
These have been bugging me for a while so I thought I'd start cleaning them up a little.

More to follow!


ljalves:
I don't really agree with some of the changes, like the removal of the camel case for some acronyms.

Other changes seem redundant and more of a personal "taste" (like "... use the name **defined** by the network ..." changing to "... use the name **given** by the network ...").

debian/tvheadend.service
```diff
@@ -1,6 +1,11 @@
 [Unit]
 Description=Tvheadend - a TV streaming server and DVR
-After=auditd.service syslog.target network.target local-fs.target *.mount
+# Add existing services you want started before Tvheadend, e.g, mount
+# units..
```
Martii: nit

PR #1307, opened 2019-10-09
webui_api is not expecting negative error codes.

joeusercz:


ljalves:
This PR should be closed.
@perexg properly fixed all the negative return codes at commit c67ba3ce1ba445cf2aea28315bdf97477f43198b

PR #1316, opened 2019-10-28
#4182: Add 'duplicate handling' - feature in DVR-Profiles

lomion0815:
This patch adds drop-down list to "DVR profile "configuration which is then used by default for newly created "autorec" entry. Can be altered in "Autorec" configuration.
Solves issue [#4182](https://tvheadend.org/issues/4182)


lomion0815:
Closing this PR and opening PR 1330which covers only this change on dedicated branch.

PR #1317, opened 2019-10-29
Webui: visual fixes

Trujulu:
* Sightly increase 'SAT>IP Servers' width (design collision).

* Sightly increase 'TVadapters' width (design collision).

* Fix width for spinner arrows (image was repeated)

**(Attached before/after)**

image
image

834bbaac897002aabeae6d45d05a71a6
547060ed56fdb07c5483139ed4cb4e79

image
image


Trujulu:
I have another upcoming change...

PR #1318, opened 2019-10-30
 epg: allow epg_match_event_fuzzy() to match events on different netwo…  …rks. Fixes #5494

dave-p:
On UK Freeview (DVB-T) and Freesat (DVB-S) events can have different eids for the same channel on the two networks. If a timer is scheduled on one network but the epg is gathered from the other, the eids will not match so a comparison of other parameters is needed to select the correct r/s. This patch removes the short-circuit when eids are present but different.


dave-p:
This change is not sufficient to fix the issue.

PR #1323, opened 2019-11-23
Fix #5782

Flole998:


perexg:
Thanks, but it does not explain the 100% CPU loop without this forced "connection close" transfer mode.

Flole998:
Correct, but without this even if we fix the 100% CPU Loop any Connection: upgrade response would cause tvheadend to discard the response. We can open a "invalid Connection response header causes CPU loop" bug though as that would still be an issue if the server sends "Connection: tvheadendIsAwesome" for example.

As I am not sure if the server does support keep alive I am assuming connection close, we could aswell assume keep alive probably (I am pretty sure all http2 servers do support that for normal http) but with close we are on the safe side.

PR #1325, opened 2020-01-24
Increase IPTV kill timeout

rasmusbe:
The default timeout limit is a bit short when using pipes to ffmpeg, so increasing it should make it a bit more stable.

### DIFF
--- a/src/input/mpegts/iptv/iptv_mux.c
+++ b/src/input/mpegts/iptv/iptv_mux.c
@@ -218,7 +218,7 @@ const idclass_t iptv_mux_class =
       .name     = N_("Kill timeout (pipe/secs)"),
       .off      = offsetof(iptv_mux_t, mm_iptv_kill_timeout),
       .opts     = PO_EXPERT,
-      .def.i    = 5
+      .def.i    = 15
     },
     {
       .type     = PT_STR,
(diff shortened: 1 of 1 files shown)


Flole998:
It's a configurable value, there is no need to change this in the source code for everyone....

rasmusbe:
But as I said, the _default_ value is a bit low for best results. And I don't find a way to configure it on a global level, so it means you have to configure every mux. Not really an option if you have over 9000 muxes. 

image

Flole998:
It's not a bad value, for me and many others it works fine as you are the first one to complain about it. Instead of you changing your muxes you want everyone to change theirs back do a default that was working since years?

Just select all muxes and change it for all of them. Done.

rasmusbe:
Just so I understand your argument. 
You run into the time limit much and therefore wan't it low? 
Or are you not affected by the time limit and therefore do not see it as a problem?

And just because no body else has complaint (wonder why with this welcoming response...) it should be left as is?

And just to be clear about the probI get the muxes via the IPTV Auto Network that changes from time to time, so I can change every mux now, but they will be changed back with time.

perexg:
I agree with @Flole998 . 5 seconds is a good start time. If you have other requirements, change this value for your custom tvheadend build. Even the 5 seconds is too much for the real-time streaming.

There is also some low latency settings for ffmpeg to ensure, that the buffering is minimal.

PR #1329, opened 2020-02-28
Upgrade to libhdhomerun_20200225

Trujulu:
Note that from this version Constellation must set to AUTO to tune.

photo5963265626539667613


Flole998:
We need to mention this new "limitation"/configuration issue somewhere. Or best, override it automatically if libhdhomerun is used so the input doesn't matter anymore in that case.

Trujulu:
Looks good to set AUTO as default for homerun devices. Hdhomerun support told me that was the problem that adapters don't tune.

Anyway the tuning failure occurs by updating homerun devices not tvheadend homerun firmware.

PR #1330, opened 2020-03-04
#4182: Add 'duplicate handling' - feature in DVR-Profiles

lomion0815:
This patch adds drop-down list to "DVR profile "configuration which is then used by default for newly created "autorec" entry. Can be altered in "Autorec" configuration.
Solves issue #4182
Moved PR 1316 to this PR.

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -83,6 +83,7 @@ typedef struct dvr_config {
   uint32_t dvr_removal_after_playback;
   uint32_t dvr_autorec_max_count;
   uint32_t dvr_autorec_max_sched_count;
+  uint32_t dvr_autorec_record;
   char *dvr_charset;
   char *dvr_charset_id;
   char *dvr_preproc;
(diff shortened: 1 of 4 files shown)


Flole998:
Could you please squash your commits together?

lomion0815:
"Sqash" done and rebased to latest master

Flole998:
@dave-p This is kinda related (or possibly the same) of what you did in #1488, maybe we can improve your code based on this?

dave-p:
They are similar but not quite the same. 
#1488 adds a Duplicate Handling drop-down to the autorec config screen which is used for all newly-created autorecs. However changing the config entry has no effect on existing autorecs. #1330 does the same but also adds an entry "Use config setting" to the individual autorec edit screen; if this setting is selected (the default) then changing the config entry will change all existing as well as future autorecs.
#1488 (obviously) meets my needs. I'm not sure what the "expected behaviour" should be, changing existing autorecs does seem a little dangerous.

lomion0815:
Hi, to my understanding this is the usual way TVHeadend configuration escalates changes: Changes in the recording settings are used for all (auto)recordings unless set differently in the specific autorec. Take the early start and late stop settings for instance.

Flole998:
Replaced by #1661

PR #1331, opened 2020-04-10
Consider to change default PID path to /run/tvheadend.pid

anthonytex:
Hello following [FSH3](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html) consider to change default pid path to /run/tvheadend.pid from /var/run/tvheadend.pid.

Some utilities (i.e. systemd-analyze) are now reporting *legacy path* /var/run

`systemd-analyze verify tvheadend.service`
> /usr/lib/systemd/system/tvheadend.service:13: PIDFile= references a path below legacy directory /var/run/, updating /var/run/tvheadend.pid → /run/tvheadend.pid; please update the unit file accordingly.

### DIFF
--- a/debian/tvheadend.service
+++ b/debian/tvheadend.service
@@ -9,8 +9,8 @@ After=auditd.service syslog.target network.target local-fs.target
 
 [Service]
 EnvironmentFile=/etc/default/tvheadend
-ExecStart=/usr/bin/tvheadend -f -p /var/run/tvheadend.pid $OPTIONS
-PIDFile=/var/run/tvheadend.pid
+ExecStart=/usr/bin/tvheadend -f -p /run/tvheadend.pid $OPTIONS
+PIDFile=/run/tvheadend.pid
 Type=forking
 Restart=on-failure
 RestartSec=54s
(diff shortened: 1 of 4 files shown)


Flole998:
This could break several things if I am not mistaken? Or is there some kind of support that redirects /var/run to /run so other software that is expecting the PID File in /var/run can find it there aswell?

anthonytex:
Hello, well in most systemd's based distros */var/run* is a symbolic link to */run* so it should not be a problem because actually the pid file is really in the */run* directory.
That said in older version of all system (e.g. Debian **<**7, Redhat <= 6) this may be a problem but i don't think that for newer version of TVH we should look for such old compatibility.
We are talking about *very* old versions. Note: **maybe** some particular distros are  not following FSH3 and so this change could be a problem. This PR could be left open anyway as a reminder for the future.

Flole998:
I will definitely leave it open, I need to think just a little more about what to do with it now (merge or wait).

EricV:
If it could help, I can add that I have a message at startup saying that systemd changed the location to /run by itself.

Flole998:
How about this: We do the changes in systemd now so the warning will go away but leave the old path in the binary to keep older systems happy. Please let me know what you think about that.

EricV:
Well I do not understand about old systems and new tvheadend. Unless you build everyting static, or compile bty yourself you will have to upgrade your system...

And doing what you propose is inconstent either we move all location or none.

Flole998:
Tvheadend is not just running on "normal" systems but also on Routers, NAS Devices and whatnot. Those can not be upgraded and they do not always provide /run, I want to make sure we do not cause any issues for those devices. When we only update systemd those devices will not be affected.

anthonytex:
Well IMHO I wouldn't change it in a half way because it's better to reflect program PID path with systemd' unit: you could leave this PR as a reminder, add it to a milestone if you prefer, *or* simply merge.
Thanks

anthonytex:
This is by the way the dev version of tvh so *one" should expect "breaking" changes ;)

PR #1334, opened 2020-04-17
Release/4.2

mojso:
it would be good to be able to use Stalker as a source for iptv in tvheadend


Flole998:
What exactly are you trying to do here? It looks like you are trying to merge 4.2 into master?

Leatherface75:
Probably a mistake when trying to merge to own repository but did it the other way.

PR #1335, opened 2020-05-03
Correct description of Change Parameters flag.

dave-p:
The Change Parameters flag on the Access Entries screen for a user determines whether that user's settings will override any previously-set parameters (for example from a wildcard user) - it does not affect the ability of subsequent users to override settings in turn. The exception is the 'Rights' settings where all matched users with the Change flag set are ORed together.

### DIFF
--- a/docs/property/change_parameters.md
+++ b/docs/property/change_parameters.md
@@ -3,15 +3,11 @@
 The *Change parameters* flag allows you to control which parameters
 (permission flags, all types of profiles, channel tags and ranges) are
 combined when multiple entries match a username/login. When the
-change parameter flag is enabled (checked) for a parameter that
-isn't set, the next matching entry (if any) in the sequence can set it.
-
-Username/Entry (matched in sequence)   | "Change parameters" "Rights" checked?     | Admin enabled/set?  | Can access configuration tab?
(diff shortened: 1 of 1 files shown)


Flole998:
Is there any reason why the table with the explanations was removed? Is it wrong? Or do you just think it is no longer necessary?

dave-p:
The table is incorrect.

The code implementing the 'Change Parameters' function is access_update() in access.c. If a 'change' flag is set then the corresponding setting for this user overrides any value inherited from a user earlier in the user list, otherwise it is ignored. The flag has no effect on subsequent matching users.

Ideally the UI should use 3-way check boxes (on, off, ignore) or the 'enable' box which appears if you edit multiple muxes or services, but that's one for another day.

Flole998:
Could you please rebase your commit?

dave-p:
It seems I can't rebase this PR as I no longer have the fork that it was created from...

I'll submit a new PR.

Flole998:
Ah, that's probably why I wasn't able to rebase it either ;) I thought I simply don't have permissions to do so.

dave-p:
Superseded by PR #1628

PR #1342, opened 2020-05-23
Fix building with -fno-common (default from GCC 10)

chewi:


chewi:
This deals with https://tvheadend.org/issues/5862.

sergiomb2:
works for me , thank you.

PR #1343, opened 2020-05-29
AAC ADTS and LATM correct flagging. Fixes #5887

mrwish7:
Some changes I made to flagging of AAC (ADTS) and AAC-LATM (LATM) the correct way round. They were labelled the wrong way round before. There's a related change to HTSP for this as AAC-LATM streams have to be flagged as 'AAC' for Kodi to have audio to match the new tagging.

Also a change to always stream raw AAC as 'audio/aac', seeing as LATM gets internally converted to ADTS by TVHeadend before streaming out anyway, both should raw stream as the same type. audio/aac works on internet radios and speakers.

Please test and give any feedback! I've been running this now for a couple of weeks with no issues, AAC satellite radio streams now all play properly on my Frontier Silicon chip internet radio and Sonos smart speaker.


hagaygo:
Might it fix this issue ?

https://github.com/kodi-pvr/pvr.hts/issues/431

mrwish7:
> Might it fix this issue ?
> 
> [kodi-pvr/pvr.hts#431](https://github.com/kodi-pvr/pvr.hts/issues/431)

Maybe. I don't think so, because for the Kodi HTSP side of it, it just needed the AAC workaround mapped to the AAC_LATM audio format to match them being flagged the opposite way round. It's still treating both anything flagged as AAC_LATM or AAC as 'AAC', which was what happened before.

Might be worth trying out though just in case it does trigger that to work correctly with the correct mapping!

hagaygo:
Any reason this PR is not getting merged ?

Flole998:
All PRs are staying open for some time to allow reviews. I recently included it in my build for testing and when I have some time I will look through it and test it. Have you tested it? What were your results?

hagaygo:
I can not test whether it fixes my issue (https://github.com/kodi-pvr/pvr.hts/issues/431) since my setup is on RPI using latest raspbian package (which based on master) .

Flole998:
You can always self compile and pull in whatever patches you want. I will definitely not merge an untested change, and that's why this is still open.

Saentist:
> You can always self compile and pull in whatever patches you want. I will definitely not merge an untested change, and that's why this is still open.
https://tvheadend.org/issues/5741

Flole998:
> > You can always self compile and pull in whatever patches you want. I will definitely not merge an untested change, and that's why this is still open.
> 
> https://tvheadend.org/issues/5741

Unfortunately I don't see a PR for that one yet :/ If I missed it please link directly to it, thanks!

PR #1344, opened 2020-06-05
Fix infinite loop when parsing invalid EIT CRID data

lorenz:
Fixes a logic issue which resulted in an infinite busy loop when TVH encountered certain invalid EIT data.

In case the if condition inside the loop does not evaluate to true the variable in the loop condition (len) doesn't get touched and thus gets stuck inside the loop endlessly evaluating the same if and loop conditions. This patch makes it return an error instead of indefinitely looping.


Flole998:
Thank you very much for your contribution.

PR #1345, opened 2020-06-08
xmltv: add program icon to exported xmltv. Fixes: #5685

isaksamsten:
Adds support for exporting program icons. This has been supported when importing xmltv files but is not supported in the exported xmltv.

The fix is very simple, but if needed the pull request can be updated with additional flags to enable/disable this behaviour.

### DIFF
--- a/src/webui/xmltv.c
+++ b/src/webui/xmltv.c
@@ -207,6 +207,11 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
       htsbuf_append_and_escape_xml(hq, lse->str);
       htsbuf_append_str(hq, "</desc>\n");
     }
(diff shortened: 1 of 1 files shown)


Trujulu:
Nice change, I never understood why the image was not exported

isaksamsten:
Neither do I. The argument I've seen is that it might be a struggle for low powered devices. The argument in favor is that m3u export and xmltv export now work very well in Jellyfin and Emby.

If needed, we can add a flag to enable/disable icon export.

Flole998:
Since you added it to the non-low-memory part (there is a shorter version for low memory devices) this should be fine.

Trujulu:
Thanks Flole!

PR #1346, opened 2020-06-10
Handle bad UTF-8 in xmltv (#5909)

azlm8t:
This is the patch that was tested by Ryan Li in the forum.

Let's delay adding it for a few days to see if anyone can test it with OTA data to check it doesn't break with other bad OTA data.  Ideally we'd probably split the functionality out in to a tvh_safe_utf8_strcpy and add it to all incoming data code paths (in particular incoming xmltv), but I think it's best to let it bed in for a while before doing that.

We had a string where we had a rogue byte (0x8a) which was not part of
a UTF-8 string.  This then caused some downstream parsers to abort
processing the document; other parsers ignored the bad character.

As an interim fix, we now parse the individual characters and filter
out invalid characters.  We replace such characters with a space
character (instead of a U+FFFD replacement character) since this
is typically user presentable data on a "10ft interface".  An
alternative would be to completely discard the character, but the
examples we had would then have words combined where the invalid
character used to be.

### DIFF
--- a/src/htsbuf.c
+++ b/src/htsbuf.c
@@ -353,14 +353,49 @@ htsbuf_append_and_escape_xml(htsbuf_queue_t *hq, const char *s)
 {
   const char *c = s;
   const char *e = s + strlen(s);
-  const char *esc;
+  const char *esc = 0;
   int h;
 
   if(e == s)
     return;
 
-  while(1) {
+  while(c<e) {
     h = *c++;
+
+    if (h & 0x80) {
+      // Start of UTF-8.  But we sometimes have bad UTF-8 (#5909).
+      // So validate and handle bad characters.
+
+      // Number of top bits set indicates the total number of  bytes.
+      const int num_bytes =
+        ((h & 240) == 240) ? 4 :
+        ((h & 224) == 224) ? 3 :
+        ((h & 192) == 192) ? 2 : 0;
+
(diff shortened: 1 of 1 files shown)


src/htsbuf.c
```diff
@@ -353,15 +353,50 @@ htsbuf_append_and_escape_xml(htsbuf_queue_t *hq, const char *s)
...
+        continue;
+      } else {
+        // Start of valid UTF-8.
+        if (e - c < num_bytes - 1) {
+          // Invalid sequence - too few characters left in buffer for the sequence.
+          // Append what we already have accumulated and ignore remaining characters.
+          htsbuf_append(hq, s, c - s - 1);
+          break;
+        } else {
+          // We should probably check each character in the range is also valid.
```
oldium, line 389: Your intention is to validate and handle bad characters (according to comment on top). This comment actually says that you are not validating them.

Flole998: There are 2 kinds or validation needed: Validation of UTF and validation of the characters in xml. This patch addresses the first one, so invalid UTF will be filtered. Valid UTF that isn't supposed to be in xml isn't filtered, that's what's left to do. So validation is added but not to the extend that it should be. However, this is already a benefit (as in "better than before") so it can be merged in my opinion.

PR #1347, opened 2020-06-13
iptv: new features for multicast, rtsp & rtcp (retransmission & remote time shift)

spdfrk:
Dear all,

Attached some new features I implemented for IPTV. Looking forward to your comments.

- Implement RTCP Negative Acknowledge (a.k.a. Retransmission) support for RTP streams.
    When packet loss is detected the client will send a RTCP Generic Feedback report to the server. The server can than resend these lost packets.
    Retransmitted packets are send through a second connection or as part of the main stream, both cases are supported.
    For Multicast manual setup of the RTCP server is required, for RTSP automatic setup (was already implemented for Receiver Reports) or manual override is possible.
- General clean-up of unused RTCP code and restructure to allow for easy implementation of different types of RTCP messages.
- Make RTCP Receiver Reports optional.
- RTSP start session with DESCRIBE and parse response content.
- RTSP DESCRIBE redirect support.
- Parse DESCRIBE response for AVPF support (required for Retransmission).
- Implement remote timeshift support for RTSP streams.
    This option can be enabled for a channel to pass through timeshift commands to the RTSP server, the internal timeshift buffer is then disabled.

### DIFF
--- a/src/input/mpegts/iptv/iptv_rtsp.c
+++ b/src/input/mpegts/iptv/iptv_rtsp.c
@@ -27,12 +27,14 @@
 
 typedef struct {
   http_client_t *hc;
-  udp_multirecv_t um;
   char *path;
   char *query;
   mtimer_t alive_timer;
   int play;
-  iptv_rtcp_info_t * rtcp_info;
+  time_t start_position;
+  time_t range_start;
+  time_t range_end;
+  time_t position;
 } rtsp_priv_t;
 
 /*
@@ -56,7 +58,10 @@ iptv_rtsp_alive_cb ( void *aux )
     rtsp_get_parameter(rp->hc, "position");
   else if(rp->hc->hc_rtsp_keep_alive_cmd == RTSP_CMD_OPTIONS)
     rtsp_send(rp->hc, RTSP_CMD_OPTIONS, rp->path, rp->query, NULL);
-  else
+  else if(rp->hc->hc_rtsp_keep_alive_cmd == RTSP_CMD_DESCRIBE) {
+    rtsp_send(rp->hc, RTSP_CMD_DESCRIBE, rp->path, rp->query, NULL);
+    rtsp_get_parameter(rp->hc, "position");
+  } else
     return;
   mtimer_arm_rel(&rp->alive_timer, iptv_rtsp_alive_cb, im,
                  sec2mono(MAX(1, (rp->hc->hc_rtp_timeout / 2) - 1)));
@@ -70,7 +75,9 @@ iptv_rtsp_header ( http_client_t *hc )
 {
   iptv_mux_t *im = hc->hc_aux;
   rtsp_priv_t *rp;
+  url_t url;
   int r;
+  char *p;
 
   if (im == NULL) {
     /* teardown (or teardown timeout) */
@@ -89,17 +96,67 @@ iptv_rtsp_header ( http_client_t *hc )
     return 0;
   }
 
-  if (hc->hc_code != HTTP_STATUS_OK) {
+  if (hc->hc_cmd != RTSP_CMD_DESCRIBE && hc->hc_code != HTTP_STATUS_OK) {
     tvherror(LS_IPTV, "invalid error code %d for '%s'", hc->hc_code, im->mm_iptv_url_raw);
     return 0;
   }
 
+  if (hc->hc_cmd == RTSP_CMD_DESCRIBE && hc->hc_code != HTTP_STATUS_OK && hc->hc_code != HTTP_STATUS_SEE_OTHER) {
+    tvherror(LS_IPTV, "DESCRIBE request returned an invalid error code (%d) for '%s', "
+             "fall back to GET_PARAMETER in keep alive loop.", hc->hc_code, im->mm_iptv_url_raw);
+    hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_GET_PARAMETER;
+    return 0;
+  }
+
   rp = im->im_data;
+  if(rp == NULL)
+    return 0;
 
   switch (hc->hc_cmd) {
+  case RTSP_CMD_DESCRIBE:
+    if(rp->play) {
+      // Already active, most probably a keep-alive response
+      break;
+    }
+    if (hc->hc_code == HTTP_STATUS_SEE_OTHER) {
+      if (!hc->hc_handle_location) {
+        tvherror(LS_IPTV, "received code 303 from RTSP server but redirects disabled '%s'",
+            im->mm_iptv_url_raw);
+        return -1;
+      }
+      // Redirect from RTSP server, parse new location and use that instead
+      p = http_arg_get(&hc->hc_args, "Location");
+      if (p == NULL) {
+        tvherror(LS_IPTV, "received code 303 from RTSP server but no new location given for '%s'",
+            im->mm_iptv_url_raw);
+        return -1;
+      }
+      tvhinfo(LS_IPTV, "received new location from RTSP server '%s' was '%s'", p,
+          im->mm_iptv_url_raw);
+      urlinit(&url);
+      if (urlparse(p, &url) || strncmp(url.scheme, "rtsp", 4) != 0) {
+        tvherror(LS_IPTV, "%s - invalid URL [%s]", im->mm_nicename, p);
+        return -1;
+      }
+      if(rp->path)
+        free(rp->path);
+      if(rp->query)
+        free(rp->query);
+      rp->path = strdup(url.path ? : "");
+      rp->query = strdup(url.query ? : "");
+      urlreset(&url);
+      r = rtsp_describe(hc, rp->path, rp->query);
+      if (r < 0) {
+        tvherror(LS_IPTV, "rtsp: DESCRIBE failed");
+        return -1;
+      }
+    }
+    break;
   case RTSP_CMD_SETUP:
     r = rtsp_setup_decode(hc, 0);
     if (r >= 0) {
+      if(hc->hc_rtp_timeout > 20)
+        hc->hc_rtp_timeout = 20;
       rtsp_play(hc, rp->path, rp->query);
       rp->play = 1;
     }
@@ -107,8 +164,24 @@ iptv_rtsp_header ( http_client_t *hc )
   case RTSP_CMD_PLAY:
     // Now let's set peer port for RTCP
     // Use the HTTP host for sending RTCP reports, NOT the hc_rtp_dest (which is where the stream is sent)
-    if (udp_connect(rp->rtcp_info->connection, "rtcp", hc->hc_host, hc->hc_rtcp_server_port)) {
-        tvhwarn(LS_RTSP, "Can't connect to remote, RTCP receiver reports won't be sent");
+    if (im->mm_iptv_ret_url) {
+      if (rtcp_connect(&im->im_rtcp_info, im->mm_iptv_ret_url, NULL, 0,
+          im->mm_iptv_interface, im->mm_nicename) == 0) {
+        im->im_use_retransmission = 1;
+      }
+    } else if (rtcp_connect(&im->im_rtcp_info, NULL, hc->hc_host,
+        hc->hc_rtcp_server_port, im->mm_iptv_interface, im->mm_nicename) == 0) {
+      im->im_use_retransmission = 1;
+    }
+    if (rp->start_position == 0) {
+      if (rtsp_play_decode(hc) == 0)
+        rp->position = rp->start_position = hc->hc_rtsp_stream_start;
+      else
+        rp->position = rp->start_position = time(NULL);
+    } else if (rtsp_play_decode(hc) == 0) {
+      rp->position = hc->hc_rtsp_stream_start;
+      tvhdebug(LS_IPTV, "rtsp: position update: %" PRItime_t,
+          hc->hc_rtsp_stream_start);
     }
     hc->hc_cmd = HTTP_CMD_NONE;
     tvh_mutex_lock(&global_lock);
@@ -132,14 +205,61 @@ static int
 iptv_rtsp_data
   ( http_client_t *hc, void *buf, size_t len )
 {
+  rtsp_priv_t *rp;
   iptv_mux_t *im = hc->hc_aux;
+  int r;
 
   if (im == NULL)
     return 0;
 
-  if (len > 0)
-    tvherror(LS_IPTV, "unknown data %zd received for '%s'", len, im->mm_iptv_url_raw);
+  if (hc->hc_code != HTTP_STATUS_OK)
+    return 0;
 
+  rp = (rtsp_priv_t*) im->im_data;
+
+  switch (hc->hc_cmd) {
+  case RTSP_CMD_DESCRIBE:
+    if (rp == NULL)
+      break;
+    if (rtsp_describe_decode(hc, buf, len) >= 0) {
+      if(rp->range_start == 0)
+        rp->range_start = hc->hc_rtsp_range_start;
+      rp->range_end = hc->hc_rtsp_range_end;
+      tvhdebug(LS_IPTV, "rtsp: buffer update, start: %" PRItime_t ", end: %" PRItime_t,
+          rp->range_start, rp->range_end);
+    }
+    if(rp->play) {
+      // Already active, most probably a keep-alive response
+      break;
+    }
+    r = rtsp_setup(hc, rp->path, rp->query, NULL,
+                   ntohs(IP_PORT(im->mm_iptv_connection->ip)),
+                   ntohs(IP_PORT(im->im_rtcp_info.connection->ip)));
+    if (r < 0) {
+      udp_close(im->im_rtcp_info.connection);
+      udp_close(im->mm_iptv_connection);
+      http_client_close(hc);
+      return -1;
+    }
+    udp_multirecv_init(&im->im_um1, IPTV_PKTS, IPTV_PKT_PAYLOAD);
+    udp_multirecv_init(&im->im_um2, IPTV_PKTS, IPTV_PKT_PAYLOAD);
+    sbuf_alloc_(&im->im_temp_buffer, IPTV_BUF_SIZE);
+    break;
+  case RTSP_CMD_GET_PARAMETER:
+    if (rp == NULL)
+      break;
+    // Generic position update
+    if (strncmp(buf, "position", 8) == 0) {
+      rp->position = strtoumax(buf + 10, NULL, 10);
+      tvhdebug(LS_IPTV, "rtsp: position update: %" PRItime_t, rp->position);
+    }
+    break;
+  default:
+    if (len > 0) {
+      tvherror(LS_IPTV, "unknown data %zd received for '%s':\n%s", len,
+          im->mm_iptv_url_raw, (char* )buf);
+    }
+  }
   return 0;
 }
 
@@ -175,33 +295,26 @@ iptv_rtsp_start
   hc->hc_hdr_received        = iptv_rtsp_header;
   hc->hc_data_received       = iptv_rtsp_data;
   hc->hc_handle_location     = 1;                      /* allow redirects */
-  hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_GET_PARAMETER; /* start keep alive loop with GET_PARAMETER */
+  hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_DESCRIBE;      /* start keep alive loop with DESCRIBE */
   http_client_register(hc);                            /* register to the HTTP thread */
-  r = rtsp_setup(hc, u->path, u->query, NULL,
-                 ntohs(IP_PORT(rtp->ip)),
-                 ntohs(IP_PORT(rtcp->ip)));
+  r = rtsp_describe(hc, u->path, u->query);
   if (r < 0) {
-    udp_close(rtcp);
-    udp_close(rtp);
-    http_client_close(hc);
+    tvherror(LS_IPTV, "rtsp: DESCRIBE failed");
     return SM_CODE_TUNING_FAILED;
   }
-
   rp = calloc(1, sizeof(*rp));
-  rp->rtcp_info = calloc(1, sizeof(iptv_rtcp_info_t));
-  rtcp_init(rp->rtcp_info);
-  rp->rtcp_info->connection = rtcp;
+  rtcp_init(&im->im_rtcp_info);
+  im->im_rtcp_info.connection = rtcp;
   rp->hc = hc;
-  udp_multirecv_init(&rp->um, IPTV_PKTS, IPTV_PKT_PAYLOAD);
   rp->path = strdup(u->path ?: "");
   rp->query = strdup(u->query ?: "");
 
   im->im_data = rp;
   im->mm_iptv_fd = rtp->fd;
   im->mm_iptv_connection = rtp;
-  im->mm_iptv_fd2 = rtcp->fd;
-  im->mm_iptv_connection2 = rtcp;
-
+  im->im_rtcp_info.connection_fd = rtcp->fd;
+  im->im_rtcp_info.connection = rtcp;
+  im->mm_iptv_rtp_seq = -1;
   return 0;
 }
 
@@ -227,13 +340,13 @@ iptv_rtsp_stop
     rtsp_teardown(rp->hc, rp->path, "");
   tvh_mutex_unlock(&iptv_lock);
   mtimer_disarm(&rp->alive_timer);
-  udp_multirecv_free(&rp->um);
+  udp_multirecv_free(&im->im_um1);
+  udp_multirecv_free(&im->im_um2);
   if (!play)
     http_client_close(rp->hc);
   free(rp->path);
   free(rp->query);
-  rtcp_destroy(rp->rtcp_info);
-  free(rp->rtcp_info);
+  rtcp_destroy(&im->im_rtcp_info);
   free(rp);
   tvh_mutex_lock(&iptv_lock);
 }
@@ -241,8 +354,7 @@ iptv_rtsp_stop
 static void
 iptv_rtp_header_callback ( iptv_mux_t *im, uint8_t *rtp, int len )
 {
-  rtsp_priv_t *rp = im->im_data;
-  iptv_rtcp_info_t *rtcp_info = rp->rtcp_info;
+  rtcp_t *rtcp_info = &im->im_rtcp_info;
   ssize_t hlen;
 
   /* Basic headers checks */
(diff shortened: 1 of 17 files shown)


pinkflozd:
Could "Implement remote timeshift support for RTSP streams." be implemented also for normal http/hls streams?
My provider just changes the url and adds a timestamp in the url:
Ex. Live stream url: https://xyz/live/index.m3u8
Ex Timeshift url: https://xyz/vod/1592240220,1592241900/variant/2/index.m3u8

Saentist:
> Could "Implement remote timeshift support for RTSP streams." be implemented also for http/hls streams?
> My provider just changes the url and adds a timestamp in the url (start-time,end-time):
> Ex. Live stream url: https://xyz/live/index.m3u8
> Ex Timeshift url: https://xyz/vod/1592240220,1592241900/variant/2/index.m3u8

Your links are HLS not RTSP

pinkflozd:
> > Could "Implement remote timeshift support for RTSP streams." be implemented also for http/hls streams?
> > My provider just changes the url and adds a timestamp in the url (start-time,end-time):
> > Ex. Live stream url: https://xyz/live/index.m3u8
> > Ex Timeshift url: https://xyz/vod/1592240220,1592241900/variant/2/index.m3u8
> 
> Your links are HLS not RTSP

Yes i know. That is why i was asking, if it could be implemented also for HLS streams, there are multiple providers using similar HLS streams and timeshifting with timestamps. It would be nice to have Remote timeshift also for those

Flole998:
Tvheadend doesn't even support HLS yet as far as I know

spdfrk:
Well I don't know how these media controls are implemented in HLS but think what you're asking is basically having dynamic URL's for which the format is supplier specific. Then on a timeshift reload the stream with an updated URL.
Sure that can be done but that's really outside the scope of this PR and a lot of work (also as Frole mentions even HLS support is unsure and I don't have any sources to test at this moment).

First let's see if the PR as it is right now is acceptable. After that new challenges are always welcome of course.

Saentist:
> Tvheadend doesn't even support HLS yet as far as I know

It support without problem but no timeshift supported
https://tvheadend.org/boards/4/topics/32005

Actually there is a tree diferent type of timeshift scenarios for HLS

Flole998:
Using ffmpeg as middleware is not proper support for it. You can use ffmpeg for (almost) every possible video source, that doesn't mean it's supported, you just convert it to something that is supported.

Leatherface75:
Youtube livetv streams and some of all free TV channels you can find on internet have support for timeshift.

Flole998:
Unfortunately this PR is causing segfaults (in profile.c:644). In this state it can not be merged.

spdfrk:
Hi Flore, 

Thanks for testing, I guess in your case profile_deliver was called with sm == NULL. This didn't happen to me during testing. Can you tell me wat source you where playing and how you started it? Anyway I can of course just add the check for sm != NULL. Thanks!

Flole998:
I was watching from an Amazon Echo Show over HTTP and the source was a DVB-S Card.

spdfrk:
Sorry can't test this use case but it should be fixed now.

faperea:
Hi,
I'm really interested in RTSP server timeshift feature. After a struggling a lot I finaly managed to compile this version!
In iptv_rtsp.c line 294 I changed the keep_alive_cmd from 
hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_DESCRIBE;      /* start keep alive loop with DESCRIBE */
by 
hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_GET_PARAMETER;      /* start keep alive loop with RTSP_CMD_GET_PARAMETER */
because a second DESCRIBE command triggers a "405 Method Not allowed" answer.
Because of those small things I wouldn't be a better idea to use standard SDP messages (stored in a file) instead of software-fixed options?
I was testing / reviewing the code but I couldn't find where are implemented the server timeshift functions. 
My IPTV service provides play-backwards as well as fast-forward (1x, 2x, 4x, 8x, 16x, 32x), rewind, "jump back", play, pause in server side. It requires just to send a RTSP PLAY command, using standard SDP messages. As an example, for play-backwards the SDP content (apart from URL) is:
CSeq: 18\r\n
User-Agent: CTC RTSP/1.0\r\n
Session: 3746902668
Range: clock=20200806T001926Z\r\n
Scale: 16\r\n
\r\n
With this command the stream (UDP or RTP) will play backwards from the server at 16x speed!.
The time format is yyyyMMddTHHmmssZ. It is UTC time (ending with Z) with date separated by a T from time.
How could I implement those PLAY messages? How timeshift requests are passed and where in the code? 
Thanks a lot!

faperea:
By the way, this branch works great, but with local timeshift!

spdfrk:
Hi Faperea,

Thanks for testing.

> In iptv_rtsp.c line 294 I changed the keep_alive_cmd from
> hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_DESCRIBE; /* start keep alive loop with DESCRIBE _/ by hc->hc_rtsp_keep_alive_cmd = RTSP_CMD_GET_PARAMETER; /_ start keep alive loop with RTSP_CMD_GET_PARAMETER */
> because a second DESCRIBE command triggers a "405 Method Not allowed" answer.
> Because of those small things I wouldn't be a better idea to use standard SDP messages (stored in a file) instead of software-fixed options?

https://github.com/spdfrk/tvheadend/blob/iptv_features/src/input/mpegts/iptv/iptv_rtsp.c#L92

It appears that different implementations use different keep alive commands. Some time ago we implemented this fall back in case GET_PARAMETER wasn't accepted, I probably should have added the fall back from DESCRIBE as well :). This way at least we can solve your issue in an automated way, or is the stream actually stopped in case of the 405 error?
Now the tricky part is that my server returns the current time shift buffer limits which we need to do skips. Not sure how your server communicates current buffer status.

> I was testing / reviewing the code but I couldn't find where are implemented the server timeshift functions.
> My IPTV service provides play-backwards as well as fast-forward (1x, 2x, 4x, 8x, 16x, 32x), rewind, "jump back", play, pause in server side. It requires just to send a RTSP PLAY command, using standard SDP messages. As an example, for play-backwards the SDP content (apart from URL) is:
> CSeq: 18\r\n
> User-Agent: CTC RTSP/1.0\r\n
> Session: 3746902668
> Range: clock=20200806T001926Z\r\n
> Scale: 16\r\n
> \r\n
> With this command the stream (UDP or RTP) will play backwards from the server at 16x speed!.
> The time format is yyyyMMddTHHmmssZ. It is UTC time (ending with Z) with date separated by a T from time.
> How could I implement those PLAY messages? How timeshift requests are passed and where in the code?
> Thanks a lot!

Actually this is just the RTSP header, SDP would be in the body.

The entry point for the time shift callbacks is here: https://github.com/spdfrk/tvheadend/blob/iptv_features/src/input/mpegts/iptv/iptv_rtsp.c#L435

The callback is called when a SMT_SKIP or SMT_SPEED stream message is generated by the htsp server. It can then generate the playback speed (Scale) command that you need. However it appears KODI will only use skips. 

Hope this helps!

faperea:
Hi spdfrk,

Thanks for answering. 
It seems that my IPTV server (Huawei) does not comply at all with RTSP standards, neither RFC 2326 nor RFC 7826. As an example, in RFC7826 it is mandatory to implement "npt" SDP controls, but it seems that my server doesn't, and continues using "clock" SDP controls. 
So, I had to implement my own ISO86010 time parser including text parsing because (and that IS in standards), SDP controls could be range:npt=now, range:npt=beginning or (in my case) range:clock=now- as well as range:clock=yyyymmddThhssZ-yyyymmddThhssZ. 
I wrote those functions and they works perfect, giving the same results as npt SDP commands but TVH timeshift doesn't works as expected. 
In KODI, play starts normal, but when a skip command is sent to TVH, it does weird things. When, let's say, skip back command is sent, the player counters in KODI (position/end) show the position adding time instead of moving backwards. When many times left arrow is pressed the counters in KODI are correct but then in TVH those values are mixed, and finally position is reset to zero and player stalls. I wrote a lot of debugger messages to show the position pointers to see what's going on and it seems TVH inverts the scale or perhaps some functions uses absolute values and others relative values. 
I have some questions about timeshifting implementation:
- I guess hc->hc_rtsp_range_start and hc->hc_rtsp_range_end are time vars in seconds using Unix epoch scheme. It's that true?
- If the time variables are set differently, how is the TVH time range set? It is set in absolute seconds (epoch format) or relative seconds to movie/stream lenght (for example for 2 hours of remote buffer: 0=start 7200=end)? 
- Where in the code are KODI commands received? How could I see what values are KODI sending so I could establish what is the format, absolute or relative, in seconds, in epoch seconds or hour, minutes, seconds?
- In function rtsp_timeshift_fill_status in rtsp_iptv.c code (https://github.com/spdfrk/tvheadend/blob/iptv_features/src/input/mpegts/iptv/iptv_rtsp.c#L420)
the status var are updated with current, start and end values to MPEG DTS format. I think those values are used to update ts stream, but this should not be neccesary because remote server does the job. That means, the stream should not be changed locally at all. Server will receive the time control and will start playing to the required time. It this appreciaton true?  
- I think there are three time scales: ones in play_decode as well as in skeep and scale functions, others in timeshift implementation and other in http functions (to and from KODI). Is that right? How are they related? Any of those time functions reverse order or change from absolute to relative?

Thanks in advance

faperea:
Hi spdfrk,

Worried because *clock* ISO86010 format didn't work as expected, I setup  a *npt*  ISO86010 format server with [live555](http://www.live555.com/mediaServer/) and a 2 minutes TS file to test timeshift server functions, with a m3u file including rtsp address. I tested with KODI and local timeshift works fine but server timeshift does not. When enabled, KODI timeline points to the beginning (time 0), only a timer is displayed and timeshift functions do not work at all.
The same problem than using *clock* SDP control. 
I think the problem is in start, end and current variables. Also, I think the function `streaming_target_deliver2(pd->output, tsm); `  changes TS stream and -as I pointed before - no changes should be done to the stream, because server does them! But also I think that function serves as signaling for TVH timeshift activities. Am I right? How could I test server timeshift functions?
Thanks for your time and support

spdfrk:
Hi,

I guess every provider is putting it's own "swing" on it to get it working.

Basically what I would expect from KODI is that I could send a start and stop UNIX timestamp together with current time to get timeshift working from the client (KODI) side, apparently not though :). 
So right now I assume: 
- start = 0, 
- end is the total length of the buffer 
- shift is the time relative to the moment the stream was started. 

This is working in my case, for instance I can start streaming three hours in the past and then skip forward to current time. Except when you have an EPG source connected to this channel, then KODI starts to mess up again. Since it is working I didn't investigate any further what KODI expects for these parameters.

What I did in the past to test the server side easily (tvheadend > provider) was to write a Python script and use VLC to display the stream.

spdfrk:
A small update, with this commit the start time will actually be calculated based on the remote buffer. So if you start watching at the end of the buffer (e.g. live tv) you can skip back to the beginning of the buffer. 

Now one problem remains, if you skip beyond the point when you start streaming in Kodi (e.g. skip to -10:00) Tvheadend will not stream any content from the provider to the client until it's back at 0:00.

faperea:
Hi spdfrk,

I tried again to test timeshift but always have 
`H264 @ #101 Continuity counter error (total 1)` and 
`AC3 @ #102 Continuity counter error (total 1)` and then 
`[  ERROR] tsfix: transport stream H264, DTS discontinuity. DTS = 1102665, last = 1114677`
I think this is happening because TS processors are not updated correctly. When timeshift commands are sent, it is obvious that there will be a difference between the PCR and DTS times. DTS will point to next frame, but server will rewind the stream, so it is "normal" that this difference exits, so no discontinuity errors should be reported. It this is happening, I think is because parsers or tsfix functions aren't well notified about stream changes. Do you have similar issues? How could I fix this? (It happens always with any server and with any time format)

Thanks

spdfrk:
Hi,

I see them in my logs as well but after a DTS error is detected the counter is updated and all is working fine. Are you experiencing issues / regressions due to this? 
We could suppress the errors but this is not easy to fix properly as we need to do it once for every sub-stream also you don't know when (delay after sending RTSP command) or if the DTS error will happen at all.  Or maybe we should handle this stream restart (I think it basically is?) in a different way all together, any suggestions?

faperea:
Hi spdfrk,

Yes. I'm having problems. The stream freezes until the times are equal. As an example, if I send a -10.0 secs rewind command from Kodi, I get discontinuity errors followed by tsfix messages. When 10.0 secs elapse, the stream unfreezes and continues playing. It behaves more like a 10 secs pause instead of skip back 10 seconds!
I think the problem is in TS stream timestamps.
TS players expect a continuous stream of DTS / PCR times and when a timeshift command (rewind, forward, skip) is sent, the revision of local and new remote PCR / DTS times obviously do not match. I have been reviewing the local timeshift code and there is no major problem because timeshift actions are performed on a file, that is, time is added, subtracted or skipped to PCR / DTS times stored in a local file.

I think in remote timeshift, servers can do timeshift operations in two ways:
- Add or subtract times to the PCR / DTS (as "usual")
- Stop the stream and restart at requested time
Then it could be tried in two ways:
- To suppose the server works as "usual", that means, it sum or subtract desired time to a remote file when a timeshift is received, therefore locally DTS / PCR times should be modified properly so local stream and new remote requested position are synchronized 
- Locally "stop" the stream (without TEARDOWN) and restart the playback. "Stop" means erase PCR / DTS times so new position behaves like a new play or really stop URL stream and restart it again. Like when an opening play is made

In theory, when adding or subtracting time received from timeshift command to the current PCR / DTS time (in `rtsp_timeshift_fill_status` and `streaming_target_deliver2` functions) it should not produce discontinuities but it does not work! I tried changing the values ​​with `ts_rescale_inv` and `ts_rescale` with no good results.
So, I tried theory 2: Stop the play (without TEARDOWN) but I got an APPCRASH
I think, then, a solution is to stop revision / correction of TS frames while the timeshift commands are being executed and then continue the normal operation, for skip commands. 
For speed commands, it should be multiplied the DCR / PTS times appropriately. 
In sum, I think it is necessary to add functions to manipulate the PCR / DTS times to synchronize locally after executing the timeshift command or temporarily stop the error checking / correction in the PCR / DTS times temporarily.
TVH is a bit cryptic for me, without good documentation or code comments so I don't know how to manipulate PCR / DTS timings. How can this be done?

faperea:
> Hi,
> 
> I see them in my logs as well but after a DTS error is detected the counter is updated and all is working fine. Are you experiencing issues / regressions due to this?
> We could suppress the errors but this is not easy to fix properly as we need to do it once for every sub-stream also you don't know when (delay after sending RTSP command) or if the DTS error will happen at all. Or maybe we should handle this stream restart (I think it basically is?) in a different way all together, any suggestions?

faperea:
> Hi,
> 
> I see them in my logs as well but after a DTS error is detected the counter is updated and all is working fine. Are you experiencing issues / regressions due to this?
> We could suppress the errors but this is not easy to fix properly as we need to do it once for every sub-stream also you don't know when (delay after sending RTSP command) or if the DTS error will happen at all. Or maybe we should handle this stream restart (I think it basically is?) in a different way all together, any suggestions?

Yes. I also think skip-timeshift command is a restart. So, the easiest way to fix it (at least for skip commands) is to treat the stream as a restart. I also tried sending a PAUSE command (parsing the code) and then sending the PLAY command, but discontinuity errors keep occurring. 
As a suggestion, a local counter is also required to update start and end times. It is supposed that a GET_PARAMETER command in RTSP with `position, range or rtptime` content should be answered with updated values but I couldn't get it to work. Many RTSP servers uses GET_PARAMETER for proprietary commands but mainly as a "ping" command to keep alive RTSP connection. 
So, every X seconds (1 or 2) the start as well as end could be updated with something like:
`rp->range_start = time(NULL) - LenghtOfServerbuffer`
`rp->range_end = time(NULL) /* Live */`

spdfrk:
What happens if you let it play for ~20 seconds and then skip back 10 seconds? In my case I see the DTS error but it play just fine with the correct skip. 

Skipping back 10 seconds immediately after start will also pause my playback till its back to zero. Also here the same DTS error.  I expect it has something to do with TvHeadend not expecting the playback to go before 0:00 as with normal TS this isn't possible.

So basically this proves to me the DTS error isn't the problem. If the first test doesn't work for you then we have an additional problem.

faperea:
Hi,
As you suggested I made a few attempts to see its behavior. 
I let it play for 30 seconds and then sent a skip back command  from Kodi. The same Discontinuity error message appeared and play got freeze as well. Tried again letting it run 1 minute and then I tried 2 skips (20 seconds). After a while, the stream got back 10 seconds. So I think TVH is taking a local buffer instead of a remote command. If I send a long skip back command (say 3 minutes) play hangs until it reaches the start of the play (like timeshifting from a local file)
So, I think the problem is still with TS timestamps. This operation should be transparent to TVH because if we consider the play as a continous stream, TVH *should* play whatever server sends (obviously if the stream is TS valid). It is ok for TVH to create a local buffer but if server rewinds (that's the expected operation in remote timeshift) TVH should play the received frames

Just to check if PCR/DTS was the problem, I ran TVH with `tvheadend --noacl -l  tvh2.log  --trace pcr` and skip back 10 seconds. Part of log file is:

>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265686260
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265688628
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265690980
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265693366
>[  ERROR]:tsfix: transport stream AC3, DTS discontinuity. DTS = 4008919, last = 4610833
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265695735
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265698104
>[  TRACE]:pcr: IPTV Automatic Network/short-rtsp-test.m3u - World HD/208 World HD: change   : 6265700472

So for me this also proves, as you suggested, that PCR time is not the issue. Checking information about H264, AC3 and DTS and reviewing those traces, maybe the problem is in audio and video synchronization.
But how could I try the fix #2? (Restart the stream when a skip command is received) How could I stop RTCP/UDP stream when a skip command is received then send a RTSP PLAY command and then start the play again? (This could be a fix for skip command. I have not tested yet speed commands...)

spdfrk:
Hi,
You mentioning buffers gives me another idea, are you running low bitrate streams? I increased the iptv buffer for NAK support but this buffer could potentially cause a big delay which might be what you are seeing. Can you try to reduce the iptv buffer back to stock (or lower) and see if that helps with your media streams? 
https://github.com/spdfrk/tvheadend/commit/970d64219777d35cc187e73a93034e1655391269#diff-0468d65e0e8d50c5bb1dc9e8f20b3142L29

I added a RTSP PAUSE command right before the skip. This should stop the stream coming in before making the skip and continue, like we discussed. This could be to quick, we might need a delay or first wait on the ACK for this command.
https://github.com/spdfrk/tvheadend/blob/iptv_features/src/input/mpegts/iptv/iptv_rtsp.c#L466

faperea:
Hi,
Speaking of buffers, I was thinking that timeshift should behave like a Youtube video. Playback can be started at any position in the buffer range: start, middle, or near end. The red line in browser player shows the actual position and gray line (below red line) shows the buffer. If user starts playing in the middle and then returns to the beginning and buffer is large enough, the gray line is marked and playback starts at the new position requested by user. When the playback reaches the starting point, I think it is not downloaded again. It plays from the buffer. 
So I think it is powerful to use a local buffer to store downloaded stream and play from this buffer and download from the server if it is empty. In this case, when a DESCRIBE response is received, the buffer length can theoretically be set. In my case, I have to set it by hand because the DESCRIBE response is `clock = now-`. I set it by default to 7200 seconds (2 hours). So the PCR check or DTS check should consider anything between "now" and -7200 seconds as valid data and maybe TVH could fill the buffer when a timeshift command is sent...
I've also tried it before with an RTSP PAUSE command, but no luck! I tracked it with Wireshark and the server successfully receives the PAUSE command and also sends the ACK, but I keep getting errors. I set up a lot of `tvhinfo` prints to check if the server was doing something wrong or values are being sent incorrectly, but everything seems fine. So the problem is not on server side! When multiple jump back commands are sent, the tsfix routines appear to send an RTSP speed command when multiple DTS discontinuities are present. I'm not sure if Kodi actually sends speed commands or is the tsfix routine that send them!
For now, I'll try the buffer correctness you suggest. I'm a little busy today, so I'll give it a try tomorrow and let you know my findings.

spdfrk:
Quite busy here as well for some time, so wasn't able to go into detail about your comments. 

After some debugging I noticed the resume after pause / change playback speed was broken, fixed now.

Also didn't notice before DTS will go negative skipping before 0:00. This will cause Tvheadend to drop them:
https://github.com/spdfrk/tvheadend/blob/iptv_features/src/plumbing/tsfix.c#L241
Removing the check doesn't help, needs some more digging.

spdfrk:
Here is a small hack for testing, this will allow Kodi to time shift to a point before start (with negative DTS / time) and playback should not freeze. On my side it's working fine like this but might of course be different for your sources.

faperea:
Hi,

We're moving forward! On a virtual machine it seems to work, but when I compile it on a Raspberry 4 TVH has strange behavior. It starts off fine and timeshift commands from Kodi work, but the image freezes. Play like a PowerPoint slide! but the audio is perfect. I think it is because the local buffer does not fill when playback is skipped. I changed the buffers in iptv_private.h and made sure RPi has enough resources, but it still shows slides!
I still think that timeshift shouldn't be that complicated. It's like a remote control DVD. When a time shift control is sent from the remote, the DVD player simply moves the stream smoothly, even if it is playing TS streams or AVI movies. Tomorrow I will try to make a stop without TEARDOWN and restart the playback as if the timeshift is a "new start". I am confident that it is the solution!
Also, when using RPi 4 and the remote and Kodi, the speed commands don't work at all! It need more research!

In other topic, when reviewing the code, an RTSP GET_PARAMETER command is used to get the "position". But, as I pointed out before, this is an unusual way to get the position timer. In my case (and many other RTSP implementations) GET_PARAMETER is only used as a "ping" command to keep the connection alive. So the position variable (rp-> position) in iptv_rtsp.c is never updated. This would cause problems after 2 hours of playback, because the set_position function uses this value to send the new position to the server, but as long as the playback is within range, it will work. Therefore, functions are needed to update the start, end, and position timers, but locally. I looked for some specific implementation of RTSP but couldn't find any that can be used to retrieve data from the server while gaming.

spdfrk:
Hi,
Are you running an 32 or 64 bit OS on the rpi? 
Actually can't really say how to easily restart the stream within the current session while also improving your situation. We would need to do some handling with rtsp, plumping and property setting the position in Kodi, sounds complicated. By the way, we are already fushing the buffer to the player: https://github.com/spdfrk/tvheadend/blob/c18e7c613d574ee3777038a049ef5d1e58d64606/src/htsp_server.c#L4463 

On my side get_parameter is also used as keep alive ping but it is conveniently sending position by default. So yes in your situation we would need to replace get_parameter and describe with a local time keeping.  But how do you know where you are in the buffer? It is possible to start at the beginning or the end of the buffer so we need te position at least once.

faperea:
Hi,
RPi is 32 bits. I don't think this could be an issue because after a skip command it plays like slides with sound, but when another skip forward command is sent to return to original position (initial play position) it works smoothly. I think the problem is in Kodi. At this time, Kodi in RPi is 18.7 version for Raspbian buster. I ran Kodi in a windows PC and TVH in the RPi and timeshift worked. By the way, keep in mind that this timeshift version have a long lag! 
Also I setup a man-in-the-middle to review with Wireshark how timeshift works with an old SBC and found that no PAUSE commands are sent when timeshifting and also PLAY skip-commands have "Scale" as well as "Range" values. I also took a look at TS flow and the server makes the changes! The local player simply plays what server sends.
With Wireshark on TVH, I also noticed that there are some strange implementations for the speed commands. In TVH only the "switch-case" SMT_SPEED is reached when pause or pause-returning (play at 100% speed) commands are received. Speed is implemented sending successive position commands! Because the long lag in every skip command, this approach obviously doesn't work. It should send a "Scale" command with speed multiplicator!
I'll be busy this week so I can't test the restart option, but I took a look at the code and I think it should work (I called `rtsp_init` function and nothing special happened). If we consider a skip command as a new start, then we could send a STOP command (without TEARDOWN) and then restart everything as after the DESCRIBE-> PLAY commands. As you pointed out, the code would not be easy to change, because RTSP is two open channels, but using helper variables to store the start, end, and current can easily solve the problem. Those are the "important" values ​​to uphold.
Regarding GET_PARAMETER, I tested before your new code changes adding a new variables to hc struct such as "current" and "range". That way when a initial ACK for a PLAY command is received (`rtsp_play_decode`) all those new vars are set. I tried without luck to update "end" and "start" timers as well as "current" and "range" in GET_PARAMETER case. But that's the idea. The initial "current" value is set when the first play command ACK is received and every time when skip command is also received. It could be updated when GET_PARAMETER is called, just adding elapsed seconds since last call. I don't know exactly where the timer is set to call periodically GET_PARAMETER command but based on "local timeshift" a regular timer should be called to update time values (also times sent to Kodi or it will be always set to initial timestamps)

spdfrk:
Ok now I understand your setup better, I setup a RPi 3 with latest LibreELEC (Kodi 18.8) and can reproduce. Hardware codecs are broken (OMXPlayer, MMAL) but software decoding is working (while a little underpowered). So somewhere it's dropping the negative dts's. With positive dts remote TS works fine and overall playback is best with OMXplayer. Don't really see a reason to implement this stopping the stream in between skips.

Test setups that do work with negative dts's:
PC Linux Mint / Windows 10 + Kodi 18.8 (software & VDPAU).
Amlogic S905X, S922X + latest CoreELEC. (I can recommend the S922X, the overall experience is much faster/smoother/better PQ)

I agree, the speed implementation with Kodi could be better. But in my case setting scale to e.g. 2.0 will cause the the server to just start skipping packets (loads of dts CE's) and the result is unusable. Probably best to forget about this for now and learn to use skipping :).

I will check if we can get the dts offset implemented, this should fix the playback issues on RPi. Also the playback time in Kodi  will than show up as being on the end of the buffer and won't go negative any more, I guess this looks better.

spdfrk:
Here is a prove of concept to try, it should fix the RPi playback issue. I will add an timeshift status thread next to keep the time in Kodi updated.

faperea:
Hi  spdfrk,

Finally, it is working! I was able to open and see an RTSP stream both in my PC as well as in RPi. 
In both of them skip back or forward worked!
There are some issues left:
- In Kodi when the play starts, the slider is at left, meaning that the beggining is at buffer start (2 hours in my case) but really play starts "live". So I have to do a full skip forward and then slider is at right meaning tha play is live-tv. I am trying to change the way end , start and current variables are set at beggining to avoid this initial false jump
- Also in my case, buffer start and end times are not updated. They remains constant with the time values ​​captured when playback started.
- If no initial jump forward is performed, the slider continues to move in time, as expected, but points to the wrong time. After a while, say 1 hour, and because the timers are not refreshed, if a full buffer skip forward command is sent, the play reaches the end of the buffer which was 1 hour ago. That means it rewinds!! Yes! It is confusing behavior. It's even funny for me. As a user that is fixed by sending a full skip forward command at play beginning and everything works as expected.
- There is still a long delay when commands are sent. Also, the logs keep showing warning messages and I can't identify why it takes so long to apply a jump (between 3 and 10 seconds or so). So short jumps (10 seconds or less) don't perform well.
- I took some time to review why speed commands are not working. Reviewing logs and with Wireshark the speed commands never are sent! Instead, all Speed commands are treated as Skip commands. Only when pause (speed=0%) or "un-pause" (speed=100%) commands are sent from Kodi, the Speed functions are reached. I also reviewed pvr.hts code and problem seems to be there. I think Speed commands are coded as repeatedly Skip commands! I already sent a question / issue report to pvr.hts...

I still need to change `npt` commands by `clock` commands. Server receives `npt` but always answer with `clock`. How could I sent to you those functions that implements `clock` SDP commands and my messy code changes? (I have to confess that I didn't add all code needed to choose between those two formats correctly, but I could fix it...)

spdfrk:
Hi,

- I implemented a fall back for the buffer updates, should solve your funny stuff :)

- Yes it would be nice if we could force the position in Kodi, same problem on my side. I tried but it doesn't seem possible with the current implementation. If you find a solution please share.
- Not sure if the response time can be much faster, even local TS has some seconds delay. Your welcome to try of course.
- Interested in what the pvr.hts team thinks about this. I don't expect this will solve the issue though, see my previous note on dts skipping during fast forward. If you change the scale value it's just skipping on server side instead and breaks client side.

If you want to share some code it's probably easiest is to fork my repo and apply your changes. If you then send me a pull request we can discuss and I can then apply the changes to this pull request.

Flole998:
... and if you're done please squash your commits together and @ me so I can do the review, some testing and merging ;)

faperea:
Hello,
> 
> If you want to share some code it's probably easiest is to fork my repo and apply your changes. If you then send me a pull request we can discuss and I can then apply the changes to this pull request.

I am an electrical engineer, so software development is not my best skill! So I'll be reading how the heck a repository forks and how to "send a pull request". I'll be then trying to "clean up" my code so I can share with you all the "enhancements" to your code so that RTSP accepts the `npt` commands as well as the` clock` commands. Patience, please! 
In the meantime, I'll also test the latest changes you proposed.

I should note that in my case I am setting the remote buffer length manually because the server responds to the DESCRIBE directive with a `range: clock = now-` and includes a `Timeshift-Status: 1` directive in the header message. I don't use it, but as a future TODO, this could be used to set if the server accepts time-shifting functions.
I set a buffer length to 7200 seconds, that is 2 hours. (It could be set to 3600 or 1 hour or whatever, but for now it's a hard-coded value)
Also, I forgot to tell you that the redirect is not working as expected. I made some changes by adding `#define RTSP_STATUS_REDIRECTION 302` in http.h to add message 302 which is` RTSP / 1.0 302 Moved Tempoually` and in `iptv_rtsp_header` I changed the code, but always when a new `r ​​= rtsp_describe (hc, rp -> path, rp-> query) `is reached, the value r is 4 which means `EINTR 4 / * System call interrupted * /`. 
That was the end of my redirect testing excursion!

spdfrk:
Don't worry about it, I'm no software engineer either. 

I guess your STB is not following the RFC since `now-` is only used for live streams? Not sure how to setup the buffer limits then.  https://tools.ietf.org/html/rfc2326#section-3.6 Can you share a Wireshark capture of your STB to understand better?

Regarding redirects, I agree we should probably accept return codes 301, 302, 303. Code 300 sounds interesting. 
https://tools.ietf.org/html/rfc2326#page-24

I don't think we need `Timeshift-Status: 1` since the user needs to enable the remote time shifting anyway.

Flole998:
Could you please squash your commits together so I can review them all at once?

faperea:
Flole998,

I will gladly "squash" my confirmations if I find how... In the meantime, I disconnected from this PR for a moment, but I think it's not over yet! Some corrections are needed, as well as information that I should share. Be patient! I'm trying to take some time to do all my homework!

spdfrk:
@Flole998 I rebased and squashed my commits, thanks.

src/rtsp.c
```diff
@@ -253,16 +284,45 @@ rtsp_setup( http_client_t *hc,
...
 
-int
-rtsp_describe_decode( http_client_t *hc )
-{
-  http_arg_t *ra;  
+int rtsp_describe_decode(http_client_t *hc, const char *buf, size_t len) {
+  const char *p;
+  char transport[64];
+  int n, t, transport_type;
+  //char *argv[32],
```
Flole998, line 291: Why is there uncommented code here?

src/rtsp.c
```diff
@@ -253,16 +284,45 @@ rtsp_setup( http_client_t *hc,
...
 
   /* TODO: Probably rewrite the data to the htsmsg tree ? */
+#if 0
```
Flole998, line 294: Any reason why this is here?

Flole998 (changes requested):
I noticed some unreachable code here, could you please check why it was added and if it actually needs to be there?

spdfrk:
@Flole998, just some old debug code, I guess we can assume this is no longer necessary.  I cleaned it up.

src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -89,26 +96,92 @@ iptv_rtsp_header ( http_client_t *hc )
...
+      p = http_arg_get(&hc->hc_args, "Location");
+      if (p == NULL) {
+        tvherror(LS_IPTV, "received code 303 from RTSP server but no new location given for '%s'",
+            im->mm_iptv_url_raw);
+        return -1;
+      }
+      tvhinfo(LS_IPTV, "received new location from RTSP server '%s' was '%s'", p,
+          im->mm_iptv_url_raw);
+      urlinit(&url);
+      if (urlparse(p ? : "", &url) || strncmp(url.scheme, "rtsp", 4) != 0) {
```
Flole998, line 137: p can not be NULL here as in line 132 it would have returned for `p == NULL`. So the `p ? : ""` can be replaced by a simple `p`.

src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -270,26 +382,176 @@ iptv_rtp_header_callback ( iptv_mux_t *im, uint8_t *rtp, int len )
...
+
+static void rtsp_input(void *opaque, streaming_message_t *sm) {
+  int type = sm->sm_type;
+  rtsp_st_t *pd = (rtsp_st_t*) opaque;
+  iptv_mux_t *mux;
+  streaming_skip_t *data;
+  rtsp_priv_t *rp;
+
+  if (pd == NULL || (sm->sm_s == NULL && pd->im == NULL)) {
+    streaming_target_deliver2(pd->output, sm);
```
Flole998, line 473: You check if `pd == NULL` above and here you refer to `pd->output` if `pd == NULL`. That will cause a crash if pd is NULL.

src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -270,26 +382,176 @@ iptv_rtp_header_callback ( iptv_mux_t *im, uint8_t *rtp, int len )
...
+    return;
+  }
+
+  switch (type) {
+  case SMT_GRACE:
+    pd->im = (iptv_mux_t*) ((mpegts_service_t*) sm->sm_s)->s_dvb_mux;
```
Flole998, line 479: You do a check in Line 472 to see if `sm->sm_s` is `NULL`. Here there is no check if `sm->sm_s` is NULL and if it is, there would be a crash right here. So either the check above is not needed/correct or another check is needed here.

spdfrk:
@Flole998, thanks for the feedback. Some stupid mistakes during change on change. Latest update should fix these issues.

Flole998 (approved):

Flole998:
@spdfrk Unfortunately there seems to be another issue related to this, could you please look into it: https://tvheadend.org/issues/6024

Looks like you moved some functions to different #ifdefs

spdfrk:
@Flole998  I will fix it. Seems I started the ifdefs without finishing it.

spdfrk:
@Flole998 #1395 will fix the compilation issue. Unfortunately I can't functionally test it.

PR #1348, opened 2020-06-16
hdhomerun: Add HDHomeRun server emulation support for LiveTV only (#4461)

azlm8t:
Tested in the forums by Shane Kuzmanic against Plex.
Minimal testing also done against Jellyfin.

We now offer the ability to pretend to be an HDHomeRun server
to expose channels for LiveTV for some media players such
as Plex.

DVR functionality is not exposed.

This only works on media players that allow IP:Port
number configuration.  It is hooked in to the webui
and only exists if the feature is compiled in and
explicitly enabled in the GUI.

To use:

- Add a user and enable persistent authentication via
  Config->User->Passwords->Persistent Authentication Enable.

- Use this user for the HDHomeRun server setup in
  Config->General->Base->HDHomeRun/Local Username.

Then manually add the device to your client using the
IP address and port of the TVHeadend server, such as
192.168.0.1:9981

Auto-discovery of the TVHeadend server by downstream
media clients does not work since clients seem to use
different mechanisms.

So, although many clients (such as TVHeadend) use
a broadcast at 65001 to locate clients, one of
the popular media players does not use this mechanism.

The persistent authentication allows the client to stream
LiveTV from the TVHeadend server using the credentials for
the given user.  Using a user allows us to do filtering
such as only allowing specific channels to be passed
through to the downstream clients.

For "DeviceAuth" we use a (hard-coded) randomly generated
value since it has to be 24 lowercase/digit characters long.

For DeviceID, we base it on the server name.  Technically
a DeviceID needs to match a particular format to validate
against a CRC, but no clients seems to enforce this.

The feature is configurable.  If the feature is disabled
then we hide the options in the GUI rather than compile
them out completely.  This ensures that if the user
accidentally compiles with the options disabled
then they will not lose their configuration.

Issue: #4461

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -25,6 +25,7 @@
 #include "http.h"
 #include "tcp.h"
 #include "webui.h"
+#include "htsmsg_json.h"
 #include "dvr/dvr.h"
 #include "filebundle.h"
 #include "streaming.h"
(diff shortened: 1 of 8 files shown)


Flole998:
There are 2 things in here that should be re-considered:
- Manufacturer and Model, could those be changed to be Tvheadend instead of using a manufacturers name here without their permission?
- I do not like that authentication tokens are given out to everyone who requests the right URL. Instead, IP based authentication should be used (so basically remove the token from the URL again). The current behavior can be achieved through that aswell by configuring a wildcard that allows access for everyone. Let's not pretend that there is authentication while we give everyone access if they request the right URL.

azlm8t:
I've changed the manufacturer.

It's difficult to know for sure what checks the media players make for model and manufacturer, and what limitations there are for model number.   Perhaps someone with access to Plex can verify what happens if they change the model number.

As I discussed in the forum, the rights are based on the user since it offers some benefits.  In particular, it allows filtering of channels since many media players have limitations on how many channels they can handle.

However, it also allows the filtering by IP address that you mention.   So users who explicitly enable the functionality can filter the IP address to their media player only.  All other IP addresses are blocked so even if they use the correct URL they do not gain access.  This limitation on credentials is because clients do not specify credentials when connecting to genuine HDHomeRuns.

azlm8t:
Ideally, we'd make it all configurable, but I didn't want to clutter the UI with undue configuration.

Flole998:
Filtering of channels also works with IP based authentication.

The problem I have with this is that it gives users a false sense of security. Using IP based authentication only is reflecting the truth: Everyone from that IP can access the streams and not just HDHomeRun devices. And if a HDHomeRun client indeed supports authentication they can use it.

src/webui/webui.c
```diff
@@ -1597,6 +1597,271 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+static int
+hdhomerun_server_verify_callback(void *aux, const char *passwd)
+{
+  return 1;
+}
+
+
+static const char *hdhomerun_get_server_name(void)
+{
+  return config.server_name ?: "TVHeadend";
```
Flole998, line 1626: "Tvheadend" is the official name, so we should not use a capital V and H here.

src/webui/webui.c
```diff
@@ -1597,6 +1597,271 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+  htsbuf_append_str(hq, "{ ");
+  htsbuf_append_and_escape_jsonstr_nv(hq, "FriendlyName", server_name);
+  htsbuf_qprintf(hq, ", "                                               \
+                 "\"BaseURL\" : \"http://%s:%u\", "                     \
+                 "\"DeviceAuth\": \"3xw5UaJXhVShHEBoy76FuYQi\", "       \
+                 "\"DeviceID\": \"%08X\", "                             \
+                 "\"FirmwareName\": \"hdhomerun_atsc\", "               \
+                 "\"FirmwareVersion\": \"20200101\", "                  \
+                 "\"LineupURL\": \"http://%s:%u/lineup.json\", "        \
+                 "\"Manufacturer\": \"tvheadend\", "                    \
```
Flole998, line 1707: t should be a capital T.

src/webui/webui.c
```diff
@@ -1597,6 +1597,271 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+  htsbuf_qprintf(hq, "<root xmlns=\"urn:schemas-upnp-org:device-1-0\">"
+                 "<specVersion>"
+                 "<major>1</major>"
+                 "<minor>0</minor>"
+                 "</specVersion>"
+                 "<URLBase>http://%s:%u</URLBase>"
+                 "<device>"
+                 "<deviceType>urn:schemas-upnp-org:device:MediaServer:1</deviceType>"
+                 "<friendlyName>%s</friendlyName>"
+                 "<manufacturer>tvheadend</manufacturer>"
```
Flole998, line 1847: See above

flippz: Can you change the number of tuners without problems? I tryed changeing it to 20 and tvheadend would not start any channels after that...
BTW, what happened to tvheadend.org?

flippz:
Can you change the number of tuners without problems? I tryed changeing it to 20 and tvheadend would not start any channels after that...
BTW, what happened to tvheadend.org?

src/webui/webui.c
```diff
@@ -1597,6 +1597,271 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+  htsbuf_qprintf(hq, ", "                                               \
+                 "\"BaseURL\" : \"http://%s:%u\", "                     \
+                 "\"DeviceAuth\": \"3xw5UaJXhVShHEBoy76FuYQi\", "       \
+                 "\"DeviceID\": \"%08X\", "                             \
+                 "\"FirmwareName\": \"hdhomerun_atsc\", "               \
+                 "\"FirmwareVersion\": \"20200101\", "                  \
+                 "\"LineupURL\": \"http://%s:%u/lineup.json\", "        \
+                 "\"Manufacturer\": \"tvheadend\", "                    \
+                 "\"ModelNumber\": \"HDTC-2US\", "                      \
+                 "\"TunerCount\": 6 "                                   \
```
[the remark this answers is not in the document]
Flole998: Unfortunately it is not possible to identify the tuner count. For the SAT>IP Server there are fields to enter the number of advertised tuners but I don't think the number actually matters for a client.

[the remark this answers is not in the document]
flippz: At least i know plex counts how many are in use so i guess it will stop when it reaches that number. But will tvheadend work if i change that number? I did and did configure and autobuild.sh and the installed and after that i could not tune to any channels, be it in plex or directly in tvheadend. Might be the build so i could try again...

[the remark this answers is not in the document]
rkkoszewski: @Flole998 Oww, I thought TVheadend keeps track of the free tuners. In any case I know for certain that at least Plex does take the tuner count into account when connecting to HDHomeRun. Having it show more tuners than are available may also produce funny errors on the clients connected to it (Although I'm unaware of how HDHomeRun signals when when the tuners are in use from another application). I would suggest to add a menu entry for setting this similar to the "SAT>IP Server" then, if not now, at least in the future.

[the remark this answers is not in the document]
Flole998: Keeping track of tuners works until you use IPTV, then you have basically an infinite amount of tuners. I don't think the infinity sign would be accepted by plex though ;)

Does Plex properly distinguish between Tuners and channels? You can watch 10 Channels on 1 Tuner in certain cases.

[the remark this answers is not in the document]
flippz: IPTV has a pissibility to set limits based on your supplyer when you add it to the network. And tvheadend counts and if you reach the limit you set it wont tune to that channel. It works good in tvheadend but dont know how it works in combination whith plex setting a limit aswell.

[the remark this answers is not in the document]
Flole998: It's a possibility, not a requirement. So if someone didn't set it the tuner count is infinite, which is obviously not a valid value. That's why automatic counting of tuners is not an option.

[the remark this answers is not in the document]
flippz: I see.

azlm8t:
Now we've had a few people testing basic functionality in the forum, I've updated the patch.

I've made the tuner count configurable.  So, if Plex prevents opening more than X streams, you can set the count to a higher value.  I'm wary of just defaulting the tuner count to a very high value in case clients are doing some resource allocations for each tuner, so I've left the default at 6 which is probably adequate for basic home use.  As mentioned, Tvheadend does not use that count itself since one tuner could serve multiple channels.

We now send FirmwareVersion from Tvheadend's build version, and I've changed the server name, though there is a lot of existing inconsistencies in the capitalization.

I've currently left it using a username since this allows us to add a user and do filtering by IP, but also extra features such as filtering by tags or channel numbers.  If we moved to just IP authentication then we'd lose those features.  We could possibly make it configurable if a user is needed and just rely on anonymous streaming access, but I think it's possibly clearer having "hdhomerun" user and seeing that in the UI.  It's an unfortunate consequence that clients only allow IP and post number to be specified (as separate fields), so they can't pass authentication credentials to use in the auto-generated playlists.

src/webui/webui.c
azlm8t: Updated.

src/webui/webui.c
azlm8t: Updated.

src/webui/webui.c
azlm8t: Updated.

Flole998:
I will not merge something that bypasses authentication like this. When authentication for a user is enabled it will be enforced, giving out access keys to whoever requests it is not something we should do. While a user thinks it's all secure because a password is configured it really isn't because anybody can just ask for an authentication token.

Even with this removed you can add a user, you just need to call it *, and that represents the reality: No protection against anything, anyone who has access can access whatever they want.

flippz:
Tryed using it in emby and it wont start there. Tryed with xteve and it worked at once so dont seem to be the fault of emby:
Error:
2020-07-17 16:30:09.088 Info LiveTvManager: Opening channel stream from Emby, external channel Id: hdhr_101
2020-07-17 16:30:09.088 Info LiveTV: Streaming Channel hdhr_101
2020-07-17 16:30:09.088 Info HdHomerunHost: GetChannelStream: channel id: hdhr_101. stream id: native_7ce9fd452946804bf450f723a11f06ea_b41d4305a80a79fb34980b9cbd994d3a profile: native
2020-07-17 16:30:09.089 Info SharedHttpPipelineSource: Opening SharedHttpPipelineSource Live stream from http://192.168.xxx.xxx:9981/stream/channel/dd72b9d5b5e1eecf3f183ca568407cc3?profile=pass&auth=(secrect)?transcode=none
2020-07-17 16:30:09.089 Info HttpClient: GET http://192.168.xxx.xxx:9981/stream/channel/dd72b9d5b5e1eecf3f183ca568407cc3?profile=pass&auth=(secrect)?transcode=none
2020-07-17 16:30:09.092 Info HttpClient: Http response 401 from http://192.168.xxx.xxx:9981/stream/channel/dd72b9d5b5e1eecf3f183ca568407cc3?profile=pass&auth=(secrect)?transcode=none after 3ms. HeadersServer=HTS/tvheadend, Cache-Control=no-cache, WWW-Authenticate=Basic realm="tvheadend", Connection=close
2020-07-17 16:30:09.098 Error HdHomerunHost: Error opening tuner

flippz:
Seems that the problem is that Emby adds the ?transcode=none to the end of the url....

flippz:
I tryed without Persistent authentication code now and just added a ip that does not need credetials and it works fine in Emby. So seems Emby does something different that Plex and Jellyfin.

rkkoszewski:
> Seems that the problem is that Emby adds the ?transcode=none to the end of the url....

That smells to me like a bug in Emby. Given that Emby assumes the URL to have certain format, and disregarding that the URL might already have a question mark, producing (as far as I'm aware) an invalid URL with double '?'. I wouldn't bother with Emby and file a bug on them to check their URLs to be valid. Emby has an official Tvheadend plugin anyway.

Flole998:
It will automatically be fixed when the authentication is implemented properly. Then it's a requirement anyways to use IP based authentication (or use a client that Supports authentication by username/password), which will apparey fix this issue.

CvH:
> Emby has an official Tvheadend plugin anyway.

Just for info, the plugin is basically completely broken and not useable for years.

azlm8t:
I'd left a comment (and potential fix to try) in the boards since someone else encountered the same problem with Emby.  Unfortunately, I can't test Emby so I was awaiting feedback there.
https://tvheadend.org/issues/4461#note-144
The idea is to fake an extra argument so the auth token is then passed through ok despite the bug in Emby.
When you have time, could you try making the change suggested in there?
Basically alter src/webui.c around line 1558 from:

`       sprintf(url, "http://%s:%u/stream/channel/%s?profile=pass%s%s",
`
to
`       sprintf(url, "http://%s:%u/stream/channel/%s?profile=pass%s%s&extraarg=t",
`

chkuendig:
I've been working on a tvhProxy fork for a while as the original version wasn't working anymore with newer Plex updates. One of the changes required was to add SSDP discovery. Without it, the tuners sometimes randomly get "dropped" during the setup. 

Was it a conscious decision to not add SSDP discovery or is that something which could be added at a later point?

azlm8t:
I took a look through your code and liked your ideas for "72 hours of full details" in xmltv and "no programme."  I had added different "xmltv formats" (such as only basic details, no descriptions) for low memory devices, so perhaps those features should be added in the future as a separate "xmltv format."

I don't have Plex, but I had taken a look at SSDP since I'd noticed it in some Plex logs I'd downloaded.  I believe Plex works without SSDP but SSDP ensures that the icon for "available" is always enabled if you go back in to settings?

We have SSDP implemented for SatIP.  The main reason I didn't implement it for this patch was that it looked like it required some refactoring (see satips_upnp_discovery_received, etc.) and I wouldn't be able to test properly for either satip or this new functionality.  In theory, it should be the case of sharing all the code from that satip function and just call a service-specific "reply".  So, yes, it could be added.

I'd read that some projects integrate with "minissdp" to do the processing since only one process can easily handle the ssdp requests.  Indeed, Jellyfin now has a change request to disable SSDP and let the other process do it to avoid conflicts.  I always had to disable ssdp on another pvr I have used since otherwise I get conflicts on the port 1900 since I have other processes that do upnp.

shanekuz:
Has this been added into the build or it this still waiting on something?  Im happy to keep testing but its been stable for quite a while now.  The SSDP would be nice as it does show in plex as offline after a while but still works

Flole998:
This is still waiting for the changes I requested. Without those it won't make it in.

anthonylavado:
> I will not merge something that bypasses authentication like this.

Is this the still main blocker on this issue?

On the Jellyfin side, we (currently) suffer from the same issue as Emby would, but we'll do our best to fix anything we can on our side?

shanekuz:
If the issue is around authentication can we just post a TVH username and password in the lineup.json path?  Of am i missing whats meant with this issue?

Flole998:
Not sure if I understand it correctly, but if those credentials are exposed and anybody can get them by requesting a specific URL or if those are even broadcasted on the network that would make the issue even worse: Right now at least the password is protected, publishing it would allow everyone to get the password of a user, that's even worse. There are 2 options (which both require the same changes to this PR): The client sends proper authentication or IP based authentication is used. Then it would be consistent with how all other components of Tvheadend are using authentication (SAT>IP being the exception but I don't like that either and opened an issue a while ago that this would need chaging).

Leatherface75:
HDHomeRun aren't using any password so i don't know how an emulation with password support would work.

Flole998:
That's why I mentioned two options. If the first one doesn't work for a specific client (or even all clients) then the second option will definitely work. And what's pretty cool about that: All those features are already included in Tvheadend, so just the authentication bypass/mapping would have to be removed.

Leatherface75:
Maybe i am missunderstanding but i can't see how this would work with any sorts of bypassing authorisation.
There is no support for that with HDHomeRun devices.

rkkoszewski:
@Leatherface75  You can create a user without password in Tvheadend and white list a specific IP for that user. That way the specific user will be able to login from that specific IP without the need for a password. That way you don't need to bypass anything and just use the user for the streaming (Witjout password). And this behaviour is already part of Tvheadend.
I would just document this requirement somewhere as this is a bit of a non obvious setup in my opinion.

shanekuz:
> That's why I mentioned two options. If the first one doesn't work for a specific client (or even all clients) then the second option will definitely work. And what's pretty cool about that: All those features are already included in Tvheadend, so just the authentication bypass/mapping would have to be removed.

What if in the setup page we force an ip entry for the plex server and lock it by IP rules?  If we make it so you can not leave this blank and can not put a rage then it will only expose the XML to one IP.

rkkoszewski:
@shanekuz What if you want to have multiple clients that connect to the HDHomeRun server emulator? Or what if you want to have the server open for a whole subnet or even everybody? (This can be all achieved with the current Tvheadend permission system)

Leatherface75:
> @Leatherface75 You can create a user without password in Tvheadend and white list a specific IP for that user. That way the specific user will be able to login from that specific IP without the need for a password. That way you don't need to bypass anything and just use the user for the streaming (Without password). And this behaviour is already part of Tvheadend.
> I would just document this requirement somewhere as this is a bit of a non obvious setup in my opinion.

And how do TVheadead know which user to use?

chkuendig:
I think you can create a user named * which is picked if there's no authentication.  See also the answer above:

> Even with this removed you can add a user, you just need to call it *, and that represents the reality: No protection against anything, anyone who has access can access whatever they want.

I used this in my own trials:

image

j0nnymoe:
Been testing this for a couple of days long side Emby 4.4.3.0 - All seems to be working well.

flippz:
If i get this correct if the changes you propose are implemented then this will go into the main build?
Could azlm8t implement them? Hes the one that made this PR isent he?
Had i known any coding i would gladly implement them but i have no idea...
Iv been using it the way you say here. With using IP assigned instead if username and password and it works 100% against EMBY. Never any problems :)

shanekuz:
> @shanekuz What if you want to have multiple clients that connect to the HDHomeRun server emulator? Or what if you want to have the server open for a whole subnet or even everybody? (This can be all achieved with the current Tvheadend permission system)

This would only be limiting connection between the TVH server and the plex server so not everyone can query the XML list as iu think this is the issue as asking for the list file gives an open password.

shanekuz:
The solution is working really well, can someone that follows the issue with getting this included in the build please explain what the issue is?  Would be really good to close this out.

dazedcrazy:
> 
> 
> The solution is working really well, can someone that follows the issue with getting this included in the build please explain what the issue is? Would be really good to close this out.

@shanekuz it's around the use of authentication, see here https://github.com/tvheadend/tvheadend/pull/1348#issuecomment-657271891

dazedcrazy:
I've been trying this for a while and aside from the changes @Flole998 requested it works fine. I do have one issue and was wondering if someone could help. For EPG data i use **IP:Por/xmltv/channels** but everything that comes into Plex has a _New_ tag, well anything that has an episode number will show up as _New_ if it's new or not. 
image

I'm gussing I need to change something in the xmltv.c to make it work?

todd121383 (approved):

flippz:
Come on someone please make the changes needed to get this included :)

src/webui/webui.c
```diff
@@ -1597,6 +1598,304 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+ * since HDHomeRun clients usually do not allow entry of credentials.
+ *
+ * This is called via the "access_get" call which does IP checking
+ * (configured via the Users/IP Blocking Records tab in expert
+ * setting).
+ */
+static int
+hdhomerun_server_verify_callback(void *aux, const char *passwd)
+{
+  return 1;
```
DavideFioravanti, line 1621: Changing this line to `return 0;` is enough?

Flole998: No, there are more changes needed in the lines below.

mr-onion-2:
Hi

I wasn't sure whether to post this here or whether it is an issue for LinuxServer.io.

I am trying to build a Docker image for this branch using the Dockerfile at https://github.com/linuxserver/docker-tvheadend. I've updated the Dockerfile to use the azlm8t/tvheadend branch.

It fails with the error below. A complete log is [here](https://gist.github.com/mr-onion-2/f3bfa367ff845bbc59559bea42de5ca0)

> Building src/webui/static/intl/tvh.zh-Hans.js.gz
> WEBUI std finished
> make[2]: Leaving directory '/tmp/tvheadend'
> make -f Makefile.webui WEBUI=debug compile-debug
> make[2]: Entering directory '/tmp/tvheadend'
> WEBUI debug finished
> make[2]: Leaving directory '/tmp/tvheadend'
> make[1]: Leaving directory '/tmp/tvheadend'
> CC              src/tvhlog.o
> src/tvhlog.c: In function 'tvhlog_thread':
> src/tvhlog.c:352:9: error: 'strncpy' specified bound 512 equals destination size [-Werror=stringop-truncation]
>   352 |         strncpy(buf, tvhlog_path, sizeof(buf));
>       |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> make: *** [Makefile:641: /tmp/tvheadend/build.linux/src/tvhlog.o] Error 1


Thanks

deathbybandaid:
Not to take away from anything in this PR, but I'm the developer of fHDHR, and I have a variant of the script designed to work with tvheadend (link below).

I originally wrote fHDHR to go between tvheadend or nextpvr, so that I could share my QuadHD tuner with Emby and Plex.

fHDHR emulates being an HDHR, as well as an m3u tuner. This works with Plex, Emby, Channels, and xTeVe.

I personally use this with my tvheadend install.

https://github.com/fHDHR/fHDHR_Tvheadend

src/webui/webui.c
```diff
@@ -1597,6 +1598,304 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+}
+
+
+static const char *hdhomerun_get_server_name(void)
```
rkkoszewski, line 1625: I'm not much of a "law ninja", but given that HDHomeRun is a registered trademark, couldn't it bring legal problems for TVHeadend to have the name "hdhomerun" in the source code? Perhaps an alias like "hdhr" could be a better fit for the functions and variables. Perhaps somebody with more insight into laws could give some more info on this..

src/webui/webui.c
```diff
@@ -1597,6 +1598,304 @@ page_play_auth(http_connection_t *hc, const char *remain, void *opaque)
...
+    return http_noaccess_code(hc);
+
+  char http_ip[128];
+  htsbuf_queue_t *hq = &hc->hc_reply;
+  const char *server_name = hdhomerun_get_server_name();
+  const uint32_t deviceid = hdhomerun_get_deviceid();
+
+  tcp_get_str_from_ip(hc->hc_self, http_ip, sizeof(http_ip));
+
+  /* The contents below for the discovery message are based on tvhProxy */
```
Flole998, line 1707: The comments here are completely inconsistent. I think every possible way of writing a comment can be found in this function. That should be changed so they are all in the same format.

Flole998:
Closing due to inactivity for now. Once this PR has been updated it can be re-opened and reviewed again.

xF4m3:
Hey, I am really looking forward for this feature is there still something happening here or is this feature dead for now? 

What are the current alternatives that are working? I tried the LiveTVH plugin for Plex but that doesn't work.

Flole998:
The PR was abandoned apparently, so unless someone picks it up and addresses the issues mentioned above it's pretty much dead.

CvH:
looking at @azlm8t github page its unlikely that anything happens here :/

thats sad, its a great feature to make Tvheadend compatible to more software, for some reason the hdhomerun stuff is basically "standard" these days (likely because dead simple to use).

user897943:
Flole, I saw that you've been maintaining this package and for that I am obviously grateful, without your work this would not survive. Do you think it would be possible to spend some of your time to get this working? I wanted to add my public comment stating this is definitely wanted. I am sure that there are others who also need this for various reasons but are discouraged from voicing it because it is "pretty much dead". This would probably eliminate all issues with using plex hdhr proxies which all of us using them seem to have.

PR #1351, opened 2020-07-04
Fix scraping of 'new' flag from UK EIT.

banburybill:
1. The regular expression must contain a subgroup match to be recognised as
   a match.
2. Follow xmltv.c and set flag to 1.


Flole998:
Please squash your commits so they can be merged.

PR #1352, opened 2020-07-11
Revert "HTSP v35: Add support for recording file size"

Flole998:
Reverts tvheadend/tvheadend#1332 as dataSize is already a property.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -3077,8 +3077,6 @@ dvr_entry_class_save(idnode_t *self, char *filename, size_t fsize)
             htsmsg_add_s64(c, "start", s64);
           if (!htsmsg_get_s64(e, "stop", &s64))
             htsmsg_add_s64(c, "stop", s64);
-          if (!htsmsg_get_s64(e, "size", &s64))
-            htsmsg_add_s64(c, "size", s64);
           htsmsg_add_msg(l, NULL, c);
         }
       }
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -50,7 +50,7 @@
 
 static void *htsp_server, *htsp_server_2;
 
-#define HTSP_PROTO_VERSION 35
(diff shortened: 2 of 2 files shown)


ksooo:
Thank you for not even waiting with the revert until I have changed the Kodi Addon to work again.

ksooo:
I have to apologize. "dataSize" is documented in the official HTSP documentation (https://tvheadend.org/projects/tvheadend/wiki/Htsp). 

Nevertheless, documentation is horrible outdated. It's at v23 level. Current is v34. It would really be helpful if it would be updated at some point in time.

Flole998:
I have added a warning to the documentation for now and a notice that it would be great if someone updates it.

ksooo:
Well, may I kindly request to revert the revert? 

A recording can consist of several files. The property "dataSize" returns the sum of the size of all recording files. The property "size" I have added with #1332 returns the size for every single file of the recording.

Flole998:
Correct me if I'm wrong, but in Kodi there is only a single entry for each recording so only a single size attribute is necessary? The individual file sizes don't really matter for anything. Is it really necessary to know the size of each file that a recording consists of? What exactly is the benefit of doing that? It would waste CPU Power on Kodi by doing calculations that are not necessary and would use more bandwidth (not much, but a few bytes). I have no problem with re-adding it but this time I want to make sure it's the right decision to add it.

ksooo:
> but in Kodi there is only a single entry for each recording so only a single size attribute is necessary?

At the moment, yes. But I plan to change that. To my knowledge, you may end up with a multi-file recording if recording is interrupted due to some backend hickup and then gets resumed. For this kind of thing, I would like to have to entries in Kodi.

Flole998:
That is correct, for example a crash during a recording for example will cause multiple files.

Please use dataSize in Kodi for now instead of summing up the size of the individual files, I will add this commit again though as it's not a complete duplicate.

PR #1359, opened 2020-10-04
update Makefile.ffmpeg

fat-fred:
-Include NASM 2.15.05
-Include newer x264
-Include x265 3.4
-Include VPX 1.9.0-rc1
-Include Theora 1.2.0alpha1
-Include fdk-aac 2.0.1
-Include NVCODEC 10.0.26.1
-Include ffmpeg 4.3.1
-remove ffmpeg.libx265.diff - its now in ffmpeg includet
-edited libx265.pie.diff for newer x265


Makefile.ffmpeg
```diff
@@ -44,69 +44,72 @@ BSFS           = h264_mp4toannexb hevc_mp4toannexb
...
+LIBX265_URL    = https://github.com/videolan/x265/archive/$(LIBX265_TB)
+LIBX265_SHA1   = eef65f2f9518f3396825675aa86bdf82a55aa0d9
 ifeq ($(CONFIG_PIE),yes)
 LIBX265_DIFFS  = libx265.pie.diff
 else
 LIBX265_DIFFS  = libx265.pic.diff
 endif
 
-LIBVPX_VER     = 1.8.2
+LIBVPX_VER     = 1.9.0-rc1
```
CvH, line 71: there is a final 1.9.0

Flole998: Thanks for your review!

@fat-fred It would be great if you could change this.

Flole998:
Unfortunately there is an issue with this change: https://www.tvheadend.org/issues/5957

A patch seems to fail, probably because the sources have changed.

CvH:
what does https://github.com/tvheadend/tvheadend/blob/master/support/patches/ffmpeg.vaapi_encode.diff at all ?
we build ffmpeg without it at LibreELEC and nobody complained yet (for years) that encoding is not working

fat-fred:
Have a look in it.
Make a first pull request. But have not vaapi. i try to test it in my docker. so please wait this time ;-)

CvH:
the patch was made against ffmpeg 3.1 that added vaapi support to ffmpeg

likely this was a workaround for some stuff that is already fixed, like said we use tvheadend with a self compiled ffmpeg that never had these patch and vaapi encoding works like a charm

PR #1363, opened 2020-10-09
Ffmpeg fix

fat-fred:


fat-fred:
Disabled patch for now. patch are right, but give an error on compile..
I do not now, if we need this patch, but will have a look in it (later).

jkellerer:
Hi @fat-fred, I was about to create a similar PR.

An extra include must be added to get the patch compiled. 
The definition of `ff_side_data_set_encoder_stats(...)` was moved to `packet_internal.h`:
```diff
@@ -26,6 +26,8 @@
 
 #include "vaapi_encode.h"
 #include "avcodec.h"
+#include "internal.h"
+#include "packet_internal.h"
 
 const AVCodecHWConfigInternal *ff_vaapi_encode_hw_configs[] = {
```

I'm not familiar with this patch either but it should be checked why it is required (or not) before disabling it, I think.

fat-fred:
@jkellerer thank you. it works.
I see it like you. Before we not now why this patch is included, we should not remove it (any of it).

fat-fred:
Here it works. Can someone else confirm that its compile (or not)?

jkellerer:
It also works for me with the adjusted patch. Can compile (including vaapi) and transcode with HW support. Everything else works also.

CvH:
@jkellerer `Can compile (including vaapi) and transcode with HW support.` can you try compile without the patch ? that patch is not required for vaapi encoding so nobody really knows what this patch does at all. 
If a patch is not required and nobody knows for what that patch is doing at all the patch should be dropped instead of forward ported without any testing.

fat-fred:
i looked into it. All other encoders have this patch integrated. Only vaapi have it not.
and i found here the patch: https://patchwork.ffmpeg.org/project/ffmpeg/patch/1595763675-1857-5-git-send-email-mypopydev@gmail.com/
it is from July
Perhaps it will any time integrated - but until then we should keep it.

Flole998:
Totally agree that it should stay in. If you figure out what it's for, why it was needed and why it's no longer needed you can propose a removal, until then it will stay in. If we start removing or changing code without understanding it sooner or later something important might be affected.

Please squash your commits together, then I can merge them :) Thank you very much for your effort!

jkellerer:
Btw. if you look at what the patch does, it tells the caller what type of frame an encoded frame is. I'm not familiar with the streaming code in TVH, however when you stream in "http" you may end up with a full stream buffer when the client cannot keep up. In this case you must skip frames, but you can do this only with full frames (hence you need to know the frame type).

PR #1366, opened 2020-10-13
fix vaapi-profiles

fat-fred:
- remove "Baseline" h264 Profil (not exist)

- include "Main10" and "Rext" in hevc Profiles


Saentist:
@fat-fred 
Is it possible to add levels limitation, to improve STB's and some Android box support?

Flole998:
Is there any reason why it's commented out and not removed? If it's really non-existent it should be removed, there's no point in keeping it.

fat-fred:
@Flole998 yes. You are right. Will Remote it Complete. 

@Saentist will have a Look in it.

Flole998:
If you could squash your commits together I will merge it :) Thank you very much for your contribution.

PR #1369, opened 2020-10-27
Changed shebang of tvhmeta to python

fat-fred:
So it should support python 2 and 3


CvH:
does `tvhmeta` actually work with py3 ?

fat-fred:
In my System (focal/docker) it  works with python 3

PR #1373, opened 2020-11-03
Remove libva-x11 dependency

danielcbit:
This change removes the dependency for the libva-x11 from the tvheadend build.

The libva-x11 library can cause issue when running TvHeadend in a headless server and the installation of several unnecessary packages. Running headless is the major use case for TvHeadend.

### DIFF
--- a/configure
+++ b/configure
@@ -654,7 +654,6 @@ fi
 if enabled_or_auto vaapi; then
   if enabled libav; then
     if check_pkg libva ">=0.38.0"; then
-      check_pkg libva-x11 ">=0.38.0" || die "libva-x11 not found"
       check_pkg libva-drm ">=0.38.0" || die "libva-drm not found"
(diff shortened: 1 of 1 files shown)


Trujulu:
+1 libav-X11 is a pain

Flole998:
Does vaapi-hardware-acceleration still work with this removed? Any idea why it might have been added?

danielcbit:
I have been using TvHeadend without this dependency for a couple of years and decided to submit my patch now. intel_gpu_top shows usage just fine while transcoding a live streaming.

From my understanding, the libva-x11 is used for applications running on top of x11 so they can get access to VDPAU. I run tvheadend headless and installing this lib would require me to use the entire X.org as a dependency, which I wanted to avoid.

Please see the image attached:
 
Screen Shot 2020-11-22 at 11 17 48

PR #1375, opened 2020-11-14
Update STAT_ARG to use -c instead of --printf

hoshsadiq:
`--printf` only works with the stat provided by coreutils. `-c` is more portable and works with busybox. When running `make` in busybox, stat will fail with the error:

```
/bin/stat: unrecognized option: printf=%-35n %7s\n
BusyBox v1.31.1 () multi-call binary.

Usage: stat [OPTIONS] FILE...
```

### DIFF
--- a/Makefile.webui
+++ b/Makefile.webui
@@ -48,7 +48,7 @@ endif
 ifeq ($(PLATFORM), freebsd)
 STAT_ARG=-f "%-35N %7z"
 else
-STAT_ARG=--printf="%-35n %7s\n"
+STAT_ARG=-c "%-35n %7s\n"
 endif
 
 JAVASCRIPT =


Flole998:
Please read the man page and do some manual tests to understand the difference between these options.

hoshsadiq:
@Flole998 I'm a little confused, the only thing I missed was that `--printf` has a `\n` at the end. I did test it and it seemed to work fine (ran it etc and inputs, and streamed said inputs too). What else am I missing that warrants this to be closed instead of providing feedback so I can update the PR?

Flole998:
The -c option does not honor backslash escapes. So you are printing literally \n at the end and not a new line. That should be noticed during simple testing, so I had the impression that this is entirely untested. Feel free to open a new PR if you found a better way that offers an equivalent output (convert the output to hex and diff it to check that), there's no point in updating a single-line edit PR and then squashing it together, it's easier to just open a new one. But if you want to do edit it please let me know and I will reopen it.

PR #1376, opened 2020-11-27
docs: fix simple typo, seperately -> separately

timgates42:
There is a small typo in src/descrambler/capmt.c.

Should read `separately` rather than `seperately`.


Flole998:
Thanks!

PR #1378, opened 2020-12-09
Add ISDB-T support to HDHomeRun client

danielcbit:
Adding support to the HDHomeRun client for HDHomeRub device HDHR4-2IS which is compatible with ISDB-T standard used in Brazil and other countries.

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun.c
@@ -123,6 +123,7 @@ tvhdhomerun_device_class_override_enum( void * p, const char *lang )
   htsmsg_add_str(m, NULL, "ATSC-T");
   htsmsg_add_str(m, NULL, "ATSC-C");
   htsmsg_add_str(m, NULL, "CableCARD");
+  htsmsg_add_str(m, NULL, "ISDB-T");
   return m;
 }
 
@@ -336,6 +337,8 @@ static void tvhdhomerun_device_create(struct hdhomerun_discover_device_t *dInfo)
       type = DVB_TYPE_CABLECARD;
     if (strstr(hd->hd_info.deviceModel, "_dvbt"))
       type = DVB_TYPE_T;
+    if (strstr(hd->hd_info.deviceModel, "_isdbt"))
+      type = DVB_TYPE_ISDB_T;
   }
 
   hd->hd_override_type = strdup(dvb_type2str(type));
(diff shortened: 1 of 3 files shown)


src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
```diff
@@ -454,6 +454,9 @@ static int tvhdhomerun_frontend_tune(tvhdhomerun_frontend_t *hfe, mpegts_mux_ins
           break;
       }
       break;
+    case DVB_TYPE_ISDB_T:
+      snprintf(channel_buf, sizeof(channel_buf), "auto:%u", dmc->dmc_fe_freq);
```
Flole998, line 458: Am I missing something or is this the default from below? There's no point in making it redundant, so this should be removed.

Flole998 (changes requested):
Just a small issue that needs clarification before this can be merged

src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
danielcbit: I left a separated case for ISDB-T as I thought there might be some additional things specific for ISDB-T but you right and we can use the default case here. Will remove it.

danielcbit (commented):
Updating review to remove unnecessary case in tvhdhomerun_frontend.c

Leatherface75:
Maybe it looks better if you change "isdbt_class" to "isdb_t_class" like the others?

src/input/mpegts/tvhdhomerun/tvhdhomerun.c
```diff
@@ -336,6 +337,8 @@ static void tvhdhomerun_device_create(struct hdhomerun_discover_device_t *dInfo)
       type = DVB_TYPE_CABLECARD;
     if (strstr(hd->hd_info.deviceModel, "_dvbt"))
       type = DVB_TYPE_T;
+   if (strstr(hd->hd_info.deviceModel, "_isdbt"))
```
Flole998, line 340: You're missing a space here ;)

Flole998:
Once the missing space is added this can be merged. Sorry, I would have added it myself but I'm not at a computer at the moment.

Flole998 (approved):
Thanks!

src/input/mpegts/tvhdhomerun/tvhdhomerun.c
danielcbit: Good catch. Updated the code. Thanks!

PR #1379, opened 2020-12-16
Fix possible deadlock when using tvh_mutex_trylock()

codesnake:
Fixes possible deadlock when using tvh_mutex_trylock() macro in thread non-debug mode.
The macro expands to call pthread_mutex_lock() instead of pthread_mutex_trylock(),
which most likely is a result of copy/paste.


Flole998:
Thank you very much for your contribution!

PR #1380, opened 2020-12-20
Upgrade to libhdhomerun_20200907

ekuester:
There is a new version of libhdhomerun on the home page (also available under Fedora 32, 33)

### DIFF
--- a/Makefile.hdhomerun
+++ b/Makefile.hdhomerun
@@ -32,10 +32,10 @@ endif
 # Upstream Packages
 # ###########################################################################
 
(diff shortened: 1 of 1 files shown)


Trujulu:
Is it possible to add DVB-C both with DVB-T support for DT models? 

As seen on hdhomerun last fw these devices also support DVB-C now: 
- HDHR5-2DT/4DT: add DVB-C support.

ekuester:
@Trujulu Sorry, out of my knowlegde ... I am using tvheadend only for decoding DVB-T2 public service broadcasting here in Germany.

PR #1383, opened 2021-02-18
Fix several issues discovered by coverity

Flole998:


src/satip/rtsp.c
```diff
@@ -673,6 +673,7 @@ rtsp_start
     }
     if (rs->mux == mux && rs->subs) {
       if (rs->no_data) {
+       dvb_mux_conf_str(&rs->dmc, buf, sizeof(buf));
```
ksooo, line 676: wrong indentation?!

Flole998: Fixed, thanks!

PR #1391, opened 2021-03-01
Clean transcoding when htsp based transcoding is used

Flole998:
@perexg There was an issue when HTSP clients like kodi used a transcoding profile (especially nvidia based profiles) so the transcoder was never cleaned up which caused more and more GPU memory to be used. This fixes it for me but I have absolutely no idea if this is a proper fix or if `run` should evaluate to true in this case and the cleanup code from above should be used instead. There is no profile sharer thread started in this case if I saw that correctly, so it should be correct that run evaluates to 0 but it could be that a profile sharer thread should be started in this case. Really not sure about this and I would appreciate if you could look over this and approve this or reject this is this is not a proper fix (and I believe that this is not a proper fix).

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -954,27 +954,22 @@ profile_sharer_destroy(profile_chain_t *prch)
 {
   profile_sharer_t *prsh = prch->prch_sharer;
   profile_sharer_message_t *psm, *psm2;
-  int run = 0;
 
   if (prsh == NULL)
     return;
-  tvh_mutex_lock(&prsh->prsh_queue_mutex);
   LIST_REMOVE(prch, prch_sharer_link);
   if (LIST_EMPTY(&prsh->prsh_chains)) {
-    if ((run = prsh->prsh_queue_run) != 0) {
+    if (prsh->prsh_queue_run) {
+      tvh_mutex_lock(&prsh->prsh_queue_mutex);
       prsh->prsh_queue_run = 0;
       tvh_cond_signal(&prsh->prsh_queue_cond, 0);
-    }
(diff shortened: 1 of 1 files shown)


pinkflozd:
Helps with growing memory on vaapi too, but tvheadend segfaults when a second user that watches the same channel stops playback

Flole998:
Then this is not the correct implementation as I already assumed....

Saentist:
May be, because no one monitor session usage. 
Session process must be keep alive until there is active client/s/ exist, 
and terminated 2-5 sec after last leave.

Flole998:
A fix was provided at the Forum and I have updated this PR now to use that fix. Please check if this fix is better.

pinkflozd:
Confirming it fixed, no memory buildup with htsp+vaapi

susadmin:
I would just like to add +1 vote for adding this fix to the master branch as this (expanding GPU memory usage) is still happening for me. I have to use the fix-htsp-transcode branch until it is included.

Flole998:
This fix is not correct yet though. There's a potential race condition in it, so that needs fixing first.

susadmin:
Ok thanks for the update and for all your hard work on this project.

woooow1:
@Flole998  changes that you made, works without any problems. If you decide, close this PR!

Flole998:
What changes? I'm not aware of fixing this yet (or I forgot).

woooow1:
https://github.com/tvheadend/tvheadend/commit/fc3759a58dd9dc914166262c8b59c2d4f0ed3f53

PR #1393, opened 2021-03-15
remote timeshift: fix crash on multiple subscriptions and cleanup

spdfrk:
In case multiple users subscribe to a remote time shift channel the different code path would not initialize the service_t * sm_s properly and caused a segfault due to invalid pointer later in the code. Also I was able to revert the change from malloc to calloc, we need speed!
Furthermore the code was cleaned up slightly.

A limitation still is that all subscribers will look at the same source stream so remote time shifting will apply to all. A solution would be to spawn multiple instances of the same mux / service. Not sure how to handle this in tvheadend, it doesn't seem to be designed for this. Ideas are welcome!

### DIFF
--- a/src/input/mpegts/iptv/iptv_private.h
+++ b/src/input/mpegts/iptv/iptv_private.h
@@ -124,6 +124,7 @@ typedef struct {
   /* Connection to the RTCP remote */
   udp_connection_t *connection;
   int connection_fd;
+  udp_multirecv_t um;
 
   uint32_t source_ssrc;
   uint32_t my_ssrc;
@@ -172,6 +173,7 @@ struct iptv_mux
   uint32_t              mm_iptv_rtp_seq;
 
   sbuf_t                mm_iptv_buffer;
+  sbuf_t                im_temp_buffer;
 
   uint32_t              mm_iptv_buffer_limit;
 
@@ -189,11 +191,11 @@ struct iptv_mux
 
   void                 *im_opaque;
 
-  udp_multirecv_t      im_um1;
-  udp_multirecv_t      im_um2;
+  udp_multirecv_t      im_um;
+
   char                 im_use_retransmission;
-  sbuf_t               im_temp_buffer;
   char                 im_is_ce_detected;
+
   rtcp_t               im_rtcp_info;
 };
 
--- a/src/input/mpegts/iptv/iptv_rtcp.c
+++ b/src/input/mpegts/iptv/iptv_rtcp.c
@@ -371,8 +371,8 @@ rtcp_send_nak(rtcp_t *rtcp_info, uint32_t ssrc, uint16_t seqn, uint16_t len)
           ssrc, n, (uint32_t )sizeof(network_buffer));
     }
 
-  // Cleanup
-  sbuf_free(&network_buffer);
+    // Cleanup
+    sbuf_free(&network_buffer);
   }
   return 0;
 }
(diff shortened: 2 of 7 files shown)


Flole998:
> A limitation still is that all subscribers will look at the same source stream so remote time shifting will apply to all. A solution would be to spawn multiple instances of the same mux / service. Not sure how to handle this in tvheadend, it doesn't seem to be designed for this. Ideas are welcome!

I guess that would be very difficult to do as you would need a second stream to start once the first one pauses/timeshifts.

I am note sure if this is a good idea or if we should disallow remote timeshift once there is more than one subscription. This could break recordings for example. Maybe it would be a good idea to stop timeshifting once a recording starts. Or the option to use remote timeshift implies that a single client uses a single connection, so one client would equal one connection to the source and Tvheadend just proxies the connection.

Anyways: Merging this now to prevent the crash.

Trujulu:
Could you check this exception? @spdfrk 

In file included from /project/repo/checkout/src/tvheadend.h:39,
                 from src/input/mpegts/iptv/iptv_rtsp.c:20:
src/input/mpegts/iptv/iptv_rtsp.c: In function 'rtsp_input':
src/input/mpegts/iptv/iptv_rtsp.c:497:22: error: format '%ld' expects argument of type 'long int', but argument 7 has type 'int64_t' {aka 'long long int'} [-Werror=format=]
     tvhinfo(LS_IPTV, "rtsp: skip: %" PRItime_t " + %" PRItime_t, rp->range_start,
                      ^~~~~~~~~~~~~~~
         ts_rescale(data->time, 1));
         ~~~~~~~~~~~~~~~~~~~~~~~~~
/project/repo/checkout/src/tvhlog.h:206:66: note: in definition of macro 'tvhlog'
   _tvhlog(__FILE__, __LINE__, severity | LOG_TVH_NOTIFY, subsys, fmt, ##__VA_ARGS__)
                                                                  ^~~
src/input/mpegts/iptv/iptv_rtsp.c:497:5: note: in expansion of macro 'tvhinfo'
     tvhinfo(LS_IPTV, "rtsp: skip: %" PRItime_t " + %" PRItime_t, rp->range_start,
     ^~~~~~~
In file included from src/input/mpegts/iptv/iptv_rtsp.c:20:
/project/repo/checkout/src/tvheadend.h:338:27: note: format string is defined here
 #define PRItime_t       "ld"
cc1: all warnings being treated as errors

spdfrk:
@Trujulu, can you tell me what kind of OS / libraries you are using? It looks like the PRItime_t macro in Tvheadend is not compatible with your setup. 
I compiled this with success using defaults in Ubuntu 18 & 20.

Trujulu:
> @Trujulu, can you tell me what kind of OS / libraries you are using? It looks like the PRItime_t macro in Tvheadend is not compatible with your setup.
> I compiled this with success using defaults in Ubuntu 18 & 20.

You can see those errors on doozer. Aparently before your changes everything compiles.

Flole998:
The problem is that ts_rescale returns an int64_t, which is not PRItime_t but "long long int". I have already seen that and will fix it later.

spdfrk:
Thanks! Was looking at the wrong argument. This error didn't show up on my side. Will update my setup accordingly.

Flole998:
@spdfrk Apparently there's an issue with this code described at https://tvheadend.org/d/8812-crash-tvhtcp-startx-segfault-at-x-ip-x-sp-x-error-4-in-tvheadend/39

Could you please have a look if you find some time?

PR #1398, opened 2021-04-20
Introduce ROTOR access control via subscription priority.

smithereens:
Only subscriptions with this priority and above are allowed to control/move ROTOR
For example, set to 5 to prevent OTA EPG and other periodic scan from moving rotor.

### DIFF
--- a/src/input/mpegts/linuxdvb/linuxdvb_private.h
+++ b/src/input/mpegts/linuxdvb/linuxdvb_private.h
@@ -268,6 +268,7 @@ struct linuxdvb_satconf
   int                    ls_site_lat_south;
   int                    ls_site_lon_west;
   int                    ls_site_altitude;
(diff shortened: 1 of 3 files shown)


smithereens:
This pull request is only of use for people using motorised dish who want to avail of nightly OTA EPG and MUX scan but don't want the ROTOR to move to all satellite positions when doing do. This is main use case for me, but there likely other scenario where controlling rotor access by subscription priority could be useful.

by setting the ROTOR min priority just above the OTA EPG/MUX subscription, the rotor will refuse to move position when requested so by the automatic mux scan, and OTA EPG will only succeed on whatever satellite it is currently pointing at. A tuning error will be returned for the other conditions.

subscription failures, like in the case of tuning error, is actually poorly handled by OTA EPG mux code, so I've created a separate pull request #1399  to handle this particular problem.

PR #1400, opened 2021-04-21
Added Rewrite MPEG-TS SI settings to the MPEG-TS SPAWN profile panel.

Jumper78:
Update profile.c
Added Rewrite MPEG-TS SI settings to the MPEG-TS SPAWN profile panel.

Currently, only the MPEG-TS profile has the expert options for MPEG-TS SI settings. Unfortunately, I need these settings for the TV stations I am receiving. Up to now, I was not able to use MPEG-TS SPAWN for recoding via ffmpeg. After adding the MPEG-TS SI Table(s) Settings panel to the expert options of MPEG-TS SPAWN and deselecting some of the rewrite options, that is now possible.

PR #1402, opened 2021-04-25
Test/nis rewrite with korean tv streams

Jumper78:
Some Korean terrestrial TV streams are broadcasting the Program Map Table (PMT) and the video and the audio streams on PIDs where usually SDT, NIT or EIT should be. For the video and audio stream, `muxer_pass.c` did already take provisions and turns of SDT, NIT and EIT rewriting. For the PMT, up to now, no provisions were taken.
This patch checks if the PMT PID conflicts with the PIDs usually used for SDT, NIT or EIT. If that is the case, it disables PMT rewriting as well as the SDT, NIT or EIT rewriting.
The pull request also includes the changes from #1400.


Jumper78:
need to squash the changes

PR #1407, opened 2021-05-04
Change format string for pid, gid, uid from d to u

hochwasser:
- glist is of type gid_t aka unsigned int. Changed format string from d to u.
- gid is of type gid_t aka unsigned int. Changed format string from d to u.
- uid is of type uid_t aka unsigned int. Changed format string from d to u.
- pid is of type pid_t aka unsigned int. Changed format string from d to u.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1141,7 +1141,7 @@ main(int argc, char **argv)
           int i;
           for (i = 0; i < gnum; i++)
             snprintf(buf + strlen(buf), sizeof(buf) - 1 - strlen(buf),
-                     ",%d", glist[i]);
+                     ",%u", glist[i]);
           tvhlog(LOG_ALERT, LS_START,
                  "setgroups(%s) failed, do you have permission?", buf+1);
           return 1;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1349,7 +1349,7 @@ main(int argc, char **argv)
   pthread_sigmask(SIG_UNBLOCK, &set, NULL);
 
   tvhlog(LOG_NOTICE, LS_START, "HTS Tvheadend version %s started, "
-         "running as PID:%d UID:%d GID:%d, CWD:%s CNF:%s",
+         "running as PID:%u UID:%u GID:%u, CWD:%s CNF:%s",
```
Flole998, line 1352: According to https://www.gnu.org/software/libc/manual/html_node/Process-Identification.html `getpid()` returns an `int` and not an `unsigned int`?

hochwasser:
I changed it back to d for the pid.
As far as I can tell: POSIX does not define a type. gnu libc uses a signed int. Thanks for this finding.

Flole998:
Could you please squash your commits together?

oliv3r:
@Flole998 I think you might force a squash when merging, there should/could be an option/dropdown on the merge button iirc.

Flole998:
That option does exist, but in this repo the convention is to rebase-and-merge. This should be rebased before merging anyways to benefit from the new CI checks.

oliv3r:
Ah, there's no 'rebase' button.

<broken-record>With gitlab, there's a rebase/sync button, and in combination with the squash, you can do this as maintainer easily :)

That said, I fully agree with the process, but if the UI offers this, and you end up with the same single commit with the same commit message, it makes life easier :) YMMV

PR #1409, opened 2021-05-04
correct format string for weight from d to u

hochwasser:
weight is of type unsigned int so correct the format string from d to u

### DIFF
--- a/src/subscriptions.c
+++ b/src/subscriptions.c
@@ -857,10 +857,10 @@ subscription_create_from_channel_or_service(profile_chain_t *prch,
   if (tvhtrace_enabled()) {
     const char *pro_name = prch->prch_pro ? profile_get_name(prch->prch_pro) : "<none>";
     if (ch)
-      tvhtrace(LS_SUBSCRIPTION, "%04X: creating subscription for %s weight %d using profile %s",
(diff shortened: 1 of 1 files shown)


Flole998:
If you're planning on fixing more formatting errors then we should collect those in a single PR. I'm sure there are more mistakes like these in the code.

hochwasser:
I can collect others format string mistakes, if you like in this PR.
I would like 
- replace all malloc, calloc and realloc with xmalloc, xcalloc and xrealloc to mitigate NULL pointer problems.
- introduce a strequal function to get ride of the strcmp which misses the == 0
This needs discussion and I did not find where to ask. Here is no bug tracker.

Flole998:
As `xmalloc` is a non-standard function, so that's not a good idea. At some places we also handle those errors instead of crashing. If it fails and the error is not handled you can be pretty sure that there's a crash soon after that anyways.

We could definitely introduce a strequal function (or probably do it as define so it gets replaced during preprocessing and we don't rely on compiler optimization). That requires a huge amount of work though and I'm not sure if it's worth the effort.

There's a bug tracker at https://tvheadend.org

hochwasser:
It was a lot of work, but I think I got every case.
Could you please review this PR?

PR #1411, opened 2021-05-26
Check the return value of malloc, calloc and realloc

hochwasser:
malloc and calloc allocate memory and return NULL if this failes.
The return value is now checked and if there is a check for NULL a thvinfo is
added. If there was no NULL check a thvabort is added.
realloc is special that if it returns NULL the old pointer is still valid.
if the input pointer and the return  pointer is the same variable the old
pointer is lost. This is a memory leak.
Now there is also a thvinfo or thvabort added.

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -26,6 +26,7 @@
 #include "dvr/dvr.h"
 #include "tcp.h"
 #include "lang_codes.h"
+#include "tvhlog.h"
 
 struct access_entry_queue access_entries;
 struct access_ticket_queue access_tickets;
@@ -129,6 +130,9 @@ access_ticket_create(const char *resource, access_t *a)
   }
 
   at = calloc(1, sizeof(access_ticket_t));
+  if (at == NULL) {
+    tvhabort(LS_ACCESS, "calloc is NULL");
+  }
 
   uuid_random(buf, sizeof(buf));
   bin2hex(id, sizeof(id), buf, sizeof(buf));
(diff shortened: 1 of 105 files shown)


Flole998:
I've left a few comments. Also I think it would be better to write "xy returned NULL" instead of "is NULL".

perexg:
If there's a memory allocation error, we're in a really deep problem. The system generates SIGSEGV when the NULL pointer area is accessed, so we have already a sign that something is bad.

If you like to clean up this code, please, define own allocation wrappers or macros which calls tvhabort() or tvhwarn() if the allocation is not successful. The macros can dump the origin - the source file / line number - automatically.

Flole998:
Thats an even better idea! I think the alloca replacements are fine though?

hochwasser:
@perexg this sounds like xmalloc, xcalloc and xrealloc.
If I copy the interface and use the tvh logging functions, would this ok?
@Flole998 this was the first try and only cover the easy convertible alloca.

hochwasser:
Eli Bendersky has written a good article about malloc an out-of-memory conditions:
https://eli.thegreenplace.net/2009/10/30/handling-out-of-memory-conditions-in-c/

glib from gnome is a a good solution as libiberty,
What are your thoughts?

src/access.c
```diff
@@ -129,6 +130,9 @@ access_ticket_create(const char *resource, access_t *a)
   }
 
   at = calloc(1, sizeof(access_ticket_t));
+  if (at == NULL) {
```
oliv3r, line 133: from what I saw, hts doesn't generally use curly brackets for single statements

src/access.c
```diff
@@ -736,7 +746,7 @@ access_get(struct sockaddr_storage *src, const char *username, verify_callback_t
                        superuser_username, superuser_password))
       return access_full(a);
   } else {
-    s = alloca(50);
+    char s[50];
```
oliv3r, line 749: doesn't seem to be related to this patchset? though (i know below its hardcoded as well) should really be a define/sizeof combo :)

PR #1412, opened 2021-06-12
As we no longer have access to #hts on freenode swap to using libera 

mpmc:
See https://tvheadend.org/issues/6054


Saentist:
+1

PR #1413, opened 2021-07-18
Remove EOL/unmaintained build targets & add bullseye

mpmc:
I've removed EOL distros and those that are no longer maintained by anyone. Fedora, Centos etc used to be under @perexg's eye however it sadly appears he's no longer able to maintain them? 

Please don't commit without discussing these changes first. Thanks.


Flole998:
I think everything that still builds should stay in until it breaks, if it's an old distro and no easy fix it can be kicked out then.

Saentist:
Ubuntu 14.04 aka Trusty
cannot be build after latest ffmpeg update

There is no code tags for that

mpmc:
> 
> 
> I think everything that still builds should stay in until it breaks, if it's an old distro and no easy fix it can be kicked out then.

In that case you can just cherry-pick the bullseye commits :)

mpmc:
OK, i'll close this and rework it :)

PR #1419, opened 2021-09-09
Improve the performance of updating the pid filter table in hdhomerun…

NormRaden:
… digital tuners.

 - Construct the list of pids formatted for the hdhomerun and handle edge cases.
 - For each call to tvhdhomerun_frontend_update_pids(...), only issue one set_tuner_filter call to the hdhomerun.
 - Turn off pids when done using them.
 - Handle the rare case where the requested list of pids does not fit in to the fixed length buffer by enabling all pids (except the NUL pid)

 Previously, calls to update the pids would generate get_tuner_filter and set_tuner_filter calls for each individual pid.
 Also, while a tuner was allocated, pids would never be freed up from the hdhomerun filter table even when they were no longer needed.
 And if the list of requested pids had exceeded the fixed length buffer, then not all the requested pids would be properly enabled.

 Very noticeably, in cases where the number of pids is large enough, very frequent pid updates would overload the hdhomerun causing it to
 drop bursts of program data (mpegts broadcast packets) every several seconds on all tuners for the duration of the rapid pid updates
 (reproducible on the HDHR4-2US model).

----

This proposed patch reimplements tvhdhomerun_frontend_update_pids(...) to read the list of pids and generate a single string to set the pid filter in the hdhomerun.

tvhdhomerun_frontend_update_pids() repeatedly, for each pid, reads the hdhomerun pid filter string, appends the single pid, and writes this pid filter string to set the pid filter.

Currently, tvhdhomerun_frontend_update_pids() does this:

2021-08-27 11:32:25.036 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.041 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.041 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000 0x0030-0x0031 0x0034-0x0035 0x0001
2021-08-27 11:32:25.047 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.047 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x0030
2021-08-27 11:32:25.054 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.054 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x0031
2021-08-27 11:32:25.060 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.060 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x0034
2021-08-27 11:32:25.067 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.067 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x0035
2021-08-27 11:32:25.070 [  DEBUG]:tvhdhomerun: current pfilter: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035
2021-08-27 11:32:25.070 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x1ffb

The reimplemented tvhdhomerun_frontend_update_pids() only does this:

2021-09-01 14:11:30.610 [  DEBUG]:tvhdhomerun: setting pfilter to: 0x0000-0x0001 0x0030-0x0031 0x0034-0x0035 0x1ffb

The control traffic to the hdhomerun grows exponentially with the number of pids.  The reimplementation brings this down to linear growth.

One of the EPG modules calls tvhdhomerun_frontend_update_pids() rapidly and if there is also a large enough set of pids to update, the flood of control traffic causes the hdhomerun (verified on model HDHR4-2US) to malfunction and intermittently drop broadcast packets.

Also, the scratch buffer for assembling this list of hdhomerun formatted pids is currently hard coded to 1024 bytes/chars for both implementations.  If we can't fit the list of pids in to the buffer, then we fallback and send "0x0000-0x1ffe" to enable all but pid 0x1fff (NUL packets), instead of taking our chances on sending a truncated list of pids.

Code Notes:

  - I thought this might be a little tricky to understand so I supplied some extra comments in tvhdhomerun_frontend_update_pids().
  - I used a local function in a code block in tvhdhomerun_frontend_update_pids() to simplify the implementation.

Verification:

  - I tested different combinations of enabling/disabling programs on a tuner and verified that the appropriate pids are set.
  - I verified that the correct pids were unset when a program ended while another program continued on the same tuner.
  - I test recompiled this fix with the scratch buffer of only 64 chars long (ie. char buffer[64];) to easily test the case where this buffer is too small to handle some list of pids.
  - I tested the turn-on-all-pids/MPEGTS_FULLMUX_PID case by intermittently requesting a mux. stream, while programs were present and absent on the same tuner.
  - I've been testing this patch in production for more than a week and counting--so far so good!

If you have a hdhomerun digital tuner, have a look.  Don't forget to add/append 'tvhdhomerun' to --debug or via UI under Configuration-->Debugging-->Debug subsystems and watch for the string "pfilter to" in the message log.  Also, take notice of the message log for string matches of "pfilter to" both before and after applying this patch to compare.

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
@@ -328,61 +328,6 @@ tvhdhomerun_frontend_monitor_cb( void *aux )
   }
 }
 
-static void tvhdhomerun_device_open_pid(tvhdhomerun_frontend_t *hfe, int pid) {
-  char *pfilter;
-  char buf[1024];
-  int res;
-
-  if (hfe->hf_type == DVB_TYPE_CABLECARD)
-    return;
-
-  /* a full mux subscription should specificly set the filter */
-  if (pid == MPEGTS_FULLMUX_PID) {
-    tvhdebug(LS_TVHDHOMERUN, "setting PID filter full mux");
-    tvh_mutex_lock(&hfe->hf_hdhomerun_device_mutex);
-    res = hdhomerun_device_set_tuner_filter(hfe->hf_hdhomerun_tuner, "0x0000-0x1FFF");
-    tvh_mutex_unlock(&hfe->hf_hdhomerun_device_mutex);
-    if(res < 1)
-      tvherror(LS_TVHDHOMERUN, "failed to set_tuner_filter to 0x0000 - 0x1FFF");
-    return;
-  }
-
-  /* get the current filter */
-  tvh_mutex_lock(&hfe->hf_hdhomerun_device_mutex);
-  res = hdhomerun_device_get_tuner_filter(hfe->hf_hdhomerun_tuner, &pfilter);
-  tvh_mutex_unlock(&hfe->hf_hdhomerun_device_mutex);
-  if(res < 1) {
-      tvherror(LS_TVHDHOMERUN, "failed to get_tuner_filter: %d", res);
-      return;
-  }
-
-  tvhdebug(LS_TVHDHOMERUN, "current pfilter: %s", pfilter);
-
-  /* make sure the pid maps to a max of 0x1FFF, API will reject the call otherwise */
-  if(pid > 0x1FFF) {
-    tvherror(LS_TVHDHOMERUN, "pid %d is too large, masking to API maximum of 0x1FFF", pid);
-    pid = (pid & 0x1FFF);
-  }
-
-  memset(buf, 0x00, sizeof(buf));
-  snprintf(buf, sizeof(buf), "0x%04x", pid);
-
-  if(strncmp(pfilter, buf, strlen(buf)) != 0) {
-    memset(buf, 0x00, sizeof(buf));
-    snprintf(buf, sizeof(buf), "%s 0x%04x", pfilter, pid);
-    tvhdebug(LS_TVHDHOMERUN, "setting pfilter to: %s", buf);
-
-    tvh_mutex_lock(&hfe->hf_hdhomerun_device_mutex);
-    res = hdhomerun_device_set_tuner_filter(hfe->hf_hdhomerun_tuner, buf);
-    tvh_mutex_unlock(&hfe->hf_hdhomerun_device_mutex);
-    if(res < 1)
-      tvherror(LS_TVHDHOMERUN, "failed to set_tuner_filter: %d", res);
-  } else {
-    //tvhdebug(LS_TVHDHOMERUN, "pid 0x%x already present in pfilter", pid);
-    return;
-  }
-}
-
 static int tvhdhomerun_frontend_tune(tvhdhomerun_frontend_t *hfe, mpegts_mux_instance_t *mmi)
 {
   hfe->hf_status          = SIGNAL_NONE;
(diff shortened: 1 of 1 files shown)


Trujulu:
I am using this patch on HDHR5-4DT(EU) and HDHR5-4DT(UK) and it is working correctly at the moment. 
It seems that there are fewer errors in the streams. If there is any problem I will continue to inform.

src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
```diff
@@ -564,10 +517,87 @@ static void tvhdhomerun_frontend_update_pids( mpegts_input_t *mi, mpegts_mux_t *
...
+
+      const char *endBuffer = buffer + bufferSize; /* Have the address after the end of buffer on hand to help avoid writing past the buffer. */
+      char *pBuffer = buffer; /* Move this pointer through the buffer as we write formatted pids. */
+      int firstDelimiter = -1; /* Set this to 'true' so that we can skip writing the first delimiter/space. */
+
+      void appendPidRange(int a, int b) /* A local function that writes a range of pids to the 'buffer'.  This function is called more than once. */
+      {
+        if(firstDelimiter) /* Don't bother printing a space before the first range of pids. */
+          firstDelimiter = 0;  /* Set this to false the first time. */
+        else
```
Flole998, line 547: Please also add parentheses here for better readability

Flole998 (changes requested):

Flole998:
@NormRaden Thank you very much for your contribution and you did a very good job with the description of this PR! I will let coverity run a quick check over this to see if there are any issues that I didn't notice during my review and then it can be merged from the functional side IMO. There are 2 cases where I'd like braces to be added though for better readability.

NormRaden:
@Trujulu Thanks for testing out this minor improvement!  I appreciate that you confirmed that it works with another hdhomerun model in some other regions.

@Flole998 I applied the requested changes and slipped in few minor edits to some comments.  I've retested after these changes and, as expected, it appears to work.  I may have gotten ahead of myself, but I squashed the changes in to one commit on the same branch.

So far, I've been running my production tvheadend service with this minor change for about 2 weeks without any problems.

Apologies to anyone affected by the forced push.

Trujulu:
Im getting '2021-09-23 13:14:10.590 tvhdhomerun: failed to reset pfilter: -1' on log after some days working. Some tuners not working...

`2021-09-23 13:17:32.995 [   INFO]:mpegts: 538MHz in DVB-T - tuning on HDHomeRun DVB-T Tuner #0 (192.168.0.154)
2021-09-23 13:17:32.995 [  DEBUG]:tvhdhomerun: HDHomeRun DVB-T Tuner #0 (192.168.0.154) - starting 538MHz in DVB-T
2021-09-23 13:17:32.995 [   INFO]:tvhdhomerun: tuning to t8qam64:538000000
2021-09-23 13:17:33.416 [  ERROR]:tvhdhomerun: failed to acquire lockkey: `�O��U
2021-09-23 13:17:33.836 [  ERROR]:tvhdhomerun: failed to reset pfilter: -1
`

Could you check this?

Trujulu:
Sorry for my previous comment, there was a problem with an homerun.

NormRaden:
@Trujulu Thanks for continuing to test this patch!

@Flole998 Continues to work for me too.  There is just one commit ready for consideration.

Trujulu:
Everything ok at my side.

NormRaden:
@Flole998 It's been a couple months, what else needs to be done to this patch?

Flole998:
I need to find time to let coverity do a quick check, if that's okay it can be merged I think.

NormRaden:
No hurry, I've applied gcc's new static analyzer on src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c and it only complains about a possible memory leak in function tvhdhomerun_frontend_monitor_cb(...) around the use of strdup(...) and a possible NULL dereference in function tvhdhomerun_frontend_create(...)--probable false positives, but it doesn't complain about any code changes relating to this PR.

PR #1426, opened 2021-11-30
Update Makefile.ffmpeg

fat-fred:


Flole998:
Was this properly tested against regressions introduced in the new ffmpeg version?

fat-fred:
Not really. Cant't see any Problems (compiling/transcoding works with my config). What regressions do you mean? will have a look into it.

PR #1427, opened 2021-12-01
some changes to nvenc

fat-fred:
-cleanup of profiles

  -include level selection for nvenc


Flole998:
I assume this doesn't depend on your other PR? Then I would go ahead and merge this, otherwise this would have to be put on hold while the other one isn't merged yet

fat-fred:
Yes, it doesn't depend on the other PR. Made it last year after the other PR but had a little problem and no time to fix until now.

PR #1436, opened 2022-01-28
fix binary msg

mwalliczek:
Currently retrieving the epg_api (epg/events/grid) via binary api access leads to a crash, because the type HMF_UUID (which is used for the channelUuid) is not part of the switch and leads to an abort(). To fix this I added this type to both switch cases (for the length and the actual data).


Flole998:
Please squash your commits together. Thanks!

Flole998:
Unfortunately you still have 4 commits in this PR, please make sure it's a single one with a meaningful description.

PR #1440, opened 2022-02-12
Doozer: build for Fedora 34 and 35

rubdos:
Hi!

I'm going to attempt and get tvheadend on Doozer to build against the newer Fedora versions. As I'm currently upgrading my system from Fedora 33 to 34, I'm not online on IRC, but you can usually ping me (rubdos) to take a further look.

In this first patch, I'm only changing out the version numbers. I don't expect anything weird to happen, but if it stops building from there on, you can expect further PRs.

I tried signing the CLA, but the page at https://www.clahub.com/agreements/tvheadend/tvheadend is blank on my browser.

Best,

Ruben

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -582,10 +582,10 @@
         "support/bintray.py publish filelist.txt"
       ]
     },
-    "fedora28-x86_64": {
-      "buildenv": "docker:fedora:28",
+    "fedora34-x86_64": {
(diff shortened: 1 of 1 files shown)


Flole998:
Don't worry about the CLA.

If this doesn't fix the build you probably need to add `--nowerror` to the configure call. Assuming that issue with the defines for nasm isn't at fault for the errors.

PR #1442, opened 2022-02-15
doozer: add python-is-python3 to various dists

mpmc:
as title

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -492,7 +492,8 @@
         "python3",
         "python3-requests",
         "debhelper",
-        "ccache"
+        "ccache",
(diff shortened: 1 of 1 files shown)


Flole998:
What exactly is the advantage of doing this? All builds should be successful right now (there have been issues that should disappear on a rebuild).

mpmc:
@Flole998 Some of the py scripts used python and not python3 when I originally did this, they've since been changed to python3 adding the python-is-python3 avoided changing those & breaking something else.

Yes, I kept an eye on the builds. I also had to tweak the doozer settings and clear the cache to get them to build correctly.

- Will close :)

PR #1443, opened 2022-02-17
fixes issue #5930 - user config doesn't force DVR profile

dreamcat4:
Hello @Flole998,
Today I have tested and verified myself (independently) that this simple little patch here provided by [Dietmar Konermann](https://tvheadend.org/users/3426)

https://tvheadend.org/issues/5930#note-5

It worked, after recompile the bug of #5930 on current master branch. No other code changes necessary. And now the user's custom dvr profile is being respected. As it should be.

I do not believe this bug ever occured to my memory on the last previous stable version. Which was what I upgraded from directly. Upon finding this issue on master. Therefore it does not need to be backported to any previous older branches to my knowledge.

The code seems a small and simple change. Isolated from other things. So please look through Dietmar's patch here and check it for any other type of error. In the meantime, will keep running this code every daily myself.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -1039,8 +1039,10 @@ dvr_entry_create(const char *uuid, htsmsg_t *conf, int clone)
   dvr_entry_t *de, *de2;
   int64_t start, stop, create, watched, now;
   htsmsg_t *m;
+  htsmsg_field_t *f;
   char ubuf[UUID_HEX_SIZE], ubuf2[UUID_HEX_SIZE];
   const char *s;
+  access_t *a;
 
   if (conf) {
     if (htsmsg_get_s64(conf, "start", &start))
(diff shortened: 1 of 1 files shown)


ovaron76github:
For me it looks good.

Flole998:
I'd like to figure out why it apparently works in 4.2 and is broken in 4.3.... This code doesn't seem to be there in 4.2 but still it works, and I haven't figured out why it doesn't work in 4.3 aswell. I haven't had time to figure that out, once I'll get a chance to do that and have verified that this is indeed a proper fix (and not just ignoring the old, broken solution and adding a second one) I'll merge it. So if someone comes up with an explanation for 4.2 that'd accelerate it quite a bit I guess....

Wheemer:
Really looking forward to this fix.

Flole998:
Has anybody figured out why this works on 4.2 apparently without this fix?

Flole998:
Superseded by #1856

PR #1445, opened 2022-02-23
Add option to drop src-parameter from RTSP-requests

WereCatf:
This patch adds an option similar to the one for dropping
frontend-parameter from RTSP-requests, but this only applies
to DVB-C tuners. The reason for this patch is that it enables
Plex to work directly with the Tvheadend SAT>IP-server.

Signed-off-by: Nita Vesa <werecatf@outlook.com>

### DIFF
--- a/src/satip/rtsp.c
+++ b/src/satip/rtsp.c
@@ -1221,6 +1221,8 @@ rtsp_parse_cmd
 
   } else if (msys == DVB_SYS_DVBC_ANNEX_A || msys == DVB_SYS_DVBC_ANNEX_C) {
 
+    if(satip_server_conf.satip_drop_src)
+      http_arg_get_remove(&hc->hc_req_args, "src");
     freq *= 1000;
     if (freq < 0) goto end;
     c2tft = atoi(http_arg_get_remove(&hc->hc_req_args, "c2tft") ?: "0");
--- a/src/satip/server.c
+++ b/src/satip/server.c
@@ -970,6 +970,17 @@ const idclass_t satip_server_class = {
       .opts   = PO_EXPERT,
       .group  = 5,
     },
+    {
+      .type   = PT_BOOL,
+      .id     = "satip_drop_src",
+      .name   = N_("Drop \"src=\" parameter"),
+      .desc   = N_("Discard the source parameter in RTSP requests "
+                   "for DVB-C tuners, as some clients (e.g. Plex) "
+                   "incorrectly use it."),
+      .off    = offsetof(struct satip_server_conf, satip_drop_src),
+      .opts   = PO_EXPERT,
+      .group  = 5,
+    },
     {}
   },
 };
--- a/src/satip/server.h
+++ b/src/satip/server.h
@@ -52,6 +52,7 @@ struct satip_server_conf {
   int satip_anonymize;
   int satip_noupnp;
   int satip_drop_fe;
+  int satip_drop_src;
   int satip_restrict_pids_all;
   int satip_iptv_sig_level;
   int satip_force_sig_level;


Flole998:
What exactly is Plex specifying as src Parameter and why is this not working properly at the moment?

WereCatf:
> What exactly is Plex specifying as src Parameter and why is this not working properly at the moment?

Plex supplies a "src=1" and Tvheadend complains about extra parameters and rejects the request.

Flole998:
I guess the parameter should be properly parsed like it is done for DVB-S. That way it can be used properly with the "SAT>IP Source number" which can be specified in the network settings.

WereCatf:
> I guess the parameter should be properly parsed like it is done for DVB-S. That way it can be used properly with the "SAT>IP Source number" which can be specified in the network settings.

Alas, according to my research, apparently SAT>IP DVB-C tuners are only ever supposed to have a single network and you're technically not supposed to supply that parameter at all. Yes, it could be parsed, but properly compliant SAT>IP-clients would only ever supply src=1 or nothing at all.

WereCatf:
Besides the above, my patch would at least allow Plex (and any other misbehaving clients) to work with Tvheadend until someone bothers to implement the parsing for the source.

Flole998:
Where did you find a specification for SAT>IP specifying DVB-C?

I don't see any reason why there shouldn't be multiple cable networks connected to a single server. Just like multiple satellites can exist for a DVB-S server.

A patch that is just a workaround and adds lots of unnecessary code when really a single line (or 2 lines) needs to be moved up a little won't be merged.

WereCatf:
Lots of unnecessary code? Mm-hmm. Fine.

Flole998:
Yes, 15 lines added while all that's necessary is really 3 deletions and 2 addictions (aka move 2 lines, remove 1 line). That would also allow clients which omit the parameter to use the server.

the-gangster:
What do I read here?
The incompatibility of tvheadend to serve Plex directly could have been solved already since years???
So sad it wasn't merged.
So it remains a pain for everyone new to this topic to find out after trial and error and investigation that it has to be sent through yet another tool (in my case Antennas docker) although tvheadend already emulates an hdhomerun device which would be supported by Plex.
:-(

Flole998:
It could have, if someone would have opened a PR with a proper fix instead of doing some overly complicated workaround. But apparently it wasn't that important, so it's still as it was.

the-gangster:
Thank you for the fast reply. 
Well important or not, many people out there are affected. But even if you search for it, you hardly come across this PR. (At least I have already quite a few times without success)
I can't evaluate if the "proper fix" would have worked. To be honest, I even understood @WereCatf the other way around, that Plex always supplies src=1 parameter and not drops it. But I only read here (and might have misunderstood that detail) about it, while I faced the consequences already several times.
But if it appeared to be clear to you what the problem is. and if it really is done by deleting 3, move 2 and remove 1 line, then couldn't you implement it, knowing that it might make life easier for many people (even if "only" Plex fans) in future?
I know it shouldn't be your job, and I shouldn't beg you to do it... But even a workaround would have been nice instead of no fix at all.
But thank you once more for responding.

Flole998:
That's not how a PR is usually handled, and then it probably got lost. What needs to be done is move the `src = atoi....` and following line up where `src = 1;` is right now, then change `"0"` to `"1"` in the lines that just got moved. I'll try to do it when I'm at a computer next time, if someone else does it faster than me that'd also be great and would make sure I don't forget it.

Flole998:
Actually the fix is already in..... Seems like it didn't get lost, much to my surprise. And apparently I was the one who implemented it.

So whatever you are seeing is not related to this issue, or you are running some super old, outdated version.

the-gangster:
No, not super old. Just a couple of days ago I started updating my containers to latest versions again.
I am using images from linuxserver/tvheadend:latest

Build: 4.3-2315~gab6ea89b1 (2024-04-06T15:44:34+0000)
Build: 4.3-2317~g223f83b6e (2024-04-13T15:45:52+0000)
I think I activated the TVH server on the latter one. It was then found by Plex and all assigned channels were detected.
But when I wanted to play them from Plex it failed.

But now that I repeated my tests with same settings it indeed started working! Not as stable and not always at first try but in general it can be used now. So obviously I aborted my testing too early before.

PR #1450, opened 2022-03-21
Update for VAAPI transcoding

ovaron76github:
VAAPI: Buffer factor (buffersize manipulation) now configurable at WebUI.
VAAPI: Added rc_mode (h264/hevc) and tier (hevc), both configurable via WebUI.
VAAPI: Removed B-Frame workaround from the past at HEVC.
VAAPI: Added an option for ignore b-frames because some drivers needs this (AMD).
VAAPI: Also updated VP8 and VP9 encoding.
VAAPI: VP8 and VP9 now also reacts on options for "Buffer factor" and "Ignore B-Frames" and got the "force_key_frames" expression to increase picture quality.
VAAPI: Added destroy option for h264 codec (might have created memory issues)

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -33,6 +33,10 @@ typedef struct {
     TVHVideoCodecProfile;
     int qp;
     int quality;
+    double buff_factor;
+    int rc_mode;
+    int tier;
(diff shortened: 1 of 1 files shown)


ovaron76github:
Anything else needed fort this change or is it okay now?

Flole998:
Looks good to me, thanks!

PR #1451, opened 2022-04-03
dvr_disk_space_cleanup() - do not return error if called again too soon

dave-p:
Fixes #6152

If two recordings begin simultaneously, and the free storage space is below the minimum, the second recording to be processed will fail.

File deletion and update of disk statistics, triggered by the start of the first recording, takes a finite time (possibly several seconds if a disk drive has to spin up). The second recording may therefore see the same space shortage as the first and trigger a further cleanup. However dvr_disk_space_cleanup() returns an error if called within 10 seconds of the last cleanup, causing the recording to fail.

Instead, we return the same number of cleared bytes as was returned on the first call, allowing the second recording to proceed.

### DIFF
--- a/src/dvr/dvr_vfsmgr.c
+++ b/src/dvr/dvr_vfsmgr.c
@@ -36,6 +36,7 @@ static int64_t dvr_disk_space_config_lastdelete;
 static int64_t dvr_bfree;
 static int64_t dvr_btotal;
 static int64_t dvr_bused;
+static int64_t last_clearedBytes;
 static tvh_mutex_t dvr_disk_space_mutex;
 static mtimer_t dvr_disk_space_timer;
 static tasklet_t dvr_disk_space_tasklet;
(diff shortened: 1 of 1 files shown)


Flole998:
I am not sure that fix is correct: The actual cleaning is done before your variable is set, right? So on the second run (while the first one is still running) you'd return 0, on every other run you'd return how many bytes were cleaned on the previous run? Is the actual deletion done asynchronously?

Also in dvr_vfs_rec_start_check() there seems to be no error checking? The -1 is treated as "we cleaned -1 bytes"? So how about this: We catch -1 as an error and use -2 return code for the 10-second-case and then make the rec_start_check return true for that case?

dave-p:
> Also in dvr_vfs_rec_start_check() there seems to be no error checking? The -1 is treated as "we cleaned -1 bytes"? So how about this: We catch -1 as an error and use -2 return code for the 10-second-case and then make the rec_start_check return true for that case?

I will close this PR and resubmit using your suggestion instead.

PR #1457, opened 2022-05-05
HTSP: Add SNR and Signal status for Decibel scale.

bluzee:
Kodi is not getting signal level when TVheadend is set to use new Decibel status.  It currently only gets level if you force old signal status.  Decibel levels have actual meaning while percentage levels are less helpful.  When you know the delivery system, modulation and fec of a carrier then you can know the exact minimum dB SNR to lock without errors. 

I understand the code is crude.  Farmers probably shouldn't write code but I hope this will work at least.


bluzee:
Yes, I'm not sure why it did that.  Do I need to create a branch for the one commit?

Flole998:
I would suggest to revert the commit, stash it, reset to head and then apply it again. Or the easy (but "unprofessional") solution: Delete, clone again, apply those changes, force-push to GitHub (or delete your repo and fork again if you aren't working with a local copy).

bluzee:
I'll close this  and try again.  Sorry for the mess.

PR #1459, opened 2022-05-10
Fixed and cleanup the "PSIP: ATSC Grabber" module (Fixes #5610)

NormRaden:
- Bumped up limit on number of supported EIT/ETT tables from 5 to 256 (this is the max number of EIT/ETT tables in the ATSC specification)
 - Remove table type 4 as a valid ETT table, there is no support for this table here and table type 4 doesn't supply any EPG data--just long names for channels/services, etc.
 - Removed the very rapid toggling between mpegts_table_{add,destroy}(...) calls on EIT/ETT tables that caused continuous enabling/disabling of EIT and ETT PIDs.
 - Added an 'ETM Location' check to EIT table handling to ensure that an event's extra text is cleared in the EPG database if there is no matching ETT.
 - Removed epg_broadcast_change_finish(...) because it was erasing the extra text field in EPG database when updating title in EPG database (in EIT handler).
 - Added the ability for receiving ETT tables with the same version ID.  ETT tables use the same version ID for long runs of event IDs and 'Extra text's.
 - Fix bug in retrieving the number of tables EIT/ETT tables listed in the MGT table.

 - Removed the psip_{find,remove,add}_desc(...) functions (and supporting code) that stored ETT extra texts before
   there was a matching EIT event and then would try to match up the ETT extra texts when the EIT event showed up.
   However, the psip_*_desc(..) functions didn't keep track of the channel and would match up the ETT 'Extra text' with the wrong EIT event.

 - Removed some non-functional code.
 - Removed old commented out test code.

 - Added placeholder support for 'stop' and 'done' functions of this module for future development.

The "Over-the-air: PSIP: ATSC Grabber" module now quickly and correctly populates 'Title', 'Extra text', 'Start time', 'End time', 'Duration', etc... fields in the EPG Database.

----

While there had been some useful development done, I found that the PSIP: ATSC Grabber module really didn't work.  It wouldn't properly populate the EPG database, the "Extra text" fields would briefly flash a sometimes wrong description before disappearing, and a sparse number of "Title" fields would slowly update, and often recordings and live streaming would experience lots of dropped packets while the module was enabled.  I developed a modest patch that improves performance and correctness of the module.

The big bug was that the module was flooding the tuner hardware's PID filter with repeated requests to enable then disable the same set of ETT and EIT PIDs (via repeated calls to mpegts_table_{add,destroy}(...)).  This would severely hamper the processing of EIT and EIT tables.  Additionally, depending on the tuner hardware, would sometimes overload the digital tuner causing constant indiscriminate dropping of packets.  Also, there was a call to epg_broadcast_change_finish(ebc, changes2, 0) that would erased some fields in the EPG database that it shouldn't.  And, the ETT tables are broadcast using the same version number for each event ID, a call to dvb_table_reset(...) was needed after processing each ETT section to reset the version filtering feature built-in to dvb_table_begin(...)/dvb_table_end(...) as a work around, otherwise only one 'Extra text' would be processed.  And, the psip_{find,remove,add}_desc(...) functions would sometimes place the 'Extra text' in the wrong place.  With help from an ATSC PSIP specification, I made some other smaller fixes.  I finished up by removing some non-functional code for better readability and performance.  However, there is still some more work that can be done.

I did stumble across this in the "ATSC Standard: Program and System Information Protocol for Terrestrial Broadcast And Cable" Doc. A65/2013, 7 August 2013:

 Under Section 6.5 Event Information Table (EIT) on page 40 it says:
  "For a particular virtual channel, an event_id identifies uniquely each of the events programmed for the 3-hour interval of an EIT."

 Under Section D.4.3 The Event Information Tables (EITs) on page 116 it says:
 "The event_id is used to link events with their messages defined in the ETT, and therefore it has to be unique only within a virtual channel and a 3-hour interval defined by EITs."

The specification clearly says that event ID's can only be expected to be unique within 3-hour intervals--this would require a more complex implementation.  However, it appears that all the ATSC terrestrial broadcaster's PSIP generators receivable by my antenna in my region of the USA conveniently keep event IDs unique across all 3-hour intervals.  Its time to share the patch and find out if there are any PSIP EPG generators that reuse event IDs between EITs and ETTs in different 3-hour intervals.

Verification:

 - gcc's (version 10.2.1) static analyzer has no complaints about the changed file. 
 - I've been testing this patch for several months and the PSIP: ATSC Grabber module works as expected.
   - Works for ~50 programs/channels carried on 10 ATSC terrestrial broadcast stations located in a metropolitan area in the USA.
   - Typically, will finish updating the EPG database well under 90 seconds for all the programs/channels available on all tuned-in station(s).
   - Doesn't mangle reception of media packets for recording or watching live.
   - When I wasn't certain it was working, I temporarily instrumented some tvhdebug(...)'s to dump a lot of EIT and ETT tables in to the log, and manually verified that the EIT/ETT table's matched up with the expected EPG database contents.

Since, I only use a small subset of tvheadend features, some feedback on this patch would be nice.  I couldn't test ATSC cable broadcasts.  I couldn't test broadcaster's outside of my media market.  If you live in an ATSC region (North America and some surrounding islands, South Korea, parts of South America, etc...) please give this patch a try and let me know what you think.  Don't forget to test this module out BEFORE applying this patch first.

Enable the "PSIP: ATSC Grabber" module (under Configuration -> Channel/EPG -> EPG Grabber Modules) and disable any competing modules.  I found that a setting between 60 and 90 for "EPG scan time-out in seconds" (under Configuration -> Channel/EPG -> EPG Grabber) works well.

### DIFF
--- a/src/epggrab/module/psip.c
+++ b/src/epggrab/module/psip.c
@@ -29,7 +29,7 @@
 #include "input.h"
 #include "input/mpegts/dvb_charset.h"
 
-#define PSIP_EPG_TABLE_LIMIT 5
+#define PSIP_EPG_TABLE_LIMIT 512
 
 static int _psip_eit_callback(mpegts_table_t *mt, const uint8_t *ptr, int len, int tableid);
 static int _psip_ett_callback(mpegts_table_t *mt, const uint8_t *ptr, int len, int tableid);
@@ -40,19 +40,10 @@ static int _psip_ett_callback(mpegts_table_t *mt, const uint8_t *ptr, int len, i
 
 typedef struct psip_table {
   TAILQ_ENTRY(psip_table) pt_link;
-  mpegts_table_t         *pt_table;
-  int64_t                 pt_start;
-  uint16_t                pt_pid;
-  uint16_t                pt_type;
-  uint8_t                 pt_complete;
+  mpegts_table_t         *pt_table; /* pointer to mpegts_table to receive EIT/ETT tables/sections */
+  uint16_t                pt_pid;  /* packet ID */
+  uint16_t                pt_type; /* EIT/ETT table type */
 } psip_table_t;
-
-typedef struct psip_desc {
-  RB_ENTRY(psip_desc)     pd_link;
-  uint16_t                pd_eventid;
-  uint8_t                *pd_data;
-  uint32_t                pd_datalen;
-} psip_desc_t;
 
 typedef struct psip_status {
   epggrab_module_ota_t    *ps_mod;
@@ -61,10 +52,6 @@ typedef struct psip_status {
   epggrab_ota_mux_t       *ps_ota;
   mpegts_mux_t            *ps_mm;
   TAILQ_HEAD(, psip_table) ps_tables;
-  RB_HEAD(, psip_desc)     ps_descs;
-  mtimer_t                 ps_reschedule_timer;
-  uint8_t                  ps_complete;
-  uint8_t                  ps_armed;
 } psip_status_t;
 
 static void
@@ -72,7 +59,6 @@ psip_status_destroy ( mpegts_table_t *mt )
 {
   psip_status_t *st = mt->mt_opaque;
   psip_table_t *pt;
-  psip_desc_t *pd;
 
   lock_assert(&global_lock);
   assert(st->ps_refcount > 0);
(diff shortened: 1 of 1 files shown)


NormRaden:
This patch has no effect if the "over-the-air: PSIP: ATSC Grabber" module is disabled.  It consists of a minimal number of changes to just get this module working.  It continues to work well for me and I expect no one else will have problems with this.  A merge to the mainline might be a practical way to get some feedback.

Please conduct a source code review and let me know of any issues.  Thanks.

PR #1463, opened 2022-07-13
Regexps for the finnish EIT scraping

garyttirn:
Added regexps for the finnish EIT scraping


garyttirn:
Thanks for quick merge.

PR #1471, opened 2022-10-24
epgdb: Resolve symlinks before using file location

DonKult:
The new epgdb is written to a temporary file and later renamed to override the old epgdb file atomically. If you diverted the epgdb to a different place away from your usual configuration (e.g. for space and/or disk usage reasons like on an OpenWrt router) this leads to overriding the symlink with a real file defeating the point.

By applying realpath on the path first we can resolve any symlink along the path, while not considering it a failure if the epgdb file doesn't exist yet. If on the other hand the path up to the file doesn't exist we default to the old way of just taking the path verbatim and let them be created by hts_settings_makedirs as before.

Note that this relies on the paths being sized PATH_MAX, which as the [manpage](https://manpages.debian.org/unstable/manpages-dev/realpath.3.en.html) notes is POSIX.1-2001 conform, but broken by design as PATH_MAX can't be relied upon, but the entire codebase makes heavy use of PATH_MAX and there is a pre-existing usage of realpath() in this way so lets pretend its okay for now.

References: b23686a55323625b15d4f99fd7af55259fa21828

---

I am unsure how contribution is supposed to work here as the guidelines make references to a CLA, but that website seems broken? Also to freenode irc… anyhow, I don't have an account in the forums/redmine either. I am not even set on using tvheadend yet as I am actually shopping around a bit in terms of if I want to drop my VDR usage for tvheadend – and if either will have a good time in their supposed new home… a router, which obviously has some space and usage constraints (although it has 4x arm64 @ 2.2GHz and 1GB RAM, so not THAT constraint). Just mentioning so you have an idea what my background is and why I might or might not come up with similar special interest features. And as a big sorry for probably ignoring all established project rules as I somewhat failed to find current ones… :pray:

### DIFF
--- a/src/epgdb.c
+++ b/src/epgdb.c
@@ -302,7 +302,9 @@ static void epg_save_tsk_callback ( void *p, int dearmed )
   int fd, r;
 
   tvhinfo(LS_EPGDB, "save start");
-  hts_settings_buildpath(path, sizeof(path), "epgdb.v%d", EPG_DB_VERSION);
+  hts_settings_buildpath(tmppath, sizeof(path), "epgdb.v%d", EPG_DB_VERSION);
+  if (!realpath(tmppath, path))
+    strlcpy(path, tmppath, sizeof(path));
   snprintf(tmppath, sizeof(tmppath), "%s.tmp", path);
   if (hts_settings_makedirs(tmppath))
     fd = -1;


M95D:
I'm running tvheadend on OpenWrt. I also made the last update for tvheadend package for OpenWrt.
OpenWrt has a setting for the location of epgdb. Is there any chance this change might break that? (I must confess I don't understand much C or the implications of this change).

DonKult:
Hi @M95D, truth be told, I am actually trying to fix that setting :wink: See [tvheadend: place epgdb.v3 in /tmp instead of the gone v2](https://github.com/DonKult/openwrt-packages/commit/b0b79616f8d939bc1ea6c2723655805065cb6b6c) which I was going to propose in OpenWrt if tvheadend upstream accepts this branch as I would like to avoid carrying a patch.

M95D:
OK. :-)

Just curious: How will you use tvheadend on a router? DVC-C? DVB-S? IPTV? Live or recording? Play on browser? What other clients will you use?

Flole998:
Have you considered moving this into `hts_settings_buildpath()`? But I assume other usages don't suffer from this issue as they don't use a move/rename operation?

Don't worry about that CLA.

DonKult:
(sry, I don't login on github all to often, got sidetracked first & after that had forgot about having this open unresolved…)

> Have you considered moving this into `hts_settings_buildpath()`? But I assume other usages don't suffer from this issue as they don't use a move/rename operation?

Briefly, but I indeed think it only happens for epgdb as most other files have a unique (uuid) name anyway and I didn't want to risk breaking something and/or making the review harder. If @Flole998  prefers I can move it through.

@M95D: Setup is rather simple (or so I claim): 2 budget tuners attached that can do dvb-t2 and dvb-c. Unencrypted only. No sat, no ip. tvheadend is mostly there as a recorder. have my grips with the UI, but I manage for now with the help of dave-p/TVHadmin-JS and some crude patches (to both). The recordings are shared via NFS and watched via mpv for now as that was my setup before with vdr as well. Might explore "proper" clients like Kodi at some point, but not sure. They seem heavy and I am more the lightweight type. I do live stream occasionally, but if I really want to watch something and not just have background noise, I tend to record it and watch the recording "live". Mostly because I can pause & rewind at will which livestreams tend not to without a "proper" client. I am the only user in the household through, so I can do whatever works for me and for now the focus is on getting the foundation solid.

Flole998:
No Problem, I just wanted to double check if this is indeed the right place to implement it, and as we seem to agree this is most likely the right place.

Thanks for your contribution!

M95D:
OpenWrt package updated! https://github.com/openwrt/packages/pull/19952
Thanks!

PR #1472, opened 2022-10-27
iptv_auto: Add support for m3u "channel-number" tag

mamarley:
SiliconDust recently added m3u playlist generation support to the HDHomeRun products.  They use "channel-number" for the channel number, so add support for that value to TVHeadend.  It is used as a fallback only if the previously-supported "tvh-chnum" and "tvg-chno" tags aren't present.  This should be particularly useful for usage with ATSC 3.0 channels, as those aren't supported in the libhdhomerun library that TVHeadend normally uses for HDHomeRun devices.  Without this, the channel number for all channels imported from the HDHomeRun m3u is "0".


th0ma7:
@mamarley and @Flole998 this is working as expected with one caveat: the channel number is being kept in the channel name resulting in naming such as `17.1 SomeChannelName`.  What's missing is the ability to remove the numbering prefix from the name portion.

PR #1473, opened 2022-11-20
Misc small things: mimetypes, esoteric crash and compiler appeasement

DonKult:
Hi,

as a maintainer I tend to dislike PRs which bundle a bunch of unrelated stuff together but… apart from the ipv6 change perhaps I guess there isn't too much to discuss about them, I assume (as I hope I didn't make some obvious errors). Feel free to cherry-pick and/or make me split it up properly.

### DIFF
--- a/src/epg.c
+++ b/src/epg.c
@@ -1061,7 +1061,6 @@ epg_broadcast_t *epg_broadcast_clone
     *save |= epg_broadcast_set_credits(ebc, src->credits, &changes);
     *save |= epg_broadcast_set_category(ebc, src->category, &changes);
     *save |= epg_broadcast_set_keyword(ebc, src->keyword, &changes);
-    *save |= epg_broadcast_set_description(ebc, src->description, &changes);
     *save |= epg_broadcast_set_serieslink_uri
                (ebc, src->serieslink ? src->serieslink->uri : NULL, &changes);
     *save |= epg_broadcast_set_episodelink_uri
--- a/src/tcp.h
+++ b/src/tcp.h
@@ -68,8 +68,7 @@ static inline int ip_check_equal_v4
 
 static inline int ip_check_equal_v6
   (const struct sockaddr_storage *a, const struct sockaddr_storage *b)
-    { return ((uint64_t *)IP_AS_V6(a, addr).s6_addr)[0] == ((uint64_t *)IP_AS_V6(b, addr).s6_addr)[0] &&
-             ((uint64_t *)IP_AS_V6(a, addr).s6_addr)[1] == ((uint64_t *)IP_AS_V6(b, addr).s6_addr)[1]; }
+    { return memcmp(IP_AS_V6(a, addr).s6_addr, IP_AS_V6(b, addr).s6_addr, sizeof(IP_AS_V6(a, addr).s6_addr)) == 0; }
 
 static inline int ip_check_equal
   (const struct sockaddr_storage *a, const struct sockaddr_storage *b)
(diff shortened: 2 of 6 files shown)


Flole998:
Actually GCC wasn't just confused, the old code was comparing pointers and even if that would have been fixed there would have been a strict aliasing violation. So your fix is the correct solution.

Thanks for your contribution!

DonKult:
Huh, that is strange… I fail to see my error but compilers tend to be right, don't they…

I am on Debian sid with gcc-12 on amd64 and get no warning (note that the Makefile by default seems to disables this specific warning since 2018) even if I enable it explicitly. The same machine cross-compiles with gcc-11(.3) for aarch64 with the default and I run tvheadend on aarch64 (in OpenWrt).

I can indeed reproduce with gcc-11 (not with gcc-12) with this line: `/usr/bin/gcc-11 -g -O2 -D_FORTIFY_SOURCE=2 -fPIE -Werror -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -fms-extensions -fno-strict-aliasing -Ibuild.linux -Isrc -I.  -c -o /dev/null src/tcp.c`

I am reluctant, but as gcc-12 doesn't complain about the code, I guess we have indeed a false positive here.

Are you adding the `-Werror=stringop-overread` manually or does that come in by default from somewhere (as in, do you build a package or …) ?

P.S.: strict-aliasing is disabled by the Makefile as well as this tcp code in particular breaks it, but it did so before, I didn't make it worse (I hope).

ProfYaffle:
FWIW, a quick Google shows up multiple conversations around false positives for `-Werror=stringop-overread` on gcc-11 - the common theme seems to be "string and memory operations on fixed address", but I've long forgotten how to use pointers properly, so don't wholly understand. As far as I can see, the logic was changed in gcc-12 and backported to gcc-11, but gcc 11.3 on my Ubuntu 22.04 box still chokes. It's fine if I use  gcc-12.

The `-Werror` flag seems to be a default. It can be suppressed by passing `--nowerror` at build time (I use Autobuild.sh, so can pass it through as part of `AUTOBUILD_CONFIGURE_EXTRA=`). This obviously suppresses all warnings-as-errors, however.

catalinii:
I am using the following command to configure:
./configure --disable-libav --disable-pie --disable-ffmpeg_static  --disable-libx264 --disable-libx265 --disable-libvpx  --disable-libtheora --disable-libvorbis --disable-libfdkaac  --disable-nvenc --disable-libx264_static --disable-libx265_static --disable-libvpx_static  --disable-libtheora_static --disable-libvorbis_static --disable-libfdkaac_static --disable-hdhomerun_static --enable-debug --enable-trace --disable-dvben50221 --disable-libopus

Then I use make.

ProfYaffle:
If you just want to suppress this specific warning (versus a very blunt `-nowerror`), I think these should work:

`./configure <existing options> --cflags="-Wno-stringop-overread"`

or

`AUTOBUILD_CONFIGURE_EXTRA="<existing options> --cflags=\"-Wno-stringop-overread\" ./Autobuild.sh -t <target>`

mralapete:
> FWIW, a quick Google shows up multiple conversations around false positives for `-Werror=stringop-overread` on gcc-11 - the common theme seems to be "string and memory operations on fixed address", but I've long forgotten how to use pointers properly, so don't wholly understand. As far as I can see, the logic was changed in gcc-12 and backported to gcc-11, but gcc 11.3 on my Ubuntu 22.04 box still chokes. It's fine if I use gcc-12.
> 
> The `-Werror` flag seems to be a default. It can be suppressed by passing `--nowerror` at build time (I use Autobuild.sh, so can pass it through as part of `AUTOBUILD_CONFIGURE_EXTRA=`). This obviously suppresses all warnings-as-errors, however.

Tried AUTOBUILD_CONFIGURE_EXTRA=--enable-libffmpeg_static\ --enable-trace\ --enable-debug --cflags="-Wno-stringop-overread" ./Autobuild.sh
This returned --cflags=-Wno-stringop-overread: command not found
Is my syntax correct here.

ProfYaffle:
This isn't really a support forum, but in case it helps people in the meantime (@DonKult has already PRed a code change that handles this, so it'll be fixed soon) - either enclose all options in quotes, and escape the `cflags` bit, or escape the spaces in between the options, i.e. one of:

`AUTOBUILD_CONFIGURE_EXTRA="--enable-libffmpeg_static --enable-trace --enable-debug --cflags=\"-Wno-stringop-overread\"" ./Autobuild.sh`

or

`AUTOBUILD_CONFIGURE_EXTRA=--enable-libffmpeg_static\ --enable-trace\ --enable-debug\ --cflags="-Wno-stringop-overread" ./Autobuild.sh`

PR #1474, opened 2022-11-23
Avoid breaking strict aliasing in TCP code

DonKult:
Hi,

as #1473 broke compiling with gcc-11 due to a false positive I feel responsible for making it work again, even if is somewhat easy to workaround and semi-temporary (just use a fixed compiler, right? :wink: ).

So what we do here is making tvheadend compile with strict-aliasing rules enabled, which as a by-product apparently doesn't lead gcc-11 down the wrong path. yeaha!

On the downside, I have basically only compile-tested this (on gcc-12 and a bit with gcc-11) as I run tvheadend with `--noacl` and many compiletime options disabled… and while I believe both Makefile changes should be done they might break compiling with older compiler versions or on code which breaks strict aliasing, but isn't detected (and hence potentially wrongly optimized). Disabling strict-aliasing on the entire codebase is not without bug potential either through as seen with the gcc-11 false positive. So I choose to do both in separate commits which could be easily reverted if it turns out the world burns with them. I would hope not, but you never know… (I wasn't really expecting running into compiler bugs either).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,7 @@ CFLAGS  += -Werror
 endif
 CFLAGS  += -Wall -Wwrite-strings -Wno-deprecated-declarations
 CFLAGS  += -Wmissing-prototypes
-CFLAGS  += -fms-extensions -funsigned-char -fno-strict-aliasing
-ifeq ($(COMPILER), gcc)
-CFLAGS  += -Wno-stringop-truncation -Wno-stringop-overflow
-endif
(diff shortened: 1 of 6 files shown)


Flole998:
Can you please squash your commits together? This is basically a single change so it doesn't really make sense to split it.

DonKult:
I said my piece on why I think otherwise, but I acknowledge that house rules can be different and so I force-pushed a squash… :man_shrugging:

Flole998:
Thanks!

PR #1475, opened 2022-11-27
Use a build matrix in GitHub Actions to build on more platforms

Jalle19:


Flole998:
I'm sorry but this can't be merged at the moment as there's a major rebuild of the entire build stuff going on at my fork. All that's left is fixing a few failing builds there by installing a more recent python version so the build can be uploaded to the new repository.

Jalle19:
I see, didn't know you were working on this already. I was mainly looking to fix the build on Debian 11 and figured it'd make sense to make a better CI pipeline while I'm at it.

Do you want any help with the workflow on your fork?

Flole998:
Sure, actually I think I can already merge it at this point and then the remaining builds can be fixed. Some targets need newer python versions, so downloading and compiling a new-ish python version is what's missing there. Other than that it's pretty much complete.

Flole998:
Actually we can still

Flole998:
Actually this can still be used as CI target so all PRs are checked with this. So if you want to resolve the merge conflict we can get this in aswell.

Jalle19:
Okay, so you're looking to have separate jobs for pull requests only?

Flole998:
I've merged my changes and there are now 3 jobs:
1. Coverity scan job (this one needs to be modified to include as many features as possible)
2. CI job (this one runs on all PRs, here your changes can be added)
3. Repo build job (that's the python issue I mentioned earlier)

PR #1486, opened 2022-12-24
update vaapi

alingherghescu:
- added new settings: platform, bitrate scale factor, low power, loop filter level, loop filter sharpness, async depth
- implemented new (dynamic) settings adjustment (in js)
- added new parameters: b, low_power, loop_filter_level, loop_filter_sharpness, async_depth

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -33,10 +33,16 @@ typedef struct {
     TVHVideoCodecProfile;
     int qp;
     int quality;
+    int low_power;
+    int async_depth;
+    double max_bit_rate;
+    double bit_rate_scale_factor;
+    int platform;
+    int loop_filter_level;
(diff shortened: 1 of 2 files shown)


alingherghescu:
I found one issue with VP9 (previously didn't had HW to test). 
This issue might affect some users ... assuming they have a CODEC that has only VAEntrypointEncSliceLP (and not VAEntrypointEncSlice).
Function vaapi_encode_setup_context() is calling tvhva_context_create() with VAEntrypointEncSlice as last parameter.
If you have a VAAPI CODEC that doesn't have support VAEntrypointEncSlice (like I have with VP9 on my Celeron 5305) the initialization will fail in function: tvhva_context_check_profile() where is going to check support.
Let me clarify ... even without the changes pushed above VP9 will still not work.
We need to discuss how to fix this elegantly.

vaapi

Saentist:
If there is a code with check HW acceleration capability on TVH start... .

PR #1460, opened 2022-05-28
Migrate libavformat muxer to FFmpeg 5 API

chewi:
As you may be aware, Tvheadend does not build against FFmpeg 5 due to API changes involving codecpar and BSF. This is going to become a headache for distributions quite soon. I tried to update the code to make it work, but have struggled due to what seems to be a slightly complex use case. I'm submitting what I have so far so that no one has to start from scratch with this. The `/* Check */` comments show where the result should be checked.

There isn't much documentation around this API or how to migrate, but I did find [a great example](https://github.com/mgba-emu/mgba/blob/31201d49033a0519e3272e3359ca75b97c183e6d/src/feature/ffmpeg/ffmpeg-encoder.c) of code supporting both the old and new APIs. I don't think you need to support both, but it still demonstrates how you can migrate. Unfortunately, it's a simpler use case. Look for `FFMPEG_USE_CODECPAR` and `FFMPEG_USE_NEW_BSF`.

PR #1487, opened 2022-12-31
remove libavresample from build scripts

pimzand:
libavresample is not being used. Builds should not depend on it.
Fixes #6084 and #6219

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,6 @@ ifeq ($(CONFIG_LIBAV),yes)
 FFMPEG_LIBS := \
     libavfilter \
     libswresample \
-    libavresample \
     libswscale \
     libavformat \
     libavcodec \
--- a/configure
+++ b/configure
@@ -642,7 +642,6 @@ else
 
     check_pkg libavfilter   ">=6.47.100"  || has_libav=false
     check_pkg libswresample ">=2.1.100"   && has_resample=true
-    check_pkg libavresample ">=3.0.0"     && has_resample=true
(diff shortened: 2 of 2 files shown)


Saentist:
What about old OS support capability?
Do generally need to cutoff some option?

pimzand:
Why would an old OS require libavresample for this version of tvheadend, and a new OS not?

woooow1:
It seems ok, but build fail when update makefile.ffmpeg to use new ffmpeg 5.x.x version!

src/libav.c:226:15: error: 'AVCodec {aka struct AVCodec}' has no member named 'encode'
   return codec->encode || codec->encode2;
               ^~
src/libav.c:226:32: error: 'AVCodec {aka struct AVCodec}' has no member named 'encode2'
   return codec->encode || codec->encode2;

                                ^~

Flole998:
Ffmpeg 5 needs to be addressed in #1460, not here.

woooow1:
I think they both related! Libavresample is also deprecated in ffmpeg 4.x.x

woooow1:
Just test this PR, everything builds and run fine with build in ffmpeg ( ./Autobuild.sh ). I also remove avresample from makefile.ffmpeg.

pimzand:
Some clarification: this is not needed for static ffmpeg builds, but with this PR applied, it becomes possible to use `compat-ffmpeg4.rpm` dynamically on Fedora/RPMFusion.
In fact, this is what the RPMFusion maintainer  is doing in the tvheadend rpm in RPMFusion.
Credits for this PR to the maintainer.

PR #1490, opened 2023-01-07
update vaapi

alingherghescu:
- added new settings: platform, bitrate scale factor, low power, loop filter level, loop filter sharpness, async depth
- implemented new (dynamic) settings adjustment (in js)
- added new parameters: b, low_power, loop_filter_level, loop_filter_sharpness, async_depth
- tvhva_context_check_profile() will change TVHVAContext->entrypoint from VAEntrypointEncSlice into VAEntrypointEncSliceLP if VAEntrypointEncSlice is not available for that CODEC (according to VAAPI info)
- moved low_power to tvh_codeo_profile in order to initialize properly the entrypoint
- many error reporting improvements
- separated some bundled conditions

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -22,10 +22,37 @@
 #include <fcntl.h>
 #include <sys/ioctl.h>
 
+#include "idnode.h"
+#include "htsmsg.h"
 
-#define AV_DICT_SET_QP(d, v, a) \
-    AV_DICT_SET_INT((d), "qp", (v) ? (v) : (a), AV_DICT_DONT_OVERWRITE)
+/* hts ==================================================================== */
 
+static htsmsg_t *
+platform_get_list( void *o, const char *lang )
+{
+    static const struct strtab tab[] = {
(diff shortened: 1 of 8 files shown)


src/transcoding/codec/codecs/libs/vaapi.c
```diff
@@ -310,14 +636,14 @@ static const codec_profile_class_t codec_profile_vaapi_hevc_class = {
         .ic_properties = (const property_t[]){
             {
                 .type     = PT_INT,
-                .id       = "tier",
+                .id       = "tier",     // Don't change
                 .name     = N_("Tier"),
-                .desc     = N_("Set tier (general_tier_flag) [0=main 1=high]"),
+                .desc     = N_("Set tier (-1, 0 or 1) [-1=skip 0=main 1=high]"),
```
Flole998, line 641: This should probably be a combo instead, allowing to choose between "" (empty), "main" and "high".

alingherghescu: I agree with you; can we take that action for next release? I plan to add more h264 profiles (4:2:2, 4:4:4) . In terms of priority I think the next few things are higher priority.

PR #1491, opened 2023-01-10
profile video resize improvements

alingherghescu:
- provide the ability to specify if scaling should be performed Up and Down, Up (only) or Down (only)

### DIFF
--- a/src/transcoding/codec/profile_video_class.c
+++ b/src/transcoding/codec/profile_video_class.c
@@ -22,6 +22,16 @@
 
 #include <libavutil/pixdesc.h>
 
+static htsmsg_t *
+scaling_mode_get_list( void *o, const char *lang )
+{
+    static const struct strtab tab[] = {
+        { N_("Up & Down"),      0 },
+        { N_("Up (only)"),      1 },
+        { N_("Down (only)"),    2 },
+    };
+    return strtab2htsmsg(tab, 1, lang);
+}
 
 /* TVHCodec ================================================================= */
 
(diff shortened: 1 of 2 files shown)


src/transcoding/codec/profile_video_class.c
```diff
@@ -63,11 +73,38 @@ tvh_codec_profile_video_setup(TVHVideoCodecProfile *self, tvh_ssc_t *ssc)
...
+            if ((self->height > 0) && (self->size.den > self->height)) {
+                self->size.den = self->height;
+                self->size.den += self->size.den & 1;
+                self->size.num = (int)((double)self->size.den * ((double)ssc->es_width / (double)ssc->es_height));
+                self->size.num += self->size.num & 1;
+            }
+            break;
+        default:
+            // should never reach this point.
+            tvherror(LS_PROFILE, "self->scaling_mode unexpected value: valid range [0,1,2] received %d", self->scaling_mode);
```
Flole998, line 106: This should be something like `Invalid scaling mode: valid range 0-2, received %d`

alingherghescu: I tend to disagree .... let me explain:
- the user will not see this message unless someone changes the source code in order to add more options in the combo box. 
- this message is intended for developers.
- I can change it ... but frankly I think is correct as it is.

alingherghescu: Another option is to remove default: completely.

Flole998: An error message can be visual for a user and even for developers it should be clear to understand.

However, if there is no way that message will be shown normally it should be removed completely.

alingherghescu:
Let me know if the trick worked.

PR #1494, opened 2023-01-19
Add configurable delays after Unicable operations

pbasista:
Allow user to adjust the length of time delays after the Unicable device is powered up and after a DiSEqC command is sent to it.

These time delays are currently [hardcoded](https://github.com/tvheadend/tvheadend/blob/760f32bf531e15346a40cef864f87edd5bae9681/src/input/mpegts/linuxdvb/linuxdvb_en50494.c#L374) to fixed values. As a result, regular users cannot easily adjust them. From my experience, the currently used default values might cause some Unicable devices to behave unreliably when operated by Tvheadend.

<details><summary><i>In more detail:</i></summary>

I have tried to use a Unicable II (EN50607) LNB marketed as Inverto [IDLU-32L412-UNBRR-OPN](https://www.inverto.tv/lnb/493/programmable-32ub-dcss-lnb) with Tvheadend. However, the scanning and tuning were both unreliable and failed in roughly 50 % of cases.

I have also tried to do the same with Enigma2. I have used the same LNB, connected via the same cable arrangement to the same receiver device. I have only changed the application which was operating the LNB from Tvheadend to Enigma2. Then, both scanning and tuning worked reliably all the time.

Knowing that this particular LNB can in fact be operated reliably, I have checked the differences in handling of the communication with Unicable devices between Tvheadend and Enigma2. I have noticed that Enigma2 uses longer delays. In particular:

| | Tvheadend | Enigma2 |
| :-- | --: | --: |
| Delay after setting voltage to 18V and before sending the DiSEqC command | [15+15](https://github.com/tvheadend/tvheadend/blob/760f32bf531e15346a40cef864f87edd5bae9681/src/input/mpegts/linuxdvb/linuxdvb_en50494.c#L340) ms | [75](https://github.com/openatv/enigma2/blob/dead1bc76eb4579e70e9257842ab674f071b0d34/lib/python/Components/NimManager.py#L1617) ms |
| Delay after sending the DiSEqC command and before setting voltage to 13V | [50](https://github.com/tvheadend/tvheadend/blob/760f32bf531e15346a40cef864f87edd5bae9681/src/input/mpegts/linuxdvb/linuxdvb_en50494.c#L374) ms | [100](https://github.com/openatv/enigma2/blob/dead1bc76eb4579e70e9257842ab674f071b0d34/lib/python/Components/NimManager.py#L1609) ms |

I have then tried to increase these time delays in Tvheadend and it indeed helped. In my case, using 200 ms delays at both places was the only change necessary to make the scanning and tuning work reliably in Tvheadend.

This PR aims to add the possibility to configure those time delays for regular users.

</details>

---

This PR also adjusts some strings which are subject to translation. However, I have not updated the translation files yet because I am unsure how to properly do so. I have noticed that the command `make intl` regenerates the `.pot` files. But I am unsure how to reasonably propagate changes from the `.pot` files to the `.po` files which contain the actual translations.

I have also noticed that regenerating the `.pot` files on code from the master branch results in a lot of changes. My understanding of the process for handling the translations is very limited so I might be missing something obvious.


Flole998:
Please squash your changes together.

pbasista:
> Please squash your changes together.

Ok, I will do that after code review.

Flole998:
Actually let's keep them as individual commits as each commit addresses a single change and they can be applied and reverted individually if necessary.

Flole998:
Thanks!

PR #1495, opened 2023-01-23
update vaapi

alingherghescu:
- updated function _video_filters_get_filters() due to a bug hidden by previous bug (Bug #6247)
- fixed Decoder HW + Encoder HW --> was generating download / upload that increase CPU usage
- remove vp9 profile0 to 3 (ffmpeg vaapi is not supporting profiles for vp9)
- improve 'tier' and added: level, qmin, qmax, desired_b_depth
- printing Bitrate, Buffer size, Max bitrate (for each transcoding start)
- added superframe for vp9

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -54,6 +54,71 @@ rc_mode_get_list( void *o, const char *lang )
     return strtab2htsmsg(tab, 1, lang);
 }
 
+// h264
+
+static htsmsg_t *
+h264_level_get_list( void *o, const char *lang )
+{
+    static const struct strtab tab[] = {
+        { N_("skip"), -100 },
+        { N_("1"),     10 },
+        { N_("1.1"),   11 },
+        { N_("1.2"),   12 },
+        { N_("1.3"),   13 },
+        { N_("2"),     20 },
+        { N_("2.1"),   21 },
+        { N_("2.2"),   22 },
+        { N_("3"),     30 },
+        { N_("3.1"),   31 },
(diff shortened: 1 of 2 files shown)


src/transcoding/codec/codecs/libs/vaapi.c
```diff
@@ -292,25 +394,69 @@ tvh_codec_profile_vaapi_h264_open(tvh_codec_profile_vaapi_t *self,
     int int_bitrate = (int)((double)(self->bit_rate) * 1024.0 * (1.0 + (self->bit_rate_scale_factor * ((double)(self->size.den) - 480.0) / 480.0)));
     int int_buffer_size = (int)((double)(self->bit_rate) * 2048.0 * self->buff_factor * (1.0 + self->bit_rate_scale_factor * ((double)(self->size.den) - 480.0) / 480));
     int int_max_bitrate = (int)((double)(self->max_bit_rate) * 1024.0 * (1.0 + (self->bit_rate_scale_factor * ((double)(self->size.den) - 480.0) / 480.0)));
+    tvherror(LS_VAAPI, "Bitrate = %d bps; Buffer size = %d bps; Max bitrate = %d bps", int_bitrate, int_buffer_size, int_max_bitrate);
```
Flole998, line 397: This is not an error though, is it?

Flole998: There's tvhinfo, tvhdebug and tvhtrace aswell if I remember correctly. The last 2 will only be printed if debugging is active, you should probably use those.

alingherghescu: I will update on my next PR. 
Thank you.

PR #1481, opened 2022-12-04
Update debian/compat to version 13

janisozaur:
Compat version 13 is currently the recommended one.

An important change introduced in 10 was change of default to target parallel builds

https://github.com/Debian/debhelper/blob/5d1bb29841043d8e47ebbdd043e6cd086cad508e/debhelper.pod#compatibility-levels

PR #1497, opened 2023-01-24
Debian build debhelper compat version should be updated

Saentist:
Let's remove message
`dh_installdeb: warning: Compatibility levels before 9 are deprecated (level 7 in use)`


Flole998:
Duplicate of #1481

PR #1498, opened 2023-01-28
Revert "Update debian/compat to version 10"

Flole998:
Reverts tvheadend/tvheadend#1481


Saentist:
This still not fix same problem on CentOS with or without this commit

Flole998:
It fixes the issue that was introduced with the above-mentioned commit, so this does exactly what it's designed to do.

Saentist:
problem is only in 22.04 not in previous version
https://github.com/tvheadend/tvheadend/blob/bed37ea208b8acaf914b4fb14498d143a1fbbd93/Autobuild/debian.sh#L28

Flole998:
That is not correct.

Saentist:
> That is not correct.

With one? 
That this is not a bug of 22.04
or there is a problem elsewhere?

```bash
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.5 LTS
Release:        20.04
Codename:       focal
```
[build_compat_7.log](https://github.com/tvheadend/tvheadend/files/10528431/build_compat_7.log)
[build_compat_10.log](https://github.com/tvheadend/tvheadend/files/10528430/build_compat_10.log)

no problem on 20.04
same on 18.04 and 16.04

But same problem exist on CentOS with is RPM distribution and do not care about Debian packager.

Flole998:
Discussion of issues not related to this PR are offtopic here.

Our Repo build system clearly shows where it failed, and that's more than just a single OS.

PR #1500, opened 2023-01-31
vaapi update:

alingherghescu:
- replaced tvherror() with tvhinfo for bitrate report
- converted from bps to kbps
- reduced B frame to max 3 (4 is generating artifacts on my system)
- I had to set also bf otherwise will be set later in profile_video_class() with 3
- when low power is enabled max B frame will be disabled (because codec is not using B frames in low power mode)
- h264_vaapi and hevc_vaapi have also dynamic enable/disable for max B frame
- vp9 super frames can be enabled/disabled from the interface
- clean-up some javascript code

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -138,6 +138,7 @@ typedef struct {
     int level;
     int qmin;
     int qmax;
+    int super_frame;
 } tvh_codec_profile_vaapi_t;
 
 #if defined(__linux__)
@@ -254,7 +255,7 @@ static const codec_profile_class_t codec_profile_vaapi_class = {
                 .type     = PT_BOOL,
                 .id       = "low_power",     // Don't change
                 .name     = N_("Low Power"),
-                .desc     = N_("Set low power mode.[if disabled will not send paramter to libav]"),
+                .desc     = N_("Set low power mode.[if disabled will not send parameter to libav, if enabled codec will disable B frames]"),
                 .group    = 3,
                 .opts     = PO_EXPERT,
                 .get_opts = codec_profile_class_get_opts,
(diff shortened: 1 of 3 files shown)


alingherghescu:
Do you think those failures are related to what I changed?

Flole998:
Unrelated, it's because libvpx somehow has a different/unexpected checksum now. If you want to fix it have a look at Makefile.ffmpeg and update the libvpx to 1.12 (and change the sha hash accordingly), then it should build successfully again

src/transcoding/codec/codecs/libs/vaapi.c
```diff
@@ -454,8 +455,12 @@ tvh_codec_profile_vaapi_h264_open(tvh_codec_profile_vaapi_t *self,
         case 0:
             // Uncontrained --> will allow any combination of parameters (valid or invalid)
             // this mode is usefull fur future platform and for debugging.
-            if (self->desired_b_depth) {
-                AV_DICT_SET_INT(opts, "b_depth", self->desired_b_depth, AV_DICT_DONT_OVERWRITE);
+            if (self->low_power == 0) {
+                if (self->desired_b_depth) {
```
Flole998, line 459: These 2 conditions can be combined using `&&`

src/webui/static/app/codec.js
```diff
@@ -252,9 +253,17 @@ var codec_profile_forms = {
                     bit_rate_scale_factor_field.setDisabled(false);
                     qp_field.setDisabled(false);
                     low_power_field.setDisabled(false);
+                    desired_b_depth_field.setDisabled(false);
                     break;
                 case 1:
                     // Intel
+                    var low_power = low_power_field.getValue();
+                    if (low_power == 1) {
+                        desired_b_depth_field.setDisabled(true);
```
Flole998, line 262: Don't use `if` here, just add the condition directly instead of true/false

src/transcoding/codec/codecs/libs/vaapi.c
alingherghescu: done

src/webui/static/app/codec.js
alingherghescu: done

PR #1501, opened 2023-02-01
descrambler: detect and support icam patched libdvbcsa

pimzand:
This PR will detect icam patches in libdvbcsa at compile time, and add tvheadend support for it conditionally.
Fixes #6227

### DIFF
--- a/src/descrambler/tvhcsa.c
+++ b/src/descrambler/tvhcsa.c
@@ -220,12 +220,20 @@ tvhcsa_set_type( tvhcsa_t *csa, struct mpegts_service *s, int type )
 }
 
 
+#ifndef DVBCSA_ICAM
 void tvhcsa_set_key_even( tvhcsa_t *csa, const uint8_t *even )
+#else
+void tvhcsa_set_key_even( tvhcsa_t *csa, const uint8_t *even, const uint8_t ecm)
+#endif
 {
   switch (csa->csa_type) {
   case DESCRAMBLER_CSA_CBC:
 #if ENABLE_DVBCSA
+#ifndef DVBCSA_ICAM
     dvbcsa_bs_key_set(even, csa->csa_key_even);
+#else
+    dvbcsa_bs_key_set_ecm(ecm, even, csa->csa_key_even);
+#endif
 #endif
     break;
   case DESCRAMBLER_DES_NCB:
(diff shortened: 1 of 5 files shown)


Flole998:
Just as an idea: All those `ifdef` are not really necessary, the only real change is what parameters are passed to 2 function calls from libdvbcsa. I think it might be easier to always pass the third parameter down and either drop it there or pass it along. That should also allow tvhcsa to use this parameter in the future aswell in case other descrambler providers are added.

Also do you have any idea if/how packages with this option can be compiled and made available using our repository? Maybe it would even be possible to use dlopen to dynamically open the library and detect which function is available at runtime, something like `DYNAMIC_DVBCSA`, then there is no need to know at compiletime which version will be available at runtime.

pimzand:
Wait, it's not only about extra arguments to existing functions, but also calls to a different function, `dvbcsa_bs_key_set()` versus `dvbcsa_bs_key_set_ecm()`. I have no idea how to handle that.

Flole998:
And that's basically the only difference between the current and the new version. So there is only a single `#ifdef` for each of those 2 functions necessary. That's the first thing I would change/improve.

For the different functions, as I said: You could use dlopen to detect which of the functions is defined at runtime and then make a decision based on that.

th0ma7:
> About the patched libdvbca, there are a couple of forks on github that contain the patches, either as a patch file, or readily applied. But nobody is providing binary packages for popular distributions as far as I know.

Not that Synology is a "popular" distro but I do currently provide it:
- https://synocommunity.com/package/tvheadend
- https://github.com/SynoCommunity/spksrc/blob/master/cross/libdvbcsa/patches/001-libdvbcsa-OSCam-support_v2.patch
- https://github.com/SynoCommunity/spksrc/blob/master/cross/tvheadend/patches/005-tvh-OSCam-support.patch

But I'll be more than happy to drop it entirely if possible :)

pimzand:
ifdefs are now reduced to the bare minimum.
CI Pipeline is still running, but I could build and run this code with both plain and patched libdvbcsa.

I can squash commits if you want me to.

src/descrambler/tvhcsa.c
```diff
@@ -220,12 +220,16 @@ tvhcsa_set_type( tvhcsa_t *csa, struct mpegts_service *s, int type )
 }
 
 
-void tvhcsa_set_key_even( tvhcsa_t *csa, const uint8_t *even )
+void tvhcsa_set_key_even( tvhcsa_t *csa, const uint8_t *even, const uint8_t ecm)
```
Flole998, line 223: You accidentally removed the space here at the end of the arguments.

Also please squash your commits together.

Flole998:
Closing due to inactivity.

TonKro:
I did build the @pimzand branch and using it together with OSCam r11718-798. It is descrambling for several CAID but somehow not for Sky Germany.
How to debug to find the problem. Anyone help or PM me.

PR #1503, opened 2023-02-10
Update Makefile.ffmpeg

uknunknown:
- update ffmpeg from 4.4.1 to 4.4.3


Flole998:
If you make a single change like this the commit message should be "update ffmpeg from xxx to yyy" and not "update makefile.ffmpeg". Like this it is not clear when looking at a collapsed list of commits what this commit does.

PR #1508, opened 2023-03-05
fix for 64bit time_t on 32bit systems

pykpkg47:
on some 32bit systems   time_t  has a size of 64 bit  ( alpine with musl => 1.2, or glibc with D_TIME_BITS 64 option) 

tested on a  armv32 arm sbc and on x86 with qemu emulation


Flole998:
As expected this broke many targets. Please have a look at what's wrong and fix it or this needs to be reverted.

PR #1512, opened 2023-03-16
migrate from AVBitStreamFilterContext to AVBSFContext

uknunknown:
- migrate from AVBitStreamFilterContext to AVBSFContext

### DIFF
--- a/src/muxer/muxer_libav.c
+++ b/src/muxer/muxer_libav.c
@@ -35,8 +35,10 @@
 typedef struct lav_muxer {
   muxer_t;
   AVFormatContext *lm_oc;
-  AVBitStreamFilterContext *lm_h264_filter;
-  AVBitStreamFilterContext *lm_hevc_filter;
+  AVBSFContext *ctx;
+  const AVBitStreamFilter *bsf_h264_filter;
+  const AVBitStreamFilter *bsf_hevc_filter;
+  const AVBitStreamFilter *bsf_vp9_filter;
   int lm_fd;
   int lm_init;
 } lav_muxer_t;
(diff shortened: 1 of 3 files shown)


uknunknown:
Let me highlight some of the changes:

src/muxer/muxer_libav.c
```diff
@@ -172,6 +174,39 @@ lav_muxer_add_stream(lav_muxer_t *lm,
...
+      }
+    }
+    else {
+      if (av_bsf_alloc(lm->bsf_h264_filter, &lm->ctx)) {
+        tvherror(LS_LIBAV,  "Failed to alloc AVBitStreamFilter for h264_mp4toannexb");
+        return -1;
+      }
+    }
+    if(avcodec_parameters_copy(lm->ctx->par_in, st->codecpar)) {
+      tvherror(LS_LIBAV,  "Failed to copy paramters to AVBSFContext");
```
Flole998, line 202: Here you fail to release the allocation from above, is that intended?

uknunknown: actually is worst than that: in case of any error with premature return -1; I don't release .. let me fix this problem :(.

Flole998: Maybe you can use a `goto` to jump to the end  of this function where all the freeing is happening. The return code is then stored in a rc variable and in the end there is a simple "return rc;"

uknunknown:
updated. 
Also the same problem was on function: lav_muxer_init()
with: av_dict_free(&opts);

src/muxer/muxer_libav.c
uknunknown: done

uknunknown:
I removed also some functions that are deprecated starting ffmpeg4 (as shown in the https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges.

PR #1513, opened 2023-03-21
update to ffmpeg codecpar

uknunknown:
- update to ffmpeg codecpar


uknunknown:
Maybe you should wait for the results from https://tvheadend.org/boards/5/topics/49414?r=49439 before submit. Is your call .... it was working on my test. (but ... more coverage is better).

uknunknown:
based on results from thread: https://tvheadend.org/boards/5/topics/49414?r=49439 the changes are not introducing a regression so my proposal is to continue with PR.

PR #1514, opened 2023-03-22
updated 'AVCodec' to 'const AVCodec'

uknunknown:
- updated 'AVCodec' to 'const AVCodec'
- "avctx->refcounted_frames = 1;" deprecated (not required with: avcodec_receive_frame())

### DIFF
--- a/src/transcoding/transcode/stream.c
+++ b/src/transcoding/transcode/stream.c
@@ -56,7 +56,7 @@ static int
 tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
 {
     enum AVCodecID icodec_id = streaming_component_type2codec_id(ssc->es_type);
-    AVCodec *icodec = NULL, *ocodec = NULL;
+    const AVCodec *icodec = NULL, *ocodec = NULL;
 
     if (icodec_id == AV_CODEC_ID_NONE) {
         tvh_stream_log(self, LOG_ERR, "unknown decoder id for '%s'",
--- a/src/transcoding/codec.h
+++ b/src/transcoding/codec.h
@@ -59,7 +59,7 @@ struct tvh_codec {
     const char *name;
     size_t size;
     const codec_profile_class_t *idclass;
-    AVCodec *codec;
+    const AVCodec *codec;
     const AVProfile *profiles;
     int (*profile_init)(TVHCodecProfile *, htsmsg_t *conf);
     void (*profile_destroy)(TVHCodecProfile *);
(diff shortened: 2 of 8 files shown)


src/transcoding/transcode/stream.c
```diff
@@ -56,7 +56,7 @@ static int
 tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
 {
     enum AVCodecID icodec_id = streaming_component_type2codec_id(ssc->es_type);
-    AVCodec *icodec = NULL, *ocodec = NULL;
+    const AVCodec *icodec = NULL, *ocodec = NULL;
```
Flole998, line 59: Why would you create a const null-pointer? It's modified later on aswell, so const is not correct.

uknunknown: When I tried to upgrade to ffmpeg5, instances with 'AVCodec' were flagged .... we can wait until I fix all other issues and check if works without this line. All instances (initialized or not with null) of AVCodec are modified later on .... What I can tell you is that transcoding works with this change. We can remove the initialization will null ... I don't think it matters. Do you want me to check that?

Flole998: It doesn't make sense like this. I assume what you want is to cast it as `const` when passing it to ffmpeg. I'm not sure why the compiler isn't complaining about assigning something to a const variable... I'll need to understand what's going on, then I can tell you what the best way is.

uknunknown: I searched through ffmpeg code (examples) and is always defined without initialization (to null). Let me remove those '= NULL' and see if are working without those. I don't think it makes a difference ... :) but let's fixed the 'right' way.

uknunknown: It doesn't work without explicit initialization to NULL :( .
init

Flole998: The problem is that `const` means constant, but you are modifying the constant. I don't get why the compiler isn't loudly complaining about const variables being changed, I might have to ask someone else to explain this to me.

That null assignment isn't a problem, the problem is that a const variable should never change, that's why it's const.

uknunknown: If you prefer to lower the risk, I can wrap all changed for ffmpeg5 behind '--enable-ffmpeg5' (that will be default no). This way we can get more feedback for other platforms.

jkellerer: `const AVCodec *icodec` is a mutable pointer to immutable `const AVCodec`
  while
`AVCodec *const icodec` is an immutable pointer to mutable `AVCodec`

Can be confusing especially when coding in many languages :)

PR #1515, opened 2023-03-23
remove deprecated av_frame_get_best_effort_timestamp()

uknunknown:
- av_frame_get_best_effort_timestamp() can be removed according to tvheadend/build.linux/ffmpeg/ffmpeg-4.4.3/libavutil/frame.h because is part of old API (FF_API_FRAME_GET_SET)

### DIFF
--- a/src/transcoding/transcode/video.c
+++ b/src/transcoding/transcode/video.c
@@ -309,7 +309,6 @@ tvh_video_context_open(TVHContext *self, TVHOpenPhase phase, AVDictionary **opts
 static int
 tvh_video_context_encode(TVHContext *self, AVFrame *avframe)
 {
-    avframe->pts = av_frame_get_best_effort_timestamp(avframe);
     if (avframe->pts <= self->pts) {
         tvh_context_log(self, LOG_WARNING,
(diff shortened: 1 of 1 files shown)


Flole998:
You're simply removing the entire check for invalid pts timestamps there? That doesn't seem right to me...

Flole998:
The correct way to do this would be to access `AVFrame->best_effort_timestamp` directly.

uknunknown:
This function is gone completely in ffmpeg5. This is the comment from ffmpeg 4.4.3.
deprecated
Let me check that :)

Flole998:
Exactly, because the field can be accessed directly, there's no need for the accessor anymore.

uknunknown:
I verified ... and it works. Fixed (pushed to PR).

Flole998:
It's still not correct. Access the field of the struct directly to copy the variable. Don't just randomly remove code, this will break things!!

uknunknown:
probably you are right ... I think is time for me to step aside (before we break something) and let the pro do the job .... anyway, was in interesting journey!

Flole998:
Feel free to continue working on it, it was not my intention to have you stop working on this. I have created a PR with the "proper" way of doing this now and mentioned you there, so it should pop up in your notifications.

PR #1517, opened 2023-03-24
Remove deprecated get_best_effort_timestamp() call

Flole998:


Flole998:
@uknunknown This is what I meant

PR #1518, opened 2023-03-29
remove deprecate struct vaapi_context and the vaapi.h

uknunknown:
- remove deprecate struct vaapi_context and the vaapi.h

### DIFF
--- a/src/transcoding/transcode/hwaccels/vaapi.c
+++ b/src/transcoding/transcode/hwaccels/vaapi.c
@@ -21,7 +21,6 @@
 #include "../internals.h"
 #include "vaapi.h"
 
-#include <libavcodec/vaapi.h>
 #include <libavutil/hwcontext.h>
 #include <libavutil/hwcontext_vaapi.h>
 #include <libavutil/pixdesc.h>
@@ -37,7 +36,9 @@ typedef struct tvh_vaapi_device {
 
 
 typedef struct tvh_vaapi_context_t {
-    struct vaapi_context;
+    void *display;
+    uint32_t config_id;
+    uint32_t context_id;
     const char *logpref;
     VAEntrypoint entrypoint;
     enum AVPixelFormat io_format;


uknunknown:
I don't think the failure is from my changes. I don't know how to restart the 'checks;.
I am not a specialist but I think this is a low risk change.
This change is required because vaapi.h will be removed from ffmpeg5 ... so we need to remove it from tvh also.

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -37,7 +36,9 @@ typedef struct tvh_vaapi_device {
 
 
 typedef struct tvh_vaapi_context_t {
-    struct vaapi_context;
+    void *display;
```
Flole998, line 39: Why are these now in here? Where are they used?

uknunknown: vaapi.h will be removed in ffmpeg5
In order to remove the dependency from vaapi_context, I copied the struct content into tvh_vaapi_device.
In my opinion the struct tvh_vaapi_device is identical before and after change (let me know if you disagree).


vaapi.h (formatting is not working ... for same reason)

`
struct attribute_deprecated vaapi_context {
    /**
     * Window system dependent data
     *
     * - encoding: unused
     * - decoding: Set by user
     */
    void *display;

    /**
     * Configuration ID
     *
     * - encoding: unused
     * - decoding: Set by user
     */
    uint32_t config_id;

    /**
     * Context ID (video decode pipeline)
     *
     * - encoding: unused
     * - decoding: Set by user
     */
    uint32_t context_id;
};
`

Flole998: No the struct is not identical. Previously it contained a struct, now it contains individual variables.

Flole998: Also I fixed the formatting of your comment so proper code tags are used

Flole998: Try just deleting it without adding any other variables. Seems like it's not used at all.

uknunknown: It doesn't work because is not able to resolve TVHVAContext->config_id, TVHVAContext->context_id and TVHVAContext->display. My understanding of 'struct in a struct' is that compiler will instantiate the struct vaapi_context with his variable inside TVHVAContext. Somehow the compiler is resolving those TVHVAContext->display (aldo the display is not defined explicit in struct TVHVAContext). If I copy the variable from vaapi_context  inside TVHVAContext when I access that struct instance is going to find those 3 variable.
error

PR #1519, opened 2023-04-01
tvhmeta: Fix tvhmeta authentication to the tvheadend API.

pool15:
Construct and add an Authorization header to the request, when a username and password are provided to tvhmeta.

This fixes #6260.


pool15:
Many thanks!

PR #1523, opened 2023-04-14
update pict_type from AVPacket to AVFrame

uknunknown:
- remove deprecated FF_API_CODED_FRAME
- remove pict_type from AVPacket_SideData
- use AVFrame->pict_type (same like ffmpeg 5.1.2 - ffprobe.c line 2595)
- remove patch for vaapi_encode


uknunknown:
Let me provide some details:
- vaapi_encode.patch is connected to video.c code.
- vaapi_encode.patch is adding pict_type to: avctx->coded_frame->pict_type and ff_side_data_set_encoder_stats()
- video.c is reading that info ( pict_type ) from both places: av_packet_get_side_data and self->oavctx->coded_frame->pict_type (in that order).
- is a little unclear for me why do we store the info in two places ... but probably because it was expected that FF_API_CODED_FRAME will be removed at some point (in ffmpeg6) they baked in the future solution (av_packet_***_side_data).
- My point is that both are not optimal because ffmpeg is offering this info in AVFrame (as you can see in ffprobe.c implementation), so we should use the same implementation like ffprobe.c
- It is worth noting that I left the backup:
    if (pict_type == AV_PICTURE_TYPE_NONE && avpkt->flags & AV_PKT_FLAG_KEY) {
        pict_type = AV_PICTURE_TYPE_I;
    }
--> only I added "pict_type == AV_PICTURE_TYPE_NONE" as a condition.
- Also I left this 'hack' explained in comments, that was also before:
        // some codecs do not set pict_type but set key_frame, in this case,
        // we assume that when key_frame == 1 the frame is an I-frame
        // (all the others are assumed to be P-frames)
        if (self->oavframe->key_frame) {
            pict_type = AV_PICTURE_TYPE_I;
        }
        else {
            pict_type = AV_PICTURE_TYPE_P;
        }
--> only that I take the keyframe info from AVFrame
ffprobe

PR #1526, opened 2023-04-26
Local ASCII to UTF-8

Saentist:
Local to UTF-8 encoding


Saentist:
@Flole998 
Codeball error is not reason PR not to be merged. 
```log
Warning: The `set-output` command is deprecated and will be disabled soon. 
Please upgrade to using Environment Files. 
For more information see: 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

data/conf/epggrab/opentv/dict/skynz
```diff
@@ -645,187 +645,187 @@
   },
   {
     "code": "101010111000110000110010111",
-    "data": "�"
+    "data": "Ў"
```
Flole998, line 648: This entire file looks broken, at least when viewing it on GitHub

Saentist: Because is on local encoding not UTF-8.

Hebrew is total unreadable on any other local encoding even on github without conversion.

image

Flole998: Well what you are showing in the screenshot is still wrong, this is supposed to be the upside down exclamation mark...

Have you tested your PR in practice? Or just converted the encoding without testing what effect it has?

Saentist: In screenshot you see pr where on left is old local encoding on right utf-8


other source of same dictionary
https://raw.githubusercontent.com/jahutchi/openTVtoXML/standalone/providers/skynz_160.0.dict

Flole998: And on the right side it is still wrong.

Unless you come up with a good reason why we should switch the encoding I'm going to close this as it's clearly breaking things and apparently you haven't tested your changes either.

PR #1528, opened 2023-05-12
Add support for 12-hour custom date formats

pmatis:
I prefer to use 12 hour format instead of having to think through what 18:30 means, for example. Added support for it.
Noticed that someone also wanted this 4 years ago and it doesn't exist yet: https://tvheadend.org/boards/5/topics/37347

### DIFF
--- a/docs/markdown/cmdline_options.md
+++ b/docs/markdown/cmdline_options.md
@@ -27,7 +27,7 @@ Usage: `tvheadend [OPTIONS]`
       --satip_bindaddr        Specify bind address for SAT>IP server
       --satip_rtsp            SAT>IP RTSP port number for server
                               (default: -1 = disable, 0 = webconfig, standard port is 554)
-      --nosatip               Disable SAT>IP client
+      --nosatipcli            Disable SAT>IP client
       --satip_xml             URL with the SAT>IP server XML location
 ```
 ### Server connectivity
--- a/man/tvheadend.1
+++ b/man/tvheadend.1
@@ -66,7 +66,7 @@ Only use specified DVB adapters (comma separated, -1 == none).
 SAT>IP RTSP port number for server (default: -1 = disable, 0 = 
 webconfig, standard port is 554).
 .TP
-\fB\-\-nosatip\fR
+\fB\-\-nosatipcli\fR
 Disable SAT>IP client.
 .TP
 \fB\-\-satip_xml\fR
--- a/src/main.c
+++ b/src/main.c
@@ -843,7 +843,7 @@ main(int argc, char **argv)
               opt_threadid     = 0,
               opt_libav        = 0,
               opt_ipv6         = 0,
-              opt_nosatip      = 0,
+              opt_nosatipcli   = 0,
               opt_satip_rtsp   = 0,
 #if ENABLE_TSFILE
               opt_tsfile_tuner = 0,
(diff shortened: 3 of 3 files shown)


src/webui/static/app/tvheadend.js
```diff
@@ -1393,6 +1393,9 @@ tvheadend.toCustomDate = function(date, format) //author: meizz, improvements by
             "\%M+" : date.getMonth()+1, //month
             "\%d+" : date.getDate(),    //day
             "\%h+" : date.getHours(),   //hour
+            "\%H+" : date.getHours() % 12 || 12,   //hour (12 hour format)
+            "\%P" : (date.getHours() >= 12) ? "PM" : "AM",   //am pr pm for 12 hour format
+            "\%p" : (date.getHours() >= 12) ? "pm" : "am",   //am pr pm for 12 hour format, lower case
```
oliv3r, line 1398: so this is a little 'confusing' (which is to blame the original code for sure).

The 'date' manual (potentially unrelated, but for consistency important) states:
```
       %b     locale's abbreviated month name (e.g., Jan)
       %h     same as %b
       %H     hour (00..23)
       %I     hour (01..12)
       %p     locale's equivalent of either AM or PM; blank if not known
       %P     like %p, but lower case
```

As `h` 'wrong' (should have been `H`). I don't see that`I` is already used here, so i'd probably recommend you do use that instead. And while at it, make `%H` do the same as `%h` here, so we have some consistency between 'regular date' and hts.

Further more (yes, it's stupid I agree) LOWER case p, is whatever your locale dicates, and UPPER case is the lower case version.

I'm not sure if there's any other locale for AM/PM, date seems to think so; so that is a bit iffy now, as you should use a localized string here. Personally (I'm not a maintainer) I'd ignore that and just use 'am/pm' here and forget about localization, because other then Americans, it's probably not really in use and it can always be fixed later on demand :)

pmatis: I agree with your assessment, and I've updated my pull request.

Flole998:
Please squash your commits

src/webui/static/app/tvheadend.js
```diff
@@ -1386,13 +1386,17 @@ tvheadend.toLocaleFormat = function()
 	return tvh_locale_lang.replace('_','-');
 };
 
-tvheadend.toCustomDate = function(date, format) //author: meizz, improvements by pablozg
+tvheadend.toCustomDate = function(date, format) //author: meizz, improvements by pablozg and pmatis
```
oliv3r, line 1389: imo i'd rather see this entire comment removed, it doesn't add anything, git history tells us this :p but i'll let @Flole998 decide :)

Flole998: Yeah there's no point in doing this. If we do this everywhere the code will be a nightmare to read, so let's remove it. To me it's unclear to which part of the code it applies anyways (this line, this function, part of this function....)

pmatis: I see the point but was following the lead. I didn't feel right removing it myself nor leaving it as "blaming" someone for my massacred code changes. :) I'll remove it.

pmatis:
> Please squash your commits

Sorry, but... what does this mean?

Flole998:
> > Please squash your commits
> 
> Sorry, but... what does this mean? A search gave me the basic idea - reduce the pull request to one commit. However, I'm not sure how to do this. I'm using git on a Linux command line, BTW. I could git --reset soft then commit and push force, but something tells me there's a better way. I'll await feedback.

Yeah, you can reset and then commit again and force push

oliv3r:
> > Please squash your commits
> 
> Sorry, but... what does this mean? A search gave me the basic idea - reduce the pull request to one commit. However, I'm not sure how to do this. I'm using git on a Linux command line, BTW. I could git --reset soft then commit and push force, but something tells me there's a better way. I'll await feedback.

Another way is to use `git rebase -i origin/master` for example and then mark the commit with `s` for squash (if you want to keep the commit message), or `f` fixup, if you just want to squash without a commit message.

It takes some getting used to, but very powerful!

Flole998:
This still needs to be rebased and squashed (revert my last commit for that)

Flole998:
Closed due to inactivity.

PR #1529, opened 2023-05-20
update vaapi - vainfo

uknunknown:


uknunknown:
Note:
- vainfo detection happens at power on so after vainfo checkbox is enabled you need to restart tvh (in order to see any difference). 
- Codec Profiles defined before enabling vainfo must be made again because those will have some 'default' value that are not valid. Each codec will populated with available info from vainfo when is generates and store those results in XML.

PR #1530, opened 2023-05-20
update vaapi

uknunknown:

### DIFF
--- a/src/prop.c
+++ b/src/prop.c
@@ -46,6 +46,7 @@ static const struct strtab typetab[] = {
   { "s64",     PT_S64_ATOMIC },
   { "dbl",     PT_DBL },
   { "time",    PT_TIME },
+  { "int",     PT_DYN_INT },
   { "langstr", PT_LANGSTR },
   { "perm",    PT_PERM },
 };
@@ -84,6 +85,7 @@ prop_write_values
   uint32_t u32, opts;
   uint16_t u16;
   time_t tm;
+  int dyn_i;
 #define PROP_UPDATE(v, t)\
   snew = &v;\
   if (!p->set && (*((t*)cur) != *((t*)snew))) {\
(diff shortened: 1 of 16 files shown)


uknunknown:
Note:
    vainfo detection happens at power on so after vainfo checkbox is enabled you need to restart tvh (in order to see any difference).
    Codec Profiles defined before enabling vainfo must be made again because those will have some 'default' value that are not valid. Each codec will populated with available info from vainfo when is generates and store those results in XML.

src/prop.c
```diff
@@ -46,6 +46,7 @@ static const struct strtab typetab[] = {
   { "s64",     PT_S64_ATOMIC },
   { "dbl",     PT_DBL },
   { "time",    PT_TIME },
+  { "int",     PT_DYN_INT },
```
Flole998, line 49: What's this for?

uknunknown: PT_"DYN_INT" is from Dynamic Integer --> is required in vaapi.c to load 'ui' and 'uilp'. It has to paired with .def.dyn_i.
dyn_int

The value has to be loaded from a function (that provide results from vinfo probing). 
Previously inode had only 'dynamic list' (that ca be loaded from a function).
list

I see that is not compiling if you don't have VAAPI ; we need to exclude vainfo.c / vainfo.h if VAAPI is not enabled.

uknunknown:
I did what I always do when I squash and force push ... but this time result was different.
I don't know how to fix it - except from Close and Open a new one.
Flole let me know what do you prefer ... and most importantly if the changes looks OK.

Flole998:
What went wrong during squashing? Most likely there is a way to easily fix it.

uknunknown:
I did the same thing I did before: selected the two commits I want to squash. After squash ask me if I want to Force Push; and I said Yes. And instead of one commit I have 4. Even worst is that when I tried to Amend I generated another two.
error

Flole998:
Unfortunately I don't really have time to look at it, so closing and opening again is probably the best thing to do.

uknunknown:
submitted a new one: https://github.com/tvheadend/tvheadend/pull/1531

Flole998:
Thanks

PR #1531, opened 2023-05-23
update vaapi - vainfo

uknunknown:


uknunknown:
This PR is a duplicate of https://github.com/tvheadend/tvheadend/pull/1530

src/prop.c
```diff
@@ -84,6 +85,7 @@ prop_write_values
   uint32_t u32, opts;
   uint16_t u16;
   time_t tm;
+  int dyn_i;
```
Flole998, line 88: There's no need to declare it here, the declaration can be moved to where it's actually assigned/used

uknunknown: If you search for 'tm' (variable above) you will see was defined and used the same. I followed the same 'code style'.

uknunknown: @Flole998 do you have any updates in this request?

uknunknown: @Flole998 this is my last attempt :( . Let me know if you have time to work on this this request.

CvH: @Flole998 ^^^^^^^^

Flole998:
It looks like there is a lot of duplicate code added/in 2 files. Could you please check if that is really needed or if there is any way to reduce duplicate code? You can click on the link of the sonarcloud-bot to get more info about what it doesn't like.

uknunknown:
I am assuming you are referring to the section vaapi.c line 348
PR1531_vaapi_c_348

We can redefine that in a single function with a parameter being passed out (codec) ... and we can add a switch for each codec ... but is really improving? In my opinion will require mode CLK to return the bits :( ... every time you call it will pass thorough the 'switch'  

- let me know if the 'duplicate code' is referring to something else.

uknunknown:
I removed duplicate code from vaapi.c .

src/config.c
```diff
@@ -2753,6 +2759,19 @@ const idclass_t config_class = {
       .opts   = PO_EXPERT,
       .group  = 7,
     },
+#if ENABLE_VAAPI
+    {
+      .type   = PT_BOOL,
+      .id     = "enable_vainfo",
+      .name   = N_("Enable vainfo detection"),
+      .desc   = N_("Enable vainfo detection in order to show only "
+                   "encoders that are advertized by VAAPI driver.\n"
```
Flole998, line 2768: advertised, not advertized I think

PR #1532, opened 2023-06-04
container: Offer TVHeadend container

oliv3r:
Containers have become quite popular, especial when running 'server software'. TVHeadend shouldn't miss out here.

This MR creates a container and uses github actions to build and push these to the github container registry (ghcr.io) which is 'included' in each github repo.

Running tvheadend becomes then as simple as `docker run --rm --interactive --tty --publish 9981-9982:9981-9982/tcp ghcr.io/tvheadend/tvheadend:latest --help`.

It is not as advanced as linuxserver/tvheadend as they add additional programs, not yet? packaged or which should really be handled differently (separate containers/volumes).

Once a few relases have been tagged, and everything works well, extending the action with a secondary login to also push to docker hub for example, is easily possible.

Enjoy!

This MR now depends on #1535 (and in a way on #1536 too, if using the compose solution proposed in the (not yet) written documentation :)

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,1 @@
+Containerfile.alpine
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.config.mk
+build.*/
--- a/.github/workflows/container-build.yaml
+++ b/.github/workflows/container-build.yaml
@@ -1,0 +1,74 @@
+name: Create and publish Container image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - master
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        include:
+          - container: Containerfile.debian
+            autotag: false
+            suffix: -debian
+          - container: Containerfile.alpine
+            autotag: auto
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=edge
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+          flavor: |
+            latest=${{ matrix.autotag }}
+            suffix=${{ matrix.suffix }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          platforms: linux/amd64,linux/386,linux/arm64,linux/arm/v7,linux/arm/v6
+          context: .
+          file: ${{ matrix.container }}
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/Containerfile.alpine
+++ b/Containerfile.alpine
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Copyright (C) 2023 Olliver Schinagl <oliver@schinagl.nl>
+
+ARG ALPINE_VERSION="latest"
+ARG TARGET_ARCH="library"
+
+FROM docker.io/${TARGET_ARCH}/alpine:${ALPINE_VERSION} AS builder
+
+WORKDIR /src
+
+COPY . /src/
+
+RUN apk add --no-cache \
+        'avahi-dev' \
+        'bash' \
+        'bsd-compat-headers' \
+        'build-base' \
+        'cmake' \
+        'coreutils' \
+        'dbus-dev' \
+        'ffmpeg4-dev' \
+        'findutils' \
+        'gettext-dev' \
+        'git' \
+        'gnu-libiconv-dev' \
+        'libdvbcsa-dev' \
+        'libhdhomerun-dev' \
+        'libva-dev' \
+        'libvpx-dev' \
+        'linux-headers' \
+        'musl-dev' \
+        'openssl-dev>3' \
+        'opus-dev' \
+        'pngquant' \
+        'python3' \
+        'uriparser-dev' \
+        'wget' \
+        'x264-dev' \
+        'x265-dev' \
+        'zlib-dev' \
+    && \
+    git config --global --add safe.directory '/src/data/dvb-scan' && \
+    ./configure \
+                --prefix='/usr/local' \
+                --disable-doc \
+                --disable-execinfo \
+                --disable-ffmpeg_static \
+                --disable-hdhomerun_static \
+                --disable-libfdkaac_static \
+                --disable-libmfx_static \
+                --disable-libopus_static \
+                --disable-libtheora_static \
+                --disable-libvorbis_static \
+                --disable-libvpx_static \
+                --disable-libx264_static \
+                --disable-libx265_static \
+                --enable-bundle \
+                --enable-dvbcsa \
+                --enable-hdhomerun_client \
+                --enable-kqueue \
+                --enable-libav \
+                --enable-nvenc \
+                --enable-pngquant \
+                --enable-qsv \
+                --python=python3 \
+        && \
+        make DESTDIR='/tvheadend' -j$(($(nproc) - 1)) install
+
+FROM docker.io/${TARGET_ARCH}/alpine:${ALPINE_VERSION}
+
+LABEL maintainer="Olliver Schinagl <oliver@schinagl.nl>"
+
+EXPOSE 9981 \
+       9982 \
+       9983
+
+RUN apk add --no-cache \
+        'avahi' \
+        'dbus-libs' \
+        'ffmpeg4' \
+        'ffmpeg4-libavcodec' \
+        'ffmpeg4-libavdevice' \
+        'ffmpeg4-libavfilter' \
+        'ffmpeg4-libavfilter' \
+        'ffmpeg4-libavformat' \
+        'ffmpeg4-libavutil' \
+        'ffmpeg4-libpostproc' \
+        'ffmpeg4-libswresample' \
+        'ffmpeg4-libswscale' \
+        'gnu-libiconv-libs' \
+        'libcrypto3' \
+        'libdvbcsa' \
+        'libhdhomerun-libs' \
+        'libssl3' \
+        'liburiparser' \
+        'libva' \
+        'libvpx' \
+        'mesa' \
+        'opus' \
+        'perl-http-entity-parser' \
+        'pngquant' \
+        'python3' \
+        'tini' \
+        'x264-libs' \
+        'x265-libs' \
+        'xmltv' \
+        'zlib' \
+    && \
+    [ "$(uname -m)" = 'x86'* ] && apk add --no-cache \
+        'libva-intel-driver' \
+        ; \
+    [ "$(uname -m)" = 'x86_64' ] && apk add --no-cache \
+        'intel-media-driver' \
+        ; \
+    addgroup -S 'tvheadend' && \
+    adduser -D -G 'tvheadend' -h '/var/lib/tvheadend' -s '/bin/nologin' -S 'tvheadend' && \
+    adduser 'tvheadend' 'audio' && \
+    adduser 'tvheadend' 'usb' && \
+    adduser 'tvheadend' 'video' && \
+    install -d -m 775 -g 'tvheadend' -o 'tvheadend' '/var/lib/tvheadend/recordings' && \
+    install -d -m 775 -g 'tvheadend' -o 'tvheadend' '/var/log/tvheadend'
+
+COPY --from=builder "/tvheadend" "/"
+COPY "./support/container-entrypoint.sh" "/init"
+
+VOLUME /var/lib/tvheadend
+VOLUME /var/lib/tvheadend/recordings
+WORKDIR /var/lib/tvheadend/
+USER tvheadend
+
+ENTRYPOINT [ "/sbin/tini", "--", "/init" ]
--- a/Containerfile.debian
+++ b/Containerfile.debian
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Copyright (C) 2023 Olliver Schinagl <oliver@schinagl.nl>
+
+ARG DEBIAN_VERSION="stable-slim"
+ARG TARGET_ARCH="library"
+
+FROM docker.io/${TARGET_ARCH}/debian:${DEBIAN_VERSION} AS builder
+
+ENV DEBIAN_FRONTEND noninteractive
+
+WORKDIR /src
+
+COPY . /src/
+
+RUN apt-get update --yes && apt-get install --yes \
+        'build-essential' \
+        'bzip2' \
+        'cmake' \
+        'debhelper' \
+        'gettext' \
+        'git-core' \
+        'libavahi-client-dev' \
+        'libavdevice-dev' \
+        'libdbus-1-dev' \
+        'libdvbcsa-dev' \
+        'libhdhomerun-dev' \
+        'libopus-dev' \
+        'libpcre2-dev' \
+        'libssl-dev' \
+        'liburiparser-dev' \
+        'libva-dev' \
+        'libvpx-dev' \
+        'libx264-dev' \
+        'libx265-dev' \
+        'lsb-release' \
+        'pkg-config' \
+        'pngquant' \
+        'python3' \
+        'python3-distutils-extra' \
+        'python3-requests' \
+        'wget' \
+        'zlib1g-dev' \
+    && \
+    git config --global --add safe.directory '/src/data/dvb-scan' && \
+    ./configure \
+                --prefix='/usr/local' \
+                --disable-doc \
+                --disable-execinfo \
+                --disable-ffmpeg_static \
+                --disable-hdhomerun_static \
+                --disable-libfdkaac_static \
+                --disable-libmfx_static \
+                --disable-libopus_static \
+                --disable-libtheora_static \
+                --disable-libvorbis_static \
+                --disable-libvpx_static \
+                --disable-libx264_static \
+                --disable-libx265_static \
+                --enable-bundle \
+                --enable-dvbcsa \
+                --enable-hdhomerun_client \
+                --enable-kqueue \
+                --enable-libav \
+                --enable-nvenc \
+                --enable-pngquant \
+                --enable-qsv \
+                --python=python3 \
+        && \
+        make DESTDIR='/tvheadend' -j$(($(nproc) - 1)) install
+
+FROM docker.io/${TARGET_ARCH}/debian:${DEBIAN_VERSION}
+
+ENV DEBIAN_FRONTEND noninteractive
+
+LABEL maintainer="Olliver Schinagl <oliver@schinagl.nl>"
+
+EXPOSE 9981 \
+       9982 \
+       9983
+
+RUN apt-get update --yes && apt-get install --yes \
+        'libavahi-client3' \
+        'libavcodec59' \
+        'libavdevice59' \
+        'libavfilter8' \
+        'libavfilter8' \
+        'libavutil57' \
+        'libdbus-1-3' \
+        'libdvbcsa1' \
+        'libhdhomerun4' \
+        'libpostproc56' \
+        'libswscale6' \
+        'libopus0' \
+        'libpcre2-8-0' \
+        'libssl3' \
+        'liburiparser1' \
+        'libva2' \
+        'libvpx7' \
+        'libx264-164' \
+        'libx265-199' \
+        'pngquant' \
+        'python3' \
+        'mesa-va-drivers' \
+        'mesa-vdpau-drivers' \
+        'zlib1g' \
+        'xmltv' \
+        'tini' \
+    && \
+    useradd -c 'TVHeadend' -d '/var/lib/tvheadend' -G 'audio','video' -m -r -s '/bin/false' -U 'tvheadend' && \
+    install -d -m 775 -g 'tvheadend' -o 'tvheadend' '/var/lib/tvheadend/recordings' && \
+    install -d -m 775 -g 'tvheadend' -o 'tvheadend' '/var/log/tvheadend'
+
+COPY --from=builder "/tvheadend" "/"
+COPY "./support/container-entrypoint.sh" "/init"
+
+VOLUME /var/lib/tvheadend
+VOLUME /var/lib/tvheadend/recordings
+WORKDIR /var/lib/tvheadend/
+USER tvheadend
+
+ENTRYPOINT [ "/usr/bin/tini", "--", "/init" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+Containerfile
--- a/README.Docker.md
+++ b/README.Docker.md
@@ -1,0 +1,490 @@
+# TVHeadend in Docker
+TVHeadend can be run within a Docker container. This provides isolation from
+other processes by running it in a containerized environment. As this is not
+and in-depth tutorial on docker, those with Docker, containers or cgroups see
+[docker.com][docker].
+
+
+## Building the TVHeadend image
+While it is recommended to pull the image from the [GHCR][ghcr] (GitHub
+Container Registry), it is certainly possible to build the image locally.
+
+To do so, clone this repository:
+
+$ git clone https://github.com/tvheadend/tvheadend
+
+Then, from within the repository
+
+```sh
+docker image build
+    --rm \
+    --tag 'tvheadend:issue-123' \
+    './'
+```
+
+The tag 'issue-123' in the example, is just that, an example, anything can be
+used for the tag.
+
+> _Note_: Omitting the tag, will use `latest` by default.
+
+
+## Running the TVHeadend image
+If the container wasn't built with the above instructions, it can be optionally
+be pulled from the [GHCR][ghcr] first instead.
+
+```sh
+docker image pull ghcr.io/tvheadend/tvheadend:latest
+```
+
+> _Note_: The `latest` tag can be replaced with any desired tag, including
+> `master` for the git development version.
+
+Running the TVHeadend container is then done as follows.
+
+```sh
+docker container run \
+    --interactive \
+    --name 'TVHeadend_container_01' \
+    --rm \
+    --tty \
+    'ghcr.io/tvheadend/tvheadend:issue-123' \
+    --firstrun
+```
+
+> _Note_: Docker will try to pull a container it doesn't know about yet. So if
+> the container was not previously built or pulled, the `run` sub command
+> will try to pull it instead. Likewise, `--pull always` can be used to force
+> a pull. See the docker documentation for more details.
+
+The above will now run TVHeadend and the log output should be visible.
+
+In the snippet above, the `--firstrun` flag was used. This flag is of course
+optional. Please do read the remainder of the next chapter
+[Persistently storing configuration](#Persistently-storing-configuration)
+to learn more about where configuration is stored.
+
+
+## Persistently storing configuration
+Containers do not store files persistently, they are ephemeral by design.
+Obviously, storing of configuration (or video) data is desirable and of course
+is there a solution for this. There are three (probably more) potential options.
+
+It is also important be aware that the configuration directory can be different
+depending on how TVHeadend is started. If tvheadend is started, and a directory
+exists in `/var/lib/tvheadend` or `/etc/tvheadend` which matches the UID of the
+user executing TVHeadend, those locations will be preferred. Since the container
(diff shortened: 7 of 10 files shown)


Flole998:
In the past there have been many issues with containers/virtualization and I am a little hesistant to add it since then we would have to either mark it as unsupported or accept bug reports from people using it.....

oliv3r:
> In the past there have been many issues with containers/virtualization and I am a little hesistant to add it since then we would have to either mark it as unsupported or accept bug reports from people using it.....

Well, it is progress after all.

Not supporting it, is of course 'not nice' for the choice of the users, and worse, you'll have people using third party containers. This would be an 'official' tvheadend container.

There will be some early issues for sure, but once settled in, it will be just another distribution format, the potential user base. As you can see, the files are relativily simple and straight-forward and can even remove/replace some of the existing workflows.

An alternative for users surely is `docker run --rm -it alpine:latest /bin/sh -c 'apk add tvheadend && tvheadend ...` but that's not as nice.

But what container issues are have you seen in the past that worry you? It actually makes things easier too, as you sort-of, have much more control over how it gets packaged/deployed, vs how distros package it?

Flole998:
The issues that have been reported had to do with tuners not being visible or having issues. So a PCI or USB tuner that doesn't work when Tvheadend is running in a container.

Flole998:
Looking at the forum the latest post is another issue report with some container: https://tvheadend.org/boards/5/topics/49938

estimadarocha:
> @oliv3r why not fork the ls.io docker (+add the changes) that is know to work properly for thousands of people (15k pull/month) instead of reinventing the wheel and start at zero. So ls.io could maintain the docker and you would just frequently update it. Or why not fixing the stuff that might be wrong at the ls.io docker and link to them as official docker ?
> 
> Looking at the PR likely not even encoding would work in current state.
> 
> In the end this PR creates yet another fork/variant that is unsupported and untested. As user you still want to use the ls.io docker because its much more feature complete, so you basically just lurk users to use the "official" docker that has no idea about the situation.

+1

oliv3r:
> @oliv3r why not fork the ls.io docker (+add the changes) that is know to work properly for thousands of people (15k pull/month) instead of reinventing the wheel and start at zero. So ls.io could maintain the docker and you would just frequently update it. Or why not fixing the stuff that might be wrong at the ls.io docker and link to them as official docker ?

First of all, ls pulls in the kitchen sink, and then some, compile stuff that are 'nice to have' etc. Sure, can spend days arguing and fixing stuff there :) but then it would still be a ls container.

> Looking at the PR likely not even encoding would work in current state.
What makes you say that? My fix for for iconv was merged, https://github.com/tvheadend/tvheadend/pull/1507 which should fix the issue on musl. The start-up tests are passing, so where do you feel the problem lies?

> In the end this PR creates yet another fork/variant that is unsupported and untested.
Well you'll end up having tons of forks, the further you get from something supported upstream. This is just the first step. "Untested and unsupported" is a bit harsh though. I do agree that it would be good if we can have some automated tests as part of the pipeline though. E.g. build the container, run the container, run some test cases to validate it is fine.

How is ls better 'supported'. I will agree though that there image is certainly better tested, as it has more users.

>  As user you still want to use the ls.io docker because its much more feature complete, so you basically just lurk users to use the "official" docker that has no idea about the situation.
I'm not lurking users into anything. You can call it whatever you want, you can also all 'ls.io the best container in the world', nobody is forcing anything :)

As for feature completeness, here comes documentation and examples in. The features they offer, for example like picon support, is best handled in a nice compose file, which an example can surely offer, but are out of scope for a pure (and proper) hts container.

Also, rome wasn't built in a day :) it can have a 'testing/experimental' tag until its on-par, yeah?

Flole998:
To be honest I don't like containers and my knowledge about operating them is close to 0 (even though I was forced to use Kubernetes a while ago). Others love them, that's fine with me as long as I don't have to support them. I've seen quite a lot of issues of people who had used that ubuntu snap stuff and got "permission denied"-errors when starting a recording and so on.

As I don't really know whether what's going on here is good or bad (I assume it's good as the GitHub actions are running through), I'll leave it to you guys who seem to know a lot about containers to figure out what the best way is and what should be done here.

CvH:
@oliv3r 
> First of all, ls pulls in the kitchen sink, and then some, compile stuff that are 'nice to have' etc.

I agree picons and comskip are not really necessary.
So you could just remove those lines from the Dockerfile and thats it and you got a well tested tvheadend docker that has **proven** to work well. 

> Also, rome wasn't built in a day :)

Thats correct, but rome was already build at ls.io with help of several docker experts.
I can really just advise to not reinvent the wheel and throw away all knowledge that was collected over several years just to start from scratch again and repeat all mistakes. 

@Flole998
> and my knowledge about operating them is close to 0

There is a reason why the ls.io dockers are so popular, they know what they are doing !
At dockerhub are a lot shitty images around from ppl that are think they know how docker works. Its not that easy and you need experience and most of you need time to deal with it and to learn how it works. 
To be realistic nobody here has enough time for that :)

oliv3r:
> @oliv3r
> 
> > First of all, ls pulls in the kitchen sink, and then some, compile stuff that are 'nice to have' etc.
> 
> I agree picons and comskip are not really necessary. So you could just remove those lines from the Dockerfile and thats it and you got a well tested tvheadend docker that has **proven** to work well.

But that's what the basis is, I didn't fully ignore their Dockerfile ;)
I also actually fixed (as linked before) the iconv stuff, so that line can actually be _removed_. The additional (unused?) dependencies in the final container are not there, because we are not depending on them, ldd doesn't show them, alpine doesn't show them, so adding stuff that's not needed is not great is it?

> 
> > Also, rome wasn't built in a day :)
> 
> Thats correct, but rome was already build at ls.io with help of several docker experts. I can really just advise to not reinvent the wheel and throw away all knowledge that was collected over several years just to start from scratch again and repeat all mistakes.

But I didn't, what makes you feel like I did?

> 
> @Flole998
> 
> > and my knowledge about operating them is close to 0
> 
> There is a reason why the ls.io dockers are so popular, they know what they are doing ! At dockerhub are a lot shitty images around from ppl that are think they know how docker works. Its not that easy and you need experience and most of you need time to deal with it and to learn how it works. To be realistic nobody here has enough time for that :)

Yes, there are lots of shitty ones, but keeping a container (with input from experts) close to the source usually is best. While one does not need to have container expertise, the author does know how to build their application best, yeah? So LS.io knows containers, tvheadend knows tvheadend. If that means I'll have to support the container, so be it. I'm a user of the container as well. I have PCI and USB devices so can ensure those work. Permission, packaging etc issues, are always a thing, when people deviate from the known path, that will always require support. Case in point, the previous link to the forum, where a guy has trouble running his own guide. He IS using LS.io container though. Also, Its not fair to dump all the support load on LS.io is it? They are just volunteers as well. Or are you suggesting LS.io should be 'the one true home for all containers'? A container distro? Noble goal, but is it feasible with just volunteers?

Flole998:
Well I've always handled it as "contact the maintainers for  support/issues with unofficial/modified builds" as they are the ones who changed stuff so they know their changes. There is no real support for Tvheadend anymore, so the maintainers of those builds are the best bet to get support. They can of course choose to ignore support requests and bug reports. Sometimes maintainers just do changes that are bad, like disabling a check so an issue isn't reported, later on Tvheadend would crash in that case and as we have a lack of resources anyways nobody has time to look at all those scary changes some maintainers like to do and what issues they might cause. Even just adding empty lines causes line numbers to be not matching, that's annoying when analyzing a log aswell.

mralapete:
It’s quite noticeable that some newcomers to the TVH project have this vision of a project team in place constantly updating the project. Unfortunately that is not the case and hasn’t been for quite some considerable time.

Maybe the skill sets being deployed on the subject of containerisation could be put to better use on the core project. Getting TVH to work in a container for the sake of containerisation may not be the way forward at this time although I’m sure anyone who has achieved this will continue to pursue this route on a personal level.

oliv3r:
> It’s quite noticeable that some newcomers to the TVH project have this vision of a project team in place constantly updating the project. Unfortunately that is not the case and hasn’t been for quite some considerable time.
> 
> Maybe the skill sets being deployed on the subject of containerisation could be put to better use on the core project. Getting TVH to work in a container for the sake of containerisation may not be the way forward at this time although I’m sure anyone who has achieved this will continue to pursue this route on a personal level.

So I've been using TVHeadend full-time for 12 or 13 years now? Must be, because my oldest _merged_ commits date from around that time, so maybe a little longer even :)

As for containerization, I think it's a key-deployment solution, for more 'natural' pickup of the project. What distribution methods do we have?

* tarball (source code). Which is probably the worst, users downloading source code, compiling it with god knows what compiler, compiler settings, weird distro's with different permissions.
* pre-compiled binaries, great if it works, but integration into the host can be difficult (see below)
* Distributions packages. Quite a bit better, as they also take some responsibility of what they do (in an ideal world). But distro packagers are more about 'integration' then anything else, it's probably 90% of their work. E.g. fix the examples so they are more applicable to their host, use the proper startup scripts (if none are available) add desktop shortcuts (if non are available) etc. Projects can lighten the burden by doing some of this work for them of course (and take control) but supporting all these distro's is a nightmare to say the least (e.g. having things for debian, redhat, alpine etc all in place).
* Snap, lets not go there :p (jk) [0]
* flatpak, the same as the snap above really, and numerous other similar methods of distribution. The same thing happens as with distro packages, but at least now the user gets a whole functional 'os-tree' if you will, with all libs and what not included. So a potentially miss-configured system (think gentoo with broken builds) won't stop you from running the snap/flatpak
* docker, similar to the above, but more aimed at servers to, with one (potential) difference, is that the container maintainer, is also responsible for actually compiling tvheadend. You get tagged releases and 'master' releases etc. The 'host OS' (in this case alpine) you rely on of course, but tvheadend can offer the best possible experience, turn on/off settings that are known working/broken etc and ensure proper distribution. It is not that different from flatpak/snap distribution to be honest, but imo those are more desktop technologies.

So looking at these distribution options, especially with regards to support issues, the first question I'd ask "are you running the official tvheadend docker container (even to only test things). Yes, absolutely, there are some tricky bits to be aware of, in how to 'forward' your devices into the container (though in the case of 'does it work' I'd say, use `--privileged --network host` to avoid having to do the hard bits.

Ironically, I was having a very _similar_ discussion with the clamav guys. The merge request was open for more then a year (my bad for being slow) where the maintainer was very much against the idea, didn't see the value of it etc. A year later (after some pings) he played with it, and was blown away and 180 degree's change his mind. cisco fully accepted the dockerized solution of clamav. "It just made a lot of sense". Also having the clamav database included (depending on the used tag) made it so they saved significantly in virus database updates. https://hub.docker.com/r/clamav/clamav is now upto 10M+ downloads.

Anyway, especially with a project spread so thin on developers, I think offering a container, is probably the best way forward, as you can, as I wrote, just first ask 'does it run in the container', which removes all of the issues listed above.

For me personally? I have my pet-server, that I like to keep organized. I've run tvheadend natively for 13 or so years now, and while it works fine, the rest of the system needs an overhaul, putting it in a container, I don't have to worry about dependencies, upgrades of libraries etc anymore. It just works. Next step, is to replace the host os (gentoo, that is foobared for the most part) with something sensible. And all I gotta do, is remount my docker partition int he new host, and kick the containers up :) So no, this is far from an idealization of containerization :) In fact, a few years ago, I was heavily against it :p irony.

[0]: I think one major difference between flatpak and snaps, is that snaps make heavy use of overlays. You need gtk? no problem, lets install the gtk snap, and 'connect' the containers together, wheras flatpak comes as a single package.

Containerfile
```diff
@@ -0,0 +1,126 @@
...
+                --python=python3 \
+        && \
+        make DESTDIR='/tvheadend' -j$(($(nproc) - 1)) install && \
+        rm -f -r '/tvheadend/doc'
+
+FROM docker.io/${TARGET_ARCH}/alpine:${ALPINE_VERSION}
+
+LABEL maintainer="Olliver Schinagl <oliver@schinagl.nl>"
+
+EXPOSE 9981 \
```
Saentist, line 76: :9983 for sat>ip default port + :554

oliv3r: I'm trying to figure out what this means. I see that the cmdline states that
```
      --satip_rtsp            SAT>IP RTSP port number for server
                              (default: -1 = disable, 0 = webconfig, standard port is 554)
```
So Default is no port, ok fair, 0 means that it uses the same port as webconfig, which is 9981. But then what is 'standard port'? Who/what is using 554? When is it binding to 554? if you do `--satip_rtsp 554`? Yes, the 'standard' rtsp port is 554, but nothing is using it.

Since 'expose' is really just for documentation purposes of the container, and by 'default' we don't listen on this port, I think it's best to leave it out.

Further more
```
### 1. Define the RTSP Port

This can be anything you like, it is recommended that you use 9983 
(to avoid permission issues). Entering zero (0) in this field will 
disable the server. 
```
So this basically recommends you to do `--satip_rtsp 9983` yeah? Which means the container by default doesn't use 9983, so we shouldn't expose it.

Granted, both notes should be put in the documentation :) (I'm trying to find a place to put it though).

dreamcat4:
As somebody who (much earlier) came out with an older / initial docker image... I have today reviewed and checked what linuxservers team are doing with their 3rd party tvheadend image. And it's pretty comprehensively great.

So I don't see much issue with that one. Or much reason(s) to criticise it. The only thing that I personally am a bit dubious about (and it's more of a general observation across all of docker). Is the reliance upon alpine / muscl based base images. Rather than a glibc... but another concern (again it's just my own personal / individual opinion speaking here)... was that this week one of the major package maintainers for alpine linux has now said that they are stepping away from their role. So I do remain a bit uncertain myself. However since it works, and since there has now been such a sustained momentum behind using alpine linux in docker images... my concerns probably can be dismissed. They were more historically based on an earlier viewpoint(s).

So I can highly recommend the linuxservers tvheadend docker image. Cannot really seem to find much faults with it really. And they seem to be providing a very high quality level of support (that 3rd party team). Including recent updates that adds practical value and fixes stuff. So that one is all great, very impressed.

Actually (today) I have been looking at updated my own (broken) docker image. And that was going OK until i hit the picons thing. Which isn't so great to deal with TBH.

Anyhow in termos of @oliv3r PR here... he seems to be doing something generally similar to the linuxservers. However of course it relies on the github infrastructure to provide. So there is a sort-of a vendor lockin to acknowledge that these builds depends on github's service.

So: for these container builds there's both a dependancy on alpine linux, and also a dependancy for github infastructure. Personally I think that is probably OK myself, so long as these container builds are considered to be as a value add.

In terms of being the upstream tvheadend project: It's also acknowledged that there isn't such good support anymore. And I actually believe **this is exactly why** we should accept this PR and bring @oliv3r onboard here. Because having created those container build, he is the person best placed to continue to maintain them. So @Flole998 doesn't need to worry about giving support for any of that. Nobody really needs to think that the upstream is extending official support to include container builds...

But here is why having @oliv3r here included now in upstream activities I think is a win (for upstream):

Because of 1 critical reason; which is that when we take regular PR's from anybody else to upstream there is the github CI/CD pipeline (auto testing) feature. And so we already need (and have in the `.github` subfolder) those github 'workflows' for building tvheadend in github... these are already OCI containers! So we are already needing to have those maintained to use the github CI/CD tool. Therefore with a greater involvement from @oliv3r he can also be well positioned to help maintain those too. Thus increasing the overall level and quality of long term support for the upstream project.

Now this is dismissing some concerns that people will end up using 'more broken' containers. And so on. However I feel those concerns really are more about not knowing what people are using or platforms they have their installs... but looking around on the internet (randomly) a lot of people tend to just have tvheadend installed natively onto distros like libreelec. Or synology (as a synology container). And it's usually fairly aparrent their environment or can be easily asked / requested.

Finally the whole "not seeing tuner hardware" within a container... it's been well establisted the hardware passthu flags in docker. That the user needs to supply some flags like `-d /dev/path/to/device(s)`... this is certainly an issue specific to docker. However I do not feel it needs to negatively impact the upstream. And by growing our comminuty and accepting people into it with experience for containers. Those broader support ecosystem should not be lacking to help those individual users identify better whether they are dealing with a container configuration issue(s) versus some genuine upstream issues.

So my overall assessment is: we should accept @oliv3r 's PR. For the above reasons. And consider that while the PR is accepted, it's not an official (nor guaranteed) upstream support. And therefore not to be promoted unduly, nor necessarily promoted (above) the 3rd party linuxserver tvheadend image.

Hopefully the aim of my comment is taken not to undermine anybody else's opinions here. It's just my own assessment having had to maintain my own container image (which has been around for a long time). That if you are not familiar with containers to take on a sort-of trust level basis.

Merging this code isn't going to break anything else in the upstream. And if @oliv3r dissapears then ok it breaks / gets broken later on... not really much harm in that. So those are all my feelings and why I generally support this MR.

mralapete:
From my initial reading of your post you’ve put together a substantial case in favour of TVH containers. The problem I see here is that you need to bring this all to the TVH forum for debate. 

It appears to have become fashionable to completely bypass the forum and as everyone knows the GitHub is not the ideal place for discussion. Keeping the forum relevant where everyone can contribute with suggestions instead of the handful now doing so will help greatly. 

I strongly recommend that a section in the forum be established where developers/would-be developers could debate these issues openly and form some sort of consensus on how the TVH project could proceed instead of being buried in the Issues section of Github. 

I’m in no doubt there are some very capable coders out there who could contribute in a meaningful way given the right conditions. As it stands from what I can gather the original crop of developers no longer contribute to the project for whatever reasons. I may be wrong here. 

It would be a shame to see the TVH project disappear completely knowing that there may be others with meaningful contributions out there ready to contribute.

CvH:
> The problem I see here is that you need to bring this all to the TVH forum for debate.

Adding a fast, simple and effective way to contribute is key for projects that are not driven by a "collective/person". Adding mandatory discussions at a forum you have separate to register and invest time to mange adds a hurdle ppl will likely scare away rather then invite to participate.

> everyone can contribute with suggestions

Just to be honest I am pretty sure nobody really likes to discuss coding stuff with ppl that have zero idea all about :) 
For the target audience github is a daily tool most of them are familiar. Usually a forum is a user centric support board. Github is rather the place where the development happens. Not saying you can't do both, but to be realistic most devs are usually short on time.

Flole998:
Since apparently @oliv3r is going to maintain this at least for some time and this isn't a "dump it here and then someone else will take care of this" I am going to merge this once I got all the CI issues out of the way.

oliv3r:
@Flole998 appearantly :)

Lets get a few of the other MR's in; let those stabalize for a few days (and I'll rebase this MR of course). Since I'm a 'heavy' TVH user (we watch TV every day) i'll use the updated container on my system for a bit to notice any breakage first.

oliv3r:
> > The problem I see here is that you need to bring this all to the TVH forum for debate.
> 
> Adding a fast, simple and effective way to contribute is key for projects that are not driven by a "collective/person". Adding mandatory discussions at a forum you have separate to register and invest time to mange adds a hurdle ppl will likely scare away rather then invite to participate.
> 
> > everyone can contribute with suggestions
> 
> Just to be honest I am pretty sure nobody really likes to discuss coding stuff with ppl that have zero idea all about :) For the target audience github is a daily tool most of them are familiar. Usually a forum is a user centric support board. Github is rather the place where the development happens. Not saying you can't do both, but to be realistic most devs are usually short on time.

I suggested to maybe migrate to *gulp* github discussions instead of the forum to avoid at least that problem.

mralapete:
In light of recent “developments” with the TVH project would it be reasonable to incorporate this PR now. Hopefully @oliv3r is still about to assist with this.

oliv3r:
I'm on holiday until the end of the month. I've been using the container since this thread, so with some final cleanups it should he good to go

mralapete:
> I'm on holiday until the end of the month. I've been using the container since this thread, so with some final cleanups it should he good to go
> […](#)

That’s great news. I’d hate to see your hard work and endeavours go by the wayside. I look forward to the end results.

rkkoszewski:
Just as feedback. The docker file in this PR for Alpine (But not Debian) is installing Intel GPU drivers. But what if the host has an Nvidia or AMD GPU? Also, I've come across the situation where the host OS was updated including kernel and Intel GPU drivers, and the GPU drivers inside the container were completely broken due to compatibility issues and would give all kind of cryptic errors. I'd suggest to keep GPU transcoding out of the equation and allow people to extend this docker image to add the drivers they need for their specific system. There are a ton of permutations in regards to this.

oliv3r:
> > I'm on holiday until the end of the month. I've been using the container since this thread, so with some final cleanups it should he good to go
> > […](#)
> 
> That’s great news. I’d hate to see your hard work and endeavours go by the wayside. I look forward to the end results.

I haven't found about the 'recent developments' what did you mean?

I just pushed whatever I had locally, I'm trying to see if I forgot anything. I've been using this dockerfile for 3 months now without any issues. I did say it needed some cleanup, maybe I had already done it just didn't push it?

Also with regards to #1594 I take it we'd still be cloning it as part of the make step (which is fine, the packages tend to be way to outdated). Leaves a 'what about runtime updates' question though ...

So from my pov; I think it's good to go.

Flole998:
I think I need to add some secrets (and possibly create accounts?) somewhere before merging this so uploads are done automatically?

Btw: If you want you can also build the container as part of the CI process (without uploading of course).

djbloc:
Request whether we could also get a container that runs a debug symbols build of TVH?

Would provide a means to provide debug logs for issues that cause TVH to crash.

Flole998:
That would certainly be a good idea, maybe this is already a build with debug symbols?

Once I know what I need to do and set for this to work I can do so and merge it

oliv3r:
> I think I need to add some secrets (and possibly create accounts?) somewhere before merging this so uploads are done automatically?
> 
> Btw: If you want you can also build the container as part of the CI process (without uploading of course).

I don't think so, in that when pushing to github's register, internal tokens work just fine. The pipeline already pushes it to the repo that runs the pipeline, see here: https://github.com/oliv3r/tvheadend/pkgs/container/tvheadend

If we want to _also_ push to docker hub; yes, an FOSS organization account on docker hub is required, and a token is needed in a variable. I think the pipeline can be easily extended to _also_ push to docker hub.

I'd do that at a later time however, let it land here first, and let users pull the ghcr.io image for a while. Once that's stable and people are happy; we can consider pushing it to docker hub for a potential wider audience.

oliv3r:
> Request whether we could also get a container that runs a debug symbols build of TVH?
> 
> Would provide a means to provide debug logs for issues that cause TVH to crash.

Doing a -debug compile of tvheadend is an idea, we'd build and push the image as :debug (though imo, we can also add debug symbols to :dev builds, and without debug symbols on the tagged builds.

Enabling debugging logs is easy regardless.

BUT! and that's a big BUT! I'm not a big fan at all of publishing debug symbols in builds. YES it helps in debugging applications, but it also changes runtime behavior in unexpected ways. 'But it works/fails with the debug build' is something you want to avoid.

afaik tvheadend does stack traces on crashes doesn't it? libunwind support is already here isn't it? If not, I think that's probably more useful then debug symbols. YMMV of course.

A nother option would be to include the debug symbols separately, as most linux distro's do; but building that, that's new territory for me :p

oliv3r:
rebased :)

oliv3r:
> Could you please squash your commits? Apparently you didn't do a rebase last time but instead merged master back into your branch.

I think github does that when you push the button 'rebase'. That's so dumb. I would never merge, i"m  a rebase person.

As for the squash, no I will not :p I will rebase which should leave us with a single clean commit :)

mralapete:
Finally added and great to see that.

Flole998:
And the build failed as it got a 403 when trying to publish :(

oliv3r:
Aww, that sucks :p ok let me double check.

Building from the current repository automatically uses the [GitHub Token](https://docs.github.com/en/actions/security-guides/automatic-token-authentication), so it does not need to be passed. is from the docker-action. Which is why we don't see any authentication related things in the workflow file.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication states it is automatically generated 'when you install the action'.

I wonder if we need to enable something on the repo?

Flole998:
> In my fork, I see the following under: https://github.com/oliv3r/tvheadend/settings/actions
> 
> image could that be it?

Nope, same settings here.

Flole998:
Looks like a simple re-run apparently was enough to fix it. Can someone please test if it works?

oliv3r:
> Looks like a simple re-run apparently was enough to fix it. Can someone please test if it works?

Weird. 403 is usually authentication error. Lets see if this keeps comming or not.

Meanwhile, I'll pull this container and use it as my primary one :)

oliv3r:
@Flole998 might still be some repo configuration needed.

If I look at https://github.com/oliv3r/tvheadend/pkgs/container/tvheadend, I can see the containers, however https://github.com/tvheadend/tvheadend/pkgs/container/tvheadend I get an authentication denied error. Also on the main repo page, on the right I have 
image
but on the tvheadend, I do not see the packages.

So could still be some settings needed. Also, and most importantly:
```console
$ docker pull ghcr.io/tvheadend/tvheadend:master
Error response from daemon: Head "https://ghcr.io/v2/tvheadend/tvheadend/manifests/master": unauthorized
```

Flole998:
I can access both URLs, yours and the "official" one. I can also see packages on the official one. I'll try without being signed in next... I can also contact GitHub support and ask them what we're doing wrong.

ProfYaffle:
It would seem to be that the tvheadend repo is Private:

image

Flole998:
Ahh, all the way at the bottom in the "Danger zone", where usually only deletion of repository and other stuff nobody needs is located, the option to make it public is hidden. Done now.

oliv3r:
I managed to pull the container now :)

I don't see the 'pkgs' on the main page on the right side, but I don't see the 'releases' there either. Which is strange, but not an issue for now.

Lets see the small number of people that had interest in this, and then a) put it in the visible area on the main page (readme and whatnot). Then, let it expand a bit in the user base.

Once that is good, see if we can also push to docker hub, and add a nice docker 'pulls' badge :)

oliv3r:
Btw, the only thing I've noticed so far, may be just a cosmetic bug,

```
tvheadend-1  | 2024-02-11 22:40:58.345 [   INFO] config: backup: migrating config from 4.3-2205~g4be3995c2 (running 0.0.0-unknown)
tvheadend-1  | 2024-02-11 22:40:58.346 [   INFO] config: backup: running, output file /var/lib/tvheadend/backup/4.3-2205~g4be3995c2.tar.bz2
```
No idea why its stating it's running `0.0.0-unknown` but `4be3995c2a1d` is certainly the hash of `my` build. Technically both builds should be identical, though my build is a few months old. So `git describe` is in the CI producing something weird. Obviously the builds happen identically, both in docker. I'll check this out at some later point though; for now, this is purely cosmetic, but may lead to issues for sure.

Flole998:
No changes there. You need to checkout the whole repo, including all tags. I remember having that issue with some other GitHub action. Check the native builds to see how I fixed it there.

Also you should pass the satip server ip as the automatic detection will fail miserably in a container I guess.

oliv3r:
> No changes there. You need to checkout the whole repo, including all tags. I remember having that issue with some other GitHub action. Check the native builds to see how I fixed it there.

Right! Of course. I know in gitlab you have to set the 'depth' as otherwise, in SOME cases, git describe will fail. The default clone depth in gitlab is 50 iirc, and for tag related jobs, I always set the depth to 0. I'll take a peak how to do this with github actions on clone, should be copy/paste from your stuff. Sorry for missing that.
 
> Also you should pass the satip server ip as the automatic detection will fail miserably in a container I guess.

It depends, are you talking about tvheadend being a client? Something certainly to put in the documentation. Something I do not use so no expert here. For those googlers, if the container is run with `--network host` you just map your network into the container adverbaitim. Not desirable from a security point of view, but should get you out in a pinch. 

I don't know how 'auto detection' works normally though. So something is definitely needed when running tvheadend with satip (I run it with -nosatip :p)

Flole998:
There's code in Tvheadend that opens a socket to 8.8.8.8 and gets the local source IP for that socket and assumes that the server is reachable under that IP. I am working on changing that (so it doesn't use 8.8.8.8 but the actual client), but that won't work through NAT either.

oliv3r:
Spent more typing here then fixing, which was indeed trivial (assuming its correct and all that is needed). #1639 :p

oliv3r:
> There's code in Tvheadend that opens a socket to 8.8.8.8 and gets the local source IP for that socket and assumes that the server is reachable under that IP. I am working on changing that (so it doesn't use 8.8.8.8 but the actual client), but that won't work through NAT either.

So by 'default' docker containers are launched as/in a NAT setup. Can be bypassed with `--network host` for now I suppose. Not ideal, but I guess TVHeadend would have to deal with being behind a NAT then.

Flole998:
That's why there is an option to pass the IP it's reachable on manually to Tvheadend. I've used kubernetes before and I think there was some variable to specify under which IP the container is reachable? Does that exist for docker? Can it be passed as an argument to Tvheadend on launch?

djbloc:
> So by 'default' docker containers are launched as/in a NAT setup. Can be bypassed with `--network host` for now I suppose. Not ideal, but I guess TVHeadend would have to deal with being behind a NAT then.

Just to add, hdhomerun requires --network host. I believe it uses UDP broadcast to identify tuners.

Personally don't think it is an issue running with network host, I would imagine most users of TVH are running it locally on hosts they own/operate.

oliv3r:
So first off, #1639 does indeed fix the version issue :) so no more issues there ;)
```
tvheadend-1  | 2024-02-12 09:10:08.439 [   INFO] config: backup: migrating config from 0.0.0-unknown (running 4.3-2229~g6b5defc76)
tvheadend-1  | 2024-02-12 09:10:08.439 [   INFO] config: backup: running, output file /var/lib/tvheadend/backup/0.0.0-unknown.tar.bz2
```

> Personally don't think it is an issue running with network host, I would imagine most users of TVH are running it locally on hosts they own/operate.

I fully agree that the risk is minimal, and that this is a choice/option for certain users.

> That's why there is an option to pass the IP it's reachable on manually to Tvheadend.

Who is 'it' in this context? Anyway, if we need to tell the 'outside' IP to the 'inside', There is of course always the question of 'which IP' if the host has multiple IP addresses. But to get the 'gateway' of the container, the IP of the host, is to add the following to the docker container (in compose this would be)
```yaml
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

But this is only the hostname. Can tvheadend resolve this itself? Probably not.

Regardless, you can always do something ugly-ish, like
```
docker run --env hostip="$(ip get-IP | sed-magic)" tvheadend --satip "${hostip}"
```

But in the end, this is mostly all about (improving) documentation :) though I'm a little hestitant to write it, as I cannot test it. I have no satip equipment (unless I can use 2 tvheadend instances, client and server, to talk satip with eachother).

PR #1533, opened 2023-06-05
Update config for Fedora 37/38

rubdos:
I'm not sure whether Doozer is still relevant, now that RPMFusion is shipping rather recent versions.

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -583,9 +583,9 @@
       ]
     },
     "fedora34-x86_64": {
-      "buildenv": "docker:fedora:34",
+      "buildenv": "docker:fedora:37",
       "builddeps": [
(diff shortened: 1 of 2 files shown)


Flole998:
Doozer is no longer used. If you want you can try to get GitHub actions to work for Fedora, then we can publish the RPMs to cloudsmith where all the Debian packages are aswell.

rubdos:
> try to get GitHub actions to work for Fedora

Attempt in cb86eb8.

rubdos:
They don't even seem to start :-/

How do we find out the reason? I'm more used to Gitlab CI which is quite a bit more verbose.

Flole998:
That's because "fedora-latest" is most likely not a valid image name that GitHub actions can use. There is a list somewhere and you need to choose one of the options from that list, otherwise it will just wait forever until the specified one becomes available (which might never happen).

PR #1535, opened 2023-06-10
Add XDG support

oliv3r:
This series adds support for the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1698,7 +1698,16 @@ static char *config_get_dir ( void )
   char hts_home[PATH_MAX + sizeof("/.hts/tvheadend")]; /* Must be largest of the 3 config strings! */
   char config_home[PATH_MAX];
   char home_dir[PATH_MAX];
+  uid_t uid = getuid();
   struct stat st;
+
+  snprintf(hts_home, sizeof(hts_home), "/var/lib/tvheadend");
+  if ((stat(hts_home, &st) == 0) && (st.st_uid == uid))
+    return strndup(hts_home, sizeof(hts_home));
+
+  snprintf(hts_home, sizeof(hts_home), "/etc/tvheadend");
+  if ((stat(hts_home, &st) == 0) && (st.st_uid == uid))
+    return strndup(hts_home, sizeof(hts_home));
 
   if (realpath(getenv("HOME"), home_dir) == NULL) {
     tvherror(LS_CONFIG, "environment variable HOME is not set");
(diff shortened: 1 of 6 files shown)


oliv3r:
Sorry, I hadn't, with codeball always failing, it's hard to miss (I was hoping we could fix codeball with the installation, I'll look at it again, but unfamiliar with codeball itself).

But, lets fix it :)

oliv3r:
> I think codeball might be broken

yeah, probably.

Anyway, I couldn't reproduce it with the :latest variant (of both alpine or ubuntu), but on my local arch compiler it did barf, it's a new (good) feature of gcc! I'm supprised it's not triggering on other locations, but regardless, it's fixed.

oliv3r:
> Now it's failing with a different error, it's missing 1 byte for the null-Termination apparently.

Yeah i saw; if these API's where just a bit more consistent with eachother, anyway, lets see what happens in the next 15 minutes :p

oliv3r:
So I found why I can't reproduce it, my Makefile doesn't contain ` -Wstringop-truncation`, I don't find it with git-grep either, so somewhere something is addiding this flag (something like `-Wall`? I haven't looked into detail. But these hidden/explicit things are not that great.

Personally, I"m a huge proponent to catch as much as we can with these kind of build flags. Where can I find these definitions, and can we not just add them to the main Makefile?

Flole998:
There's `-werror` somewhere and there's a configure option `--nowerror` or something like that, maybe that points in the right direction? I remember that it was turned off for some builds to allow them to compile, especially since it got added to ffmpeg aswell and broke compilation there when using the static ffmpeg option.

oliv3r:
Success; I"ll test it a little bit more with the fix, but gotta fix my containers a little bit first ;) But a review would be welcome in the meantime.

Flole998:
Looks good to me, is it ready for merging?

oliv3r:
@Flole998 yep, I've tried a few scenario's and they all seem to use the correct path. So lets merge it :)

oliv3r:
@Flole998 ah shit sorry, I forgot to push my very last, which means I committed some white-space snafu and forgot a commit. I'll open a new MR to fix those. This MR isn't broken or anything though.

Flole998:
Could this be the cause of https://tvheadend.org/issues/6296?

oliv3r:
@Flole998 I would not be supprised, let me try and reproduce it, I hadn't used `-f` during my testing

PR #1536, opened 2023-06-10
Add simple 'ping' endpoint

oliv3r:
To determine if the server is online, we want to be able to 'ping' it with as little configuration and access possibilities as possible. The only available option for this today, is anonymous access.

Let us thus add a new endpoint, that replies with 'PONG' when queried.

This will need anonymous access enabled, but can be restricted to: a) localhost
b) without any access permissions

A simple way to 'test' this, is with busybox's wget and grep.
```
wget -O - -q http://172.17.0.5:9981/ping | grep -q PONG
```
which should return 0 on success, 1 otherwise.

### DIFF
--- a/src/webui/simpleui.c
+++ b/src/webui/simpleui.c
@@ -17,6 +17,7 @@
  */
 
 #include "tvheadend.h"
+#include "htsbuf.h"
 #include "http.h"
 #include "webui.h"
 #include "access.h"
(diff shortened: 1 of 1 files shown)


oliv3r:
@Flole998 I think you need to install https://github.com/apps/codeball-ai-writer to this repo to make 'codeball' work. I installed it to my repo's, but that doesn't fix the issue.

Flole998:
It's installed but still doesn't work

PR #1537, opened 2023-06-12
No need for a Fedora host for fedora images

rubdos:
Follow-up for #1533; I did not realize that the Ubuntu image was a VM, not a container.

@Flole998 dixit

> That's because "fedora-latest" is most likely not a valid image name that GitHub actions can use. There is a list somewhere and you need to choose one of the options from that list, otherwise it will just wait forever until the specified one becomes available (which might never happen).

Could you run this pipeline before merging?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,18 +191,21 @@ jobs:
           CLOUDSMITH_API_KEY: ${{ secrets.CLOUDSMITH_KEY }}
 
   build-rpm-native:
-    runs-on: fedora-latest
+    runs-on: ubuntu-latest
     continue-on-error: true
-    name: Build on native ${{ matrix.container }}
+    name: Build on native fedora:${{ matrix.releasever }}
     strategy:
       matrix:
-        container: ["fedora:37", "fedora:38", "fedora:39", "fedora:rawhide"]
+        releasever: ["37", "38", "39", "rawhide"]
     container:
-      image: ${{ matrix.container }}
+      image: "fedora:${{ matrix.releasever }}"
     steps:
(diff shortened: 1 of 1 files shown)


Flole998:
I don't think I can, but you can turn on github actions for your fork, then you can run it (just expect all the uploads to be failing due to the API key missing).

rubdos:
Trying at https://github.com/rubdos/tvheadend/pull/1

Flole998:
It's filtered to only run on master, so you'll have to apply the changes to your forks master, then it will run. Otherwise it will be the same thing as in this repo.

rubdos:
Ack, then we await here and I clean up my master branch afterwards: https://github.com/rubdos/tvheadend/actions/runs/5243125285/jobs/9467408255

rubdos:
https://github.com/rubdos/tvheadend/actions/runs/5252234584/jobs/9488048568 this looks promising. If it gets through, I'll cherry-pick and squash the fixups..

Flole998:
Please squash your commits, then we can get this merged and I can check what's the best way to get the packages to upload.

Flole998:
Something like this should do it (it requires the file to be called tvheadend.rpm though, so we need to copy/rename it):

```
      - name: Push
        id: push
        uses: cloudsmith-io/action@master
        with:
          api-key: ${{ secrets.CLOUDSMITH_API_KEY }}
          command: "push"
          format: "rpm"
          owner: 'tvheadend' 
          repo: 'tvheadend'
          distro: 'fedora'
          release: ${{ matrix.releasever }}
          file: "tvheadend.rpm"
```

PR #1539, opened 2023-06-16
spawn: Do not close every possible file descriptor

oliv3r:
When close is called with a non-existant file descriptor, it will happily do so. Since we do not even check for the error code, trying to close a non existant/not-open file descriptor does not even cause an error.

This in itself is not a problem at all, however, we try to close every open file descriptor beyond stderr, upto whatever _SC_OPEN_MAX returns.

Some systems may have a very large ulimit set for `_SC_OPEN_MAX` and thus `maxfd` may return something in the millions*. This means that the simple `for() close();` loop now has to iterate a million file descriptors which may not even be open/exist.

Instead, we can use the power of `/proc/self/fd` to list our open file descriptors, and only close open ones.

*It can actually even be a billion, as it is for the root user. While nobody should be running tvheadend as root, doing so will trigger this issue for sure. Also all users that inherit ulimits from root.

### DIFF
--- a/src/spawn.c
+++ b/src/spawn.c
@@ -459,7 +459,7 @@ spawn_and_give_stdout(const char *prog, char *argv[], char *envp[],
                       int *rd, pid_t *pid, int redir_stderr)
 {
   pid_t p;
-  int fd[2], f, i, maxfd;
+  int fd[2], f, i;
   char bin[256];
   const char *local_argv[2] = { NULL, NULL };
   char **e, **e0, **e2, **e3, *p1, *p2;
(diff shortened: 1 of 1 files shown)


Flole998:
I think `/dev/fd` is a more portable solution. Also you could (and probably should) implement a fallback to the old method in case the list is not available (maybe even with a warning).

oliv3r:
> I think `/dev/fd` is a more portable solution. Also you could (and probably should) implement a fallback to the old method in case the list is not available (maybe even with a warning).

`/dev/fd` is just a symlink; but yeah, it's probably better. I'll revert to the old method in case of problems; it's not a bad idea

PR #1538, opened 2023-06-16
config: Store recordings sensibly on servers

oliv3r:
For server configurations, we want to store recordings somewhere sensible.

While here, lets 'abuse' this merge request to fix the following:

The internal print functions already add the newline for us, so adding one manually is not needed.

Further more, a tab got snook in, where spaces where intended.

This fixes commit dbf973307ae3 ("dvr_storage: Use XDG spec directories") which accidentally introduced this.

PR #1540, opened 2023-06-21
Update Debian packaging to use the new configuration directories

mamarley:
This updates the Debian packaging to use the new configuration directories introduced by
https://github.com/tvheadend/tvheadend/pull/1535 and https://github.com/tvheadend/tvheadend/pull/1538.  Instead of being in /home/hts/.hts/tvheadend, the configuration will now be stored in /var/lib/tvheadend by default, which is consistent with how many other "server"-style Debian packages handle their home directories and configuration.

It also fixes the logic in config.c:config_get_dir() to properly support the forking mode of operation.  Since config_get_dir() is executed before forking, the uid will always be 0 at this point. Instead, use the uid of the user to which we will fork if a fork will occur.

### DIFF
--- a/debian/tvheadend.postinst
+++ b/debian/tvheadend.postinst
@@ -10,16 +10,14 @@ configure)
 
    if ! getent passwd $HTS_USER >/dev/null; then
         echo >&2 "Creating user: $HTS_USER..."
-        adduser --quiet --system --group --shell /bin/bash $HTS_USER
+        adduser --quiet --system --group --home /var/lib/tvheadend $HTS_USER
    fi
 
    HTS_HOME=`getent passwd $HTS_USER | cut -d':' -f6`
 
-   mkdir -p "${HTS_HOME}/.hts/tvheadend"
-   chown ${HTS_USER}:${HTS_USER} "${HTS_HOME}/.hts"
-   chown ${HTS_USER}:${HTS_USER} "${HTS_HOME}/.hts/tvheadend"
+   install -d -g ${HTS_USER} -o ${HTS_USER} "${HTS_HOME}/recordings"
 
-   HTS_SUPERUSERCONF="${HTS_HOME}/.hts/tvheadend/superuser"
+   HTS_SUPERUSERCONF="${HTS_HOME}/superuser"
    rm -f "${HTS_SUPERUSERCONF}"
    touch "${HTS_SUPERUSERCONF}"
    chmod 600 "${HTS_SUPERUSERCONF}"
--- a/src/config.c
+++ b/src/config.c
@@ -1693,13 +1693,15 @@ config_check ( void )
 
 static int config_newcfg = 0;
 
-static char *config_get_dir ( void )
+static char *config_get_dir ( uid_t uid )
 {
   char hts_home[PATH_MAX + sizeof("/.hts/tvheadend")]; /* Must be largest of the 3 config strings! */
   char config_home[PATH_MAX];
   char home_dir[PATH_MAX];
-  uid_t uid = getuid();
   struct stat st;
+
+  if (uid == -1)
+    uid = getuid();
 
   snprintf(hts_home, sizeof(hts_home), "/var/lib/tvheadend");
   if ((stat(hts_home, &st) == 0) && (st.st_uid == uid))
@@ -1777,7 +1779,7 @@ config_boot
 
   /* Generate default */
   if (!path)
-    config.confdir = config_get_dir();
+    config.confdir = config_get_dir(uid);
   else
     config.confdir = strndup(path, PATH_MAX);
 
--- a/src/config.c
+++ b/src/config.c
@@ -1723,7 +1723,7 @@ static char *config_get_dir ( uid_t uid )
 
       if ((readlink(hts_home, hts_home_link, sizeof(hts_home_link)) == -1) ||
           (stat(hts_home_link, &st) == -1)) {
-        tvherror(LS_CONFIG, ".hts/tvheadend is inaccessable: %s", strerror(errno));
+        tvherror(LS_CONFIG, ".hts/tvheadend is inaccessible: %s", strerror(errno));
         return NULL;
       }
       strncpy(hts_home, hts_home_link, sizeof(hts_home));
--- a/src/main.c
+++ b/src/main.c
@@ -1113,7 +1113,7 @@ main(int argc, char **argv)
   signal(SIGPIPE, handle_sigpipe); // will be redundant later
   signal(SIGILL, handle_sigill);   // see handler..
 
-  /* Set priviledges */
+  /* Set privileges */
   if((opt_fork && getuid() == 0) || opt_group || opt_user) {
     const char *homedir;
     struct group  *grp = getgrnam(opt_group ?: "video");
--- a/debian/tvheadend.postrm
+++ b/debian/tvheadend.postrm
@@ -9,7 +9,7 @@ case "$1" in
 purge)
     if getent passwd $HTS_USER >/dev/null; then
         HTS_HOME=`getent passwd $HTS_USER | cut -d':' -f6`
-        rm -rf "${HTS_HOME}/.hts/tvheadend"
+        rm -rf "${HTS_HOME}"
     fi
     db_purge
    ;;


Flole998:
Just to be sure: This is a non-breaking change as it will only apply to new user accounts created during Installation?

debian/tvheadend.postinst
```diff
@@ -10,16 +10,15 @@ configure)
 
    if ! getent passwd $HTS_USER >/dev/null; then
         echo >&2 "Creating user: $HTS_USER..."
-        adduser --quiet --system --group --shell /bin/bash $HTS_USER
+        adduser --quiet --system --group --home /var/lib/tvheadend --shell /bin/bash $HTS_USER
```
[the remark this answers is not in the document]
mamarley: My testing confirms that; I have removed the shell.

mamarley:
> Just to be sure: This is a non-breaking change as it will only apply to new user accounts created during Installation?

Correct, existing installations will be untouched both as far as the user and the TVH configuration itself.

Flole998:
@oliv3r Do you see anything wrong with these changes or does it look like a good idea to you? If it's good I'll merge it.

oliv3r:
@Flole998 i'll look asap

src/config.c
```diff
@@ -1693,14 +1693,16 @@ config_check ( void )
 
 static int config_newcfg = 0;
 
-static char *config_get_dir ( void )
```
oliv3r, line 1696: When I was testing this (in a docker container mind you) I didn't have any issue with this. I don't think this is a bad idea of course, I didn't look into this too deeply!

Regardless, this should be its own commit, not connected to the debian packaging :)

src/config.c
```diff
@@ -1693,14 +1693,16 @@ config_check ( void )
...
+static char *config_get_dir ( uid_t uid )
 {
-  char hts_home[PATH_MAX + sizeof("/.hts/tvheadend")]; /* Must be largest of the 3 config strings! */
```
oliv3r, line 1698: Why is this needed?

hts_home is either something (usually $HOME) + `/.hts/tvheadend` OR `/var/lib/tvheadend`, which fits into PATH_MAX just fine.

src/config.c
```diff
@@ -1721,7 +1723,7 @@ static char *config_get_dir ( void )
 
       if ((readlink(hts_home, hts_home_link, sizeof(hts_home_link)) == -1) ||
           (stat(hts_home_link, &st) == -1)) {
-        tvherror(LS_CONFIG, ".hts/tvheadend is inaccessable: %s", strerror(errno));
+        tvherror(LS_CONFIG, ".hts/tvheadend is inaccessible: %s", strerror(errno));
```
oliv3r, line 1726: oops :)

src/main.c
```diff
@@ -1113,7 +1113,7 @@ main(int argc, char **argv)
   signal(SIGPIPE, handle_sigpipe); // will be redundant later
   signal(SIGILL, handle_sigill);   // see handler..
 
-  /* Set priviledges */
+  /* Set privileges */
```
oliv3r, line 1116: oops again! my dyslexia combined with English being my non-native language doesn't help.

I recommend to put the two spelling fixes in one commit, the UID fix in a second commit. Personally, I'm okay with having this under the debian package MR, but that's upto @Flole998

debian/tvheadend.postinst
```diff
@@ -10,16 +10,15 @@ configure)
 
    if ! getent passwd $HTS_USER >/dev/null; then
         echo >&2 "Creating user: $HTS_USER..."
-        adduser --quiet --system --group --shell /bin/bash $HTS_USER
+        adduser --quiet --system --group --home /var/lib/tvheadend $HTS_USER
```
oliv3r, line 13: what about migration? Maybe add a check 'if old, inform user of new setup?

debian/tvheadend.postinst
```diff
@@ -10,16 +10,15 @@ configure)
...
    fi
 
    HTS_HOME=`getent passwd $HTS_USER | cut -d':' -f6`
 
-   mkdir -p "${HTS_HOME}/.hts/tvheadend"
-   chown ${HTS_USER}:${HTS_USER} "${HTS_HOME}/.hts"
-   chown ${HTS_USER}:${HTS_USER} "${HTS_HOME}/.hts/tvheadend"
+   mkdir -p "${HTS_HOME}/recordings"
```
oliv3r, line 18: `install` is better here

`install -d -m 0775 -g 'tvheadend' -o 'tvheadend' '/var/lib/tvheadend/recordings'` is what I'm using in my container atm

debian/tvheadend.postinst
```diff
@@ -10,16 +10,15 @@ configure)
...
+   chown ${HTS_USER}:${HTS_USER} -R "${HTS_HOME}"
```
oliv3r, line 19: not needed with install :p

debian/tvheadend.postinst
```diff
@@ -10,16 +10,15 @@ configure)
...
 
-   HTS_SUPERUSERCONF="${HTS_HOME}/.hts/tvheadend/superuser"
+   HTS_SUPERUSERCONF="${HTS_HOME}/superuser"
```
oliv3r, line 21: no idea what this is for, so can't judge this.

Saentist:
How current configuration will be handled by new package?
What about auto-backup during updating?

debian/tvheadend.postinst
Flole998: This won't get executed on existing installs, so there is no migration. As it isn't a benefit on existing Installations I'd assume that it can stay as-is.

debian/tvheadend.postinst
Flole998: Some limited OSs might not have the install command, so I'd leave that as is, you never know where people try to install it ;)

src/config.c
mamarley: Were you using the forking mode of operation when you tested?  The problem I encountered is that the Debian package forks (starts as root and then drops privileges), so getuid() still returns 0 when the config directory is checked.  In the case that forking is not used, the passed uid will be -1 and getuid() will then be used as you originally wrote.

src/config.c
mamarley: No problem, it happens.

src/main.c
mamarley: I don't think this one is yours; I think it was there before.

debian/tvheadend.postinst
mamarley: This particular bit of code only executes on new installs, not upgrades.  Detecting when such an upgrade occurred would be potentially problematic, so I didn't go off into those weeds.  Plus, the code you added already has a warning when the old configuration directory is in use.

debian/tvheadend.postinst
mamarley: Cool, I didn't know that was a thing.

mamarley:
> How current configuration will be handled by new package? What about auto-backup during updating?

Current installations will not be touched in any way, shape, or form, so no backup is necessary.  This change affects only new installations.

debian/tvheadend.postinst
mamarley: @Flole998 , "install" is a part of coreutils and has been since at least coreutils 7.  All supported Debian and Ubuntu versions have at least 8.21 and this code only ever gets executed on Debianoid systems, so I think we are good there.

mamarley:
Thanks for the input, everyone.  I think I have addressed all the suggestions.

src/config.c
oliv3r: Yeah, but we never do $HOME/var/lib/tvheadend :) the goal is, to have `hts_home` be large enough. Ideally, the var stays at `PATH_MAX`, and everything else gets reduced to fit, but that's far more complex :p this is much lazyer ..easier :)

So we need room to store $HOME + whatever; OR `/etc/tvheadend` | `/var/lib/tvheadend`. The later two, easily fit within `PATH_MAX`.

src/main.c
oliv3r: It's a typo I make very often though :)

debian/tvheadend.postinst
oliv3r: so true :) sorry for the noise

debian/tvheadend.postinst
oliv3r: i think `install` is a posix requirement ...

Also, this will only ever be run on debian will it not, which will guaranteed have `install` :p

oliv3r:
> How current configuration will be handled by new package? What about auto-backup during updating?

"like always", My changes keep existing setups 'as is' (with a warning).

src/config.c
oliv3r: Ah! No! I did look into `geteuid()` (afaik) but your solutions is cleaner anyway. I was testing the docker container, which will run tvheadend as <user> anyhow.

Thanks for the catch!

src/config.c
mamarley: Good point, I had failed to consider that.  I changed it back.

Flole998:
I will let the build run through and merge it tomorrow when I get near a computer :)

PR #1541, opened 2023-06-23
Revert non-portable function to previous code

johnrosshunt:
A change was made in commit dbf9773 which replaced getcwd() with get_current_dir_name() (a GNU extension).  This patch reverts that change so that tvheadend can compile on non-GLIBC systems.


oliv3r:
A little disappointment this opportunity wasn't used to use strndup or something similar, now we are using unbound checked code. I'll open yet-a-nother MR to fix this :)

PR #1542, opened 2023-06-25
NetCeiver support

walinsky:
Hi,

this adds NetCeiver support.
Sven Wegener wrote this software ( https://github.com/swegener/tvheadend/tree/netceiver-support ), and I've been using it without any problems for 5 years now.
I think it's about time to add this to the main source tree. There's thousands of users of Reelbox Avantgarde and NetCeiver who could benefit from this patch.
I've updated the code so it can be merged in the master branch.

Please let me know what you think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,6 +453,17 @@ SRCS-HDHOMERUN = \
 SRCS-${CONFIG_HDHOMERUN_CLIENT} += $(SRCS-HDHOMERUN)
 I18N-C += $(SRCS-HDHOMERUN)
 
+# NetCeiver
+SRCS-NETCEIVER = \
+        src/input/mpegts/netceiver/netceiver.c \
+        src/input/mpegts/netceiver/netceiver_discovery.c \
+        src/input/mpegts/netceiver/netceiver_frontend.c \
+        src/input/mpegts/netceiver/netceiver_hardware.c \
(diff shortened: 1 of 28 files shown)


Saentist:
> I’m using TVH for many years and have never come across Netceiver. I just did a quick search on the topic and everything appears to be in the German language. Are you sure there are thousands using it.

Same number of users with use hdhomerun, there's no problem to exist.

Flole998:
> Same number of users with use hdhomerun, there's no problem to exist.

Except that it bloats up the codebase even more

Saentist:
> > Same number of users with use hdhomerun, there's no problem to exist.
> 
> Except that it bloats up the codebase even more

Offer better code :stuck_out_tongue_winking_eye: and post PR
or merge this.

Flole998:
Or reject this to prevent the codebase from being bloated up. Those are the options available to me, and I am not sure yet which one I'll choose.

mralapete:
That’s why I never heard of it. I don’t use HD Homerun. Definitely don’t need any further bloat.

Flole998:
This would add more unmaintained code that I can't test nor maintain due to lack of hardware and lack of time, so at this point this can't be merged. You can use the pipe input option to connect to hardware that Tvheadend doesn't natively support.

PR #1543, opened 2023-07-08
Fix time for 32bit systems again

oliv3r:
In issue #6257 an issue mentioning that time_t isn't properly supported when printing on 32-bit systems, specifically on FreeBSD. However, intel 32-bit systems suffer from a similar fate:

src/rtsp.c:333:30: error: format '%ld' expects argument of type 'long int', but argument 4 has type 'time_t' {aka 'long long int'} [-Werror=format=]
  333 |   snprintf(buf, sizeof(buf), "npt=%" PRItime_t "-", position);
      |                              ^~~~~~~                ~~~~~~~~
      |                                                     |
      |                                                     time_t {aka long long int}

In commit 76a6263f1be4 ("fix for 64bit time_t on 32bit systems") was attempted to be fixed by turning it into a PRId64, which was reverted again in commit 9e1eb89be731 ("Revert "fix for 64bit time_t on 32bit systems""), sadly without a reason as to why in the commit message.

We should however, migrate to 64bit timestamps on all platforms anyway, due to the Y2038 problem. Debian is heavily working on this issue too.

This commit is just the first step, in that we ensure our time_t is always 64bits.

The next steps would be to use difftime where possible instead of subtractions, and ensure all stored timestamps have room for 64bit time_t (htsmsg_get_u32_or_default for example breaks this presumption already).

To keep this issue small, and tackle one problem at a time, lets just fix time_t first. We do still have 15 years to fix the other issues.

Note, that this patch leaves out FreeBSD specifics, as it is unclear what is specific about 32bit FreeBSD. It should be using the same glibc headers after all. If not, we can always add if needed, but adding usless code doesn't help anyone generally.

```
diff --git a/src/tvheadend.h b/src/tvheadend.h
index c2fcee716..751d10d70 100644
--- a/src/tvheadend.h
+++ b/src/tvheadend.h
@@ -334,7 +334,9 @@ void tvh_qsort_r(void *base, size_t nmemb, size_t size, int (*compar)(const void
 # endif /* ULONG_MAX */
 #endif /* __WORDSIZE */

-#if __WORDSIZE == 32
+#if __WORDSIZE == 32 && defined(PLATFORM_FREEBSD)
+# define PRItime_t "d"
+#elif __WORDSIZE == 32
 # define PRItime_t "lld"
 #elif __WORDSIZE == 64
 # define PRItime_t "ld"
 #else
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ CFLAGS  += -Wmissing-prototypes
 CFLAGS  += -fno-strict-aliasing
 CFLAGS  += -fms-extensions -funsigned-char
 CFLAGS  += -D_FILE_OFFSET_BITS=64
+CFLAGS  += -D_TIME_BITS=64
 CFLAGS  += -I${BUILDDIR} -I${ROOTDIR}/src -I${ROOTDIR}
 ifeq ($(CONFIG_ANDROID),yes)
 LDFLAGS += -ldl -lm
--- a/src/tvheadend.h
+++ b/src/tvheadend.h
@@ -334,10 +334,12 @@ void tvh_qsort_r(void *base, size_t nmemb, size_t size, int (*compar)(const void
 # endif /* ULONG_MAX */
 #endif /* __WORDSIZE */
 
-#if __WORDSIZE == 32 && defined(PLATFORM_FREEBSD)
-#define PRItime_t       "d"
+#if __WORDSIZE == 32
+# define PRItime_t "lld"
+#elif __WORDSIZE == 64
+# define PRItime_t "ld"
 #else
-#define PRItime_t       "ld"
+# error "__WORDSIZE not properly defined"
 #endif
 
 /* transcoding */


oliv3r:
Relates to #1508

oliv3r:
Hmm, briefly looking at it, this seems to relate to my other commits, I thought I had the order right. There's a few 'bugs' still left in tvh which get triggered by this.

This is on armv7 that it's failing, but why didn't the pipeline catch this before the merge?

Anyway, I'll dig further into this, it looks like either these compilers are not setting up the proper time define/variable, or it's not triggered properly from the build-system. But this is fixable :) even though it's highly annoying :p Interestingly enough, it also fails on aarch64 for stretch, but works again on bullseye. So its certainly also version related. One way to deal with this is through the configure step, and detect the requested depth. Anyway, with those pipeline results, I can emulate them and trigger the bug and fix it.

Flole998:
The pipeline is only checking for a single system (improvements welcome ;) )

It is failing on ALL 32 bit systems apparently.

oliv3r:
Ah, right, yeah we can improve that. I think with the 'low churn' it's okay to have slow pipelines, that at least check everything. I'll first fix the issue, then the pipeline.

Flole998:
Yeah and I can still decide to merge without waiting for all builds to run through, so there's no downside to this.

Saentist:
https://tvheadend.org/issues/6288#change-33811

oliv3r:
@seantist thanks for that cross-reference. Though not much of value is there sadly, other then what @Flole998 kind of already concluded.

I admit, the oversight is on my end, I didn't test enough systems properly. But as I wrote in the original topic, this actually is already breaking on modern i386 builds (on alpine in this case) because they feature modern libraries/macro's that try to ensure 64bit time_t. So now we are in a situation, where sometimes, on some systems, time_t is indeed 32bits, and sometimes, on some systems, it is 64bits. The macro's to configure/do stuff with things, are not supported everywhere, so cannot be reliably used.

So next to using __WORDSIZE, we need to also (or rather) check for what size time_t is on the target system (which we can hopefully do with our handwoven configure system :)

Saentist:
Most logical is when command "configure" run to check what environment have current system and apply correct scenario to build.
Code section for both 32 or 64 version can be activated later in compilation, without hardcoding to exact one solution.
This is logic of **IF THEN ELSE** statement.

Flole998:
You're missing the entire point here....

mralapete:
> Most logical is when command "configure" run to check what environment have current system and apply correct scenario to build.
> Code section for both 32 or 64 version can be activated later in compilation, without hardcoding to exact one solution.
> This is logic of **IF THEN ELSE** statement.

This is not the platform for debating the pros and cons of any changes to the project. TVH has a fine forum where you can air your opinions publicly for others to discuss too.

PR #1544, opened 2023-07-09
Feature #6283 - epggrab/eit - OTA Genre translations

DeltaMikeCharlie:
Allow the translation of the DVB OTA 'content_descriptor' (genre) received from the broadcaster from one value to another.  Used for converting non-standard codes (as used in Australia, for example) into standard ETSI codes.

[https://tvheadend.org/issues/6283](https://tvheadend.org/issues/6283)


src/epggrab.c
```diff
@@ -328,8 +328,16 @@ epggrab_class_ota_cron_notify(void *self, const char *lang)
   epggrab_ota_set_cron();
 }
 
+static void
+epggrab_class_ota_genre_translation_notify(void *self, const char *lang)
+{
+  //tvhtrace(LS_SETTINGS, "Genre translation settings changed.");
```
Flole998, line 334: Remove any commented code if it's not necessary/wanted

src/epggrab/module/eit.c
```diff
@@ -391,12 +391,21 @@ static int _eit_desc_component
 static int _eit_desc_content
   ( epggrab_module_t *mod, const uint8_t *ptr, int len, eit_event_t *ev )
 {
+  uint8_t tempPtr = 0;  //Temporary variable to hold a (potentially) changed *prt value.
```
Flole998, line 394: "... changed *prt value": You mean *ptr?

src/epggrab.c
DeltaMikeCharlie: Done

src/epggrab/module/eit.c
DeltaMikeCharlie: Done

DeltaMikeCharlie:
@Flole998 - Thanks for your feedback.  Both issues have been addressed.

DeltaMikeCharlie:
I need some guidance in tracking down why my PR fails to build on the GitHub build system yet seemed to work on mine during development.

**Error 1** - Multiple definitions of a new global variable.
**Error 2** - Undefined variable. [RESOLVED]

To check my build system, I cloned my tvh repository into an empty temporary directory and built from scratch.  Error 1 did not occur, however, error 2 did.

**Error 2**  [RESOLVED]

This error was down to my inexperience with the tvh build environment.  I have pushed this missing file to GitHub.

Unfortunately, this 'push' resulted in the error notification being removed from the PR.  However, I was still able to read it here: [https://github.com/tvheadend/tvheadend/actions/runs/5501565457/jobs/10060993444?pr=1544](https://github.com/tvheadend/tvheadend/actions/runs/5501565457/jobs/10060993444?pr=1544)

I discovered that instead of adding my documentation directly to 'docs_inc.c', I should have created a new file '{build_directory}/docs/property{topic_name}.md'.  'docs_inc.c' is then built dynamically when 'make' is executed.

**Error 1** is still puzzling as I don't get this error on a clean build on my system.

The GitHub build linker encounters an error because my new global variable 'epggrab_ota_genre_translation' has multiple (16) definitions.  I presume that this is because the header file containing the definition is included in multiple modules.

I checked my variable declaration and it is placed within the preprocessor '#ifndef __EPGGRAB_H__' block for that module, so it should only be evaluated once.  As far as I can see, '__EPGGRAB_H__' is only set once, and that is in 'epggrab.h' as expected.

The only major difference that I can see is that the GitHub build environment uses Ubuntu 22.04 whereas my build environment uses Ubuntu 20.04.

Suggestions on how to resolve 'error 1' would be appreciated.

DeltaMikeCharlie:
The GitHub build process has failed once more.  'Error 1' (above) still exists, however, this time there appears to be a problem building ffmpeg, a system that I did not touch.

How reliable is the build system?

@Flole998 - Are you able to provide any advice?

Flole998:
It didn't fail during ffmpeg, the issue is still the same, during linking it has a symbol name collision.

DeltaMikeCharlie:
@Flole998 - Thanks.

I finally tracked the error down to different default compiler/linker options between the GitHub build system and my build system.  Once I forced the "-fno-common" flag on my build system, I could recreate the error.  I have made the required changes and TVH now compiles without error on my build system.

src/epggrab/otamux.c
```diff
@@ -1035,6 +1037,94 @@ epggrab_ota_set_cron ( void )
...
+      }
+    }
+
+    //If we have a new line or a comma or we are at the end of the config string
+    if(epggrab_conf.ota_genre_translation[tempPos] == 10 || epggrab_conf.ota_genre_translation[tempPos] == ',' || tempPos == (tempLen - 1))
+    {
+      //Only process non-null strings.
+      if(strlen(tempPair) != 0)
+      {
+        //tvhinfo(LS_EPGGRAB, "Got pair '%s'", tempPair);
```
Flole998, line 1095: Don't commit code that's commented out. If you believe it's important for trace/debugging leave it in and adjust the log level accordingly, otherwise throw it out.

src/input/mpegts/mpegts_mux_sched.c
```diff
@@ -121,6 +121,7 @@ mpegts_mux_sched_class_cron_set ( void *p, const void *v )
 
 CLASS_DOC(mpegts_mux_sched)
 PROP_DOC(cron)
+//PROP_DOC(genre_translation)
```
Flole998, line 124: Commented out code

Flole998:
Please squash your commits. Also right now this PR is in a state where it can't be rebased and merged automatically, so you'd probably have to rebase it onto master manually.

DeltaMikeCharlie:
@Flole998 - Done.

src/epggrab/otamux.c
```diff
@@ -1035,6 +1037,92 @@ epggrab_ota_set_cron ( void )
...
+  //There is some sanity checking done, however, the data is entered as freeform
+  //text by the user and errors may still slip through.
+  //Note: Each line that comes back from the WebUI is separated by a LF (0x0A).
+  //^^^^  This has been tested with a Windows WebUI client and it only sends LF, not CR/LF.
+
+  lock_assert(&global_lock);
+  tvh_mutex_lock(&epggrab_ota_mutex);
+
+  //Reset the translation table to 1:1 here before proceeding.
+  for(int i = 0; i < 256; i++)
```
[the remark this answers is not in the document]
DeltaMikeCharlie: This was actually in 2 places.  Both fixed.

PR #1546, opened 2023-07-29
ffmpeg-static: fix nasm build dependancy

dreamcat4:
Hi,
Not sure who does want to review this PR, but it's for fixing failed / breaking part of the ffmpeg build dependancy.

When its being selected the following `./configure` flag: `--enable-ffmpeg_static`

For some reason the SSL certificate authentication is now failing for NASM tarball / source code download from the ancient upstream `nasm.us` url... So it would seem like their SSL cert for that domain has now expired or fallen to neglect.

Anyhow since it was not possible (from client side) to properly (and safely) get the tarball from that historical download location. This now PR switches to a more modern / recent upstream NASM download url. Which is their (newer) official github repository. It's same official project, so not a 3rd party or other mirror.

They seem to be keeping it up to date. All was needed just some minor tweaking of the download URL (inside of tvheadend's ffmpeg's dep. Makefile).

With this changeset, then NASM now downloads, compiles, and installs*. And for modern latest ubuntu release 23.04 Lunar Lobster. These changes are fairly minimal.

To double check that this is the official NASM upstream project. it was checked the github organisation people. Are the following 2 long term (and still active) team members:

```sh
Cyrill Gorcunov | cyrillos | Admin
H. Peter Anvin | hpa | Admin
```

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -22,6 +22,8 @@
         "debhelper",
         "ccache",
         "libx11-xcb-dev"
+        "asciidoc",
+        "xmlto"
       ],
       "buildcmd": [
         "AUTOBUILD_CONFIGURE_EXTRA=--enable-ccache\\ --enable-ffmpeg_static\\ --enable-hdhomerun_static\\ --nowerror ./Autobuild.sh -t ${TARGET} -j ${PARALLEL} -w ${WORKDIR}",
@@ -48,7 +50,9 @@
         "python",
         "python-requests",
         "debhelper",
-        "ccache"
+        "ccache",
+        "asciidoc",
+        "xmlto"
       ],
       "buildcmd": [
         "AUTOBUILD_CONFIGURE_EXTRA=--enable-ccache\\ --enable-ffmpeg_static\\ --enable-hdhomerun_static\\ --nowerror ./Autobuild.sh -t ${TARGET} -j ${PARALLEL} -w ${WORKDIR}",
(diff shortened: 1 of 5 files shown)


dreamcat4:
Build is failing here:

https://github.com/tvheadend/tvheadend/actions/runs/5702254457/job/15453774429?pr=1546#step:4:373

because NASM is missing these build dependancies: `asciidoc` and `xmlto`. They are required for generating the manpages for `make install` build target...

dreamcat4:
ci build fixed (github workflows updated 5/5).

Flole998:
Something is weird here: The version is the same, the SHA-Hash is different and the new one doesn't even build anymore without changes...

CvH:
depending at the package but github releases are frequently not a "real" release, so you have different files at official release and a automated github release

in general it is just bad practice from the package creators but it is what it is :)

same here, the packages are not identical (not saying that it won't work)
Official website points the downloads to https://www.nasm.us/pub/nasm/releasebuilds/, so better follow whatever the official website thinks is good :)

Flole998:
I have triggered a rebuild of latest master to see if it goes through and it does: https://github.com/tvheadend/tvheadend/actions/runs/5603487858/job/15461948747

So there is no issue, so no need to fix it.

dreamcat4:
It was some sort of upstream problem(s) - that they have since fixed their nsam.us SSL cert

> depending at the package but github releases are frequently not a "real" release, so you have different files at official release and a automated github release

Yes indeed... because in their upstream tarball on the `nasm.us` they first regenerate the `nasm.1` manpage / documentation files. Which is missing from the upstream git repo (from which the release is tagged). There are certain deficiencies within their `Makefile` around dependancies assumed for the `make install` build target. That it cannot build without those pre-generated documentation files. (which tvheadend project does not actually require for building locally the static ffmpeg, it just needs the patched nasm executable and little else vendored).

So looking retrospectively. A different way (and to not require pulling more dependancies in). would have been to ask the upstream (nasmguys ) to fix their Makefile. Or ourselves make a simpler patch to that Makefile.

But since the SSL cert is fixed... doesn't matter now. Looking at their Cert it's a 3 month duration short lifetime cert. Which is being generated by let's encrypt:

ksnip_20230804-112026

So they need to update keep their webserver  updated every 3 months. (not to break the nasm dependancy for ffmpeg static build again in future). Github (git tagged release) is just a bit more reliable in that sense. No big deal.

PR #1547, opened 2023-08-01
Add 'age rating' field to recording metadata - Feature #6287

DeltaMikeCharlie:
An 'age_rating' field has been added to the recording objects stored in '{TVH}/dvr/log'.  It is populated from the associated EPG event when a recording is created using the WebUI, HTSP API or JSON API.  It is also populated via HTSP/JSON API if the field is specifically included during the manual creation of the recording object.

This field can be accessed via:

1) the WebUI recording list pages (Upcoming, current, finished, failed, removed);
2) the '/api/dvr/entry/???' JSON API in the new 'ageRating' property;
3) the HTSP API 'dvrEntryAdd' and 'dvrEntryUpdate' methods in the new 'ageRating' property and;
4) saved in the MKV 'LAW_RATING' tag as a number (for now).

The new property name 'ageRating' for the DVR recording API object was selected to be consistent with the existing EPG API object property name.

In order to protect legacy HTSP clients, the property is only provided when the client requests a HTSP API version greater than 35.

Client Testing:

This modification was tested with a Kodi 'Matrix' client with no issues noted.

Other Testing:

Creation: WebUI, HTSP, JSON, TVH autoRec

Modification: HTSP, EPG update

Output: WebUI lists, HTSP (Ver > 35), JSON

No issues were noted during these tests.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -1207,6 +1207,8 @@ dvr_entry_create_from_htsmsg(htsmsg_t *conf, epg_broadcast_t *e)
     genre = LIST_FIRST(&e->genre);
     if (genre)
       htsmsg_add_u32(conf, "content_type", genre->code / 16);
+    if(e->age_rating)
+      htsmsg_add_u32(conf, "age_rating", e->age_rating);
   }
 
   de = dvr_entry_create(NULL, conf, 0);
@@ -1907,7 +1909,7 @@ dvr_is_better_recording_timeslot(const epg_broadcast_t *new_bcast, const dvr_ent
     if (svf == PROFILE_SVF_UHD && !old_has_svf) {
       old_has_svf = channel_has_correct_service_filter(old_channel, PROFILE_SVF_FHD);
       new_has_svf = channel_has_correct_service_filter(new_channel, PROFILE_SVF_FHD);
-      
+
       if (!old_has_svf && new_has_svf)
         return 1;
 
@@ -2366,6 +2368,7 @@ dvr_timer_remove_files(void *aux)
 #define DVR_UPDATED_CONFIG       (1<<17)
 #define DVR_UPDATED_PLAYPOS      (1<<18)
 #define DVR_UPDATED_PLAYCOUNT    (1<<19)
+#define DVR_UPDATED_AGE_RATING   (1<<20)
 
 static char *dvr_updated_str(char *buf, size_t buflen, int flags)
 {
@@ -2413,7 +2416,7 @@ static dvr_entry_t *_dvr_entry_update
     const char *lang, time_t start, time_t stop,
     time_t start_extra, time_t stop_extra,
     dvr_prio_t pri, int retention, int removal,
-    int playcount, int playposition)
+    int playcount, int playposition, int age_rating)
 {
   char buf[40];
   int save = 0, updated = 0;
@@ -2529,6 +2532,11 @@ static dvr_entry_t *_dvr_entry_update
     updated = 1;
     dvr_entry_set_timer(de);
   }
+  /* Manual Age Rating */
+  if (age_rating != de->de_age_rating) {
+    de->de_age_rating = age_rating;
+    save |= DVR_UPDATED_AGE_RATING;
+  }
 
   /* Title */
   if (e && e->title) {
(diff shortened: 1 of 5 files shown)


Flole998 (changes requested):

src/dvr/dvr_db.c
```diff
@@ -2529,6 +2532,12 @@ static dvr_entry_t *_dvr_entry_update
     updated = 1;
     dvr_entry_set_timer(de);
   }
+  /* Manual Age Rating */
+  //if (age_rating && (age_rating != de->de_age_rating)) {
```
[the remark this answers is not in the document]
DeltaMikeCharlie: @Flole998 - Sorry, old habits die hard.  This has been resolved.

github-user1979:
When I try to add timers for recordings using dream Player for Tvheadend (https://play.google.com/store/apps/details?id=de.cyberdream.tvh.player&hl=en_US), Tvheadend crashes. Maybe because the age rating field isn't provided?

DeltaMikeCharlie:
> When I try to add timers for recordings using dream Player for Tvheadend (https://play.google.com/store/apps/details?id=de.cyberdream.tvh.player&hl=en_US), Tvheadend crashes. Maybe because the age rating field isn't provided?

'Ghost' appears to be a deleted user.  If the OP reads this and is continuing to experience this issue, please open a bug report on the TVH forum: https://tvheadend.org/projects/tvheadend/issues

Sorry, I am not familiar with 'dream Player'.

Does the TVH crash happen every time that you try to create a recording via dream Player, or is it intermittent?

Can you create recording using the normal TVH web interface?

Are you able to create any logs on dream Player or TVH?

https://dreamepg.de/index.php/en/faq/faq-dreamplayer-tvh-en/technical-issues

It would be useful to know if dream Player uses HTSP or HTTP/JSON when creating recordings.  The TVH log would help with this.

mralapete:
It might be worth establishing what version of TVH is being used here. As far as I know Dream Player is used solely on Enigma 2 Linux set top boxes https://dreamepg.de/index.php/en and as far as I know would not be covered under the scope of this project. Possibly these questions should be directed to here 
https://www.linuxsat-support.com/cms/article-amp/106-installation-tvheadend-server-on-an-enigma2-box/
Sounds like some form of fork of TVH.

src/htsp_server.c
```diff
@@ -999,6 +999,15 @@ htsp_build_dvrentry(htsp_connection_t *htsp, dvr_entry_t *de, const char *method
     htsmsg_add_u32(out, "priority",    u32);
     htsmsg_add_u32(out, "contentType", de->de_content_type);
 
+    //To not risk breaking older clients, only
+    //provide the 'age rating' via HTSP if the requested
+    //API version if greater than 35.
+    if (htsp->htsp_version > 35)
```
ksooo, line 1005: Not sure this is the correct way to implement versioning. I think htsp version needs to be increased, so that a client after connecting to tvh knows the htsp version of the server is high enough so the client can expect and use the new property. 

Iirc, this is how we did former times.

DeltaMikeCharlie: > Not sure this is the correct way to implement versioning. I think htsp version needs to be increased, so that a client after connecting to tvh knows the htsp version of the server is high enough so the client can expect and use the new property.
> 
> Iirc, this is how we did former times.

Thanks.  I will look into this.

DeltaMikeCharlie: @ksooo - I looked into this:

When initiating contact via HTSP, the client sends a ‘hello’ message which includes the maximum HTSP version that it can understand.

```
{'clientname': 'My client name',
 'htspversion': 33,
 'seq': 1,
 'method': 'hello'}
```

When TVH receives the ‘hello’ message, it replies with an acknowledgement message that contains some details about the system including, but not limited to, the maximum HTSP version that it can accept.

```
[{'htspversion': 36,
'servername': 'Tvheadend',
'serverversion': '0.0.0-unknown',
'challenge': b'<SNIP>,
'language': 'eng',
'servercapability': ['caclient',
                     'tvadapters',
                     'satip_client',
                     'satip_server',
                     'trace',
                     'libav'],
'api_version': 19,
'seq': 1}]
```

The server then takes the lowest version number of either the client-requested version or the maximum supported version and sets the response protocol version to that value.  As far as I can tell, the server never actually confirms back to the client the active protocol version it is using.

https://github.com/tvheadend/tvheadend/blob/62adbebfd062d7b97829268274aad92df2033784/src/htsp_server.c#L1387-L1388

There are numerous areas within the module that test the active protocol version when formulating a response.

```
if (htsp->htsp_version < nn)
  set value A;
else
  set value B;
```

I can see that in Sep-2021, the version was incremented from 35 to 36 when the RDS field was added.  However, as far as I can tell, this version number is never tested for.  I suspect that regardless of the version that the client requests (Kodi pvr.hts is currently asking for v35), the ‘rds_uecp’ field will always be present.

I will soon be submitting a PR for my new 'rating label' feature.  I will increment the version number for that, but also include this field under that version.

PR #1548, opened 2023-08-04
Fix time for old 32bit systems

oliv3r:
The fix introduced in fe47ecb5504a ("Fix time for 32bit systems again") made the incorrect assumption, that the macro _TIME_BITS=64 was always available. It seems like that this is not the case for old systems, causing 32bit builds to break there.

Lets completely rework 32bit time support, by actually detecting the hosts size for the time.

This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.

Fixes fe47ecb5504a ("Fix time for 32bit systems again")

### DIFF
--- a/configure
+++ b/configure
@@ -279,6 +279,42 @@ int test(void)
 }
 '
 
+check_cc_snippet time_ld '
+#define _FILE_OFFSET_BITS 64
+#define _TIME_BITS 64
+#include <assert.h>
+#include <stdio.h>
+#include <time.h>
+#define TEST test
+int test(void)
+{
+  printf("%ld", (time_t)1);
+  return 0;
+}
+' -Werror
+
+check_cc_snippet time_lld '
+#define _FILE_OFFSET_BITS 64
+#define _TIME_BITS 64
+#include <assert.h>
+#include <stdio.h>
+#include <time.h>
+#define TEST test
+int test(void)
+{
+  printf("%lld", (time_t)1);
+  return 0;
+}
+' -Werror
+
+if enabled time_lld; then
+  printf "    ^ using time_t format 'lld'\n"
+elif enabled time_ld; then
+  printf "    ^ using time_t format 'ld'\n"
+else
+  die 'Unable to determine size of "time_t"'
+fi
+
 check_cc_snippet stime '
 #include <time.h>
 #define TEST test
--- a/src/tvheadend.h
+++ b/src/tvheadend.h
@@ -334,12 +334,12 @@ void tvh_qsort_r(void *base, size_t nmemb, size_t size, int (*compar)(const void
 # endif /* ULONG_MAX */
 #endif /* __WORDSIZE */
 
-#if __WORDSIZE == 32
+#if CONFIG_TIME_LLD == 1
 # define PRItime_t "lld"
-#elif __WORDSIZE == 64
+#elif CONFIG_TIME_LD == 1
 # define PRItime_t "ld"
 #else
-# error "__WORDSIZE not properly defined"
+# error "CONFIG_TIME not properly defined"
 #endif
 
 /* transcoding */


oliv3r:
> > @Flole998 this should fix this once-and-for-all, hopefully :p by checking this at compiletime.
> > I tried first to actually determine the size of time_t with sizeof and an assert, and that worked well, but then still, it wasn't happy. So now, we just do a check from the horse's mouth.
> > I'm testing a few archs, but don't have a lot of time _right_now_. It seems to run fine for atleast alpine x86-64 and arm32v7, but again, those are modern compilers.
> > If the people who left nice comments could also test it, that would be tons of helpful of course.
> 
> I’ve just one small question about your choice of OS Alpine Linux. I do realise it’s heavily centred around embedded systems. Do you have a particular reason why you chose Alpine Linux. I’m a long time user of Ubuntu so was wondering why you opted for this.

Quite a bit off-topic; but for one, it builds about 5 times as fast :), so doing test-cycles, this matters a lot. Especially when doing arm-emulated compilations, its much much faster. I think debian takes about 10 minutes to do a full compile, though note, that I have quite an old system; a amd FX 8 core, but with 24G ram :)

Secondly, alpine is amazing for containers, because not only is it faster (both run and compiletime) it also uses considerably less resources (diskspace and some RAM).

Now, for testing, I am actually testing these changes right now with ubuntu:16.04 in all 3 flavors, arm32v7, i386 and amd64. Though i386 and amd64 might get a bit skewered results, due to emulation; though the CI uses the same form of emulation. So there's that.

Debian is being troublesome atm, as I can't run anything apt-get in the arm32v7 debian:jessie container. No idea as to why. building the container does work, which does run apt-get of course, so no idea what's up there.

oliv3r:
Lets say I tried my best :D

Jumper78:
I am using the following system:
- Raspberry Pi 3 --> so, it would have a 64 bit capable processor
- Debian 11.7 based Linux (OSMC image for RasPi 2 / 3 / 3+ / Zero W 2)
- armv7l kernel (5.15.83-3-osmc SMP PREEMPT Sat Jan 7 04:06:56 UTC 2023 armv7l GNU/Linux) --> so, it is a 32 bit kernel

I am still not able to compile. I do get the following error message(s):
```
CC              src/rtsp.o
src/rtsp.c: In function ‘rtsp_describe_decode’:
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~                           ~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                         |
      |                                                         time_t * {aka long int *}
```

Please let me know if I can help fixing this by providing additional information or by trying something out.

oliv3r:
@Jumper78 Can you give me the output of `grep TIME_L build.linux/build.h`? This should have generated the correct printf type config.

I do have a fix for the other compile error, but afaik it doesn't happen yet on master? https://github.com/oliv3r/tvheadend/commit/def109ccde80208f3718bf146ced9708a870fc04 I wanted to fix this first, and then introduce prtime_t there.

oliv3r:
To give some background, the problem is a many combination of things. As said before, time_t can be either 32bits or 64bits. But that depends on a few things. glibc, gcc and the kernel, and obviously the target arch.

So I'm a little supprised the rasp-pi build is failing, because what this patch basically does, do a printf compile with time_t, and see which one works/sticks. So that's why it's surprising its not working.

th0ma7:
> This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.

For your awareness, I'm maintaining tvheadend on Synology through the spksrc Synocommunity framework.  And I am cross-compiling from x86_64 to armv5, armv7, armv7l, armv8 and i686 and ppc...

oliv3r:
> > This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.
> 
> For your awareness, I'm maintaining tvheadend on Synology through the spksrc Synocommunity framework. And I am cross-compiling from x86_64 to armv5, armv7, armv7l, armv8 and i686 and ppc...

Hey @th0ma7

How are you cross-compiling? From what I can see, the build-system doesn't really support it? Unless you setup a fully isolated (ch)root, which of course it will work as before. The reason I (probably slightly incorrectly) mentioned cross-compiling, as normally you tell your configure/meson build system to cross-compile for you. In this case, you can't, you have to convince the entire repo, including configure what your target arch is. This because quite a lot of test check the host system/host-kernel for 'stuff'.

So unless you manually set a lot of flags, the configure script will miss-detect a few things. For the time stuff, I've actually switched from doing a static_assert on sizeof(time_t), to a printf("%ld", (time_t)1) _compilation-only_ test. So it does no longer rely on runtime behavior. I'll should have updated the commit/MR message to indicate that, sorry about that.

What is your concern? And are you also already triggering this issue? If so, it would be nice to help debugging more failure case.

Thanks

Flole998:
> > This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.
> 
> For your awareness, I'm maintaining tvheadend on Synology through the spksrc Synocommunity framework. And I am cross-compiling from x86_64 to armv5, armv7, armv7l, armv8 and i686 and ppc...

I suggest you provide a file for GitHub actions that does cross compiling for various targets on pull requests to make sure it works properly. Otherwise it might break at any time (just look at latest master, it's completely broken on many architectures right now, part of it is due to this PR, part of it due to another).

oliv3r:
@Flole998 I might actually skip ahead and enable building everything for MR's first; as this is a little bit harder then I thought ...

```
  checking for cc time_ld ...                       ok
  checking for cc time_lld ...                      ok
    ^ using time_t format 'lld'
```
is for example what I get from here
https://github.com/tvheadend/tvheadend/actions/runs/5763878349/job/15626631965#step:3:937

which right now, totally baffles me. How can it compile both snippets just fine, but then obviously fail later. The only reason I can imagine, is if `-Werror` gets swallowed and ignored, which is fundamental for the compile check.

Could be that we just need to swap the order of the check, because some systems will happily comply, but then, why would it be failing during the actual compilation ...

Enabling the pipelines, while incredibly slow, at least allows me to test this using the CI runners, and I can enable verbose `check_cc` to see what's potentially going on.

Sorry for the constant breakage :( but we'll get this.

Flole998:
Yeah that's a very good idea, I already mentioned in the forum that I won't merge any additional code unless those tests are finally there.

Ideally you can just modify the repo build and disable the cloudsmith upload (or we just make the upload script exit if the key is not defined, and I restrict the key to master)

oliv3r:
> Yeah that's a very good idea, I already mentioned in the forum that I won't merge any additional code unless those tests are finally there.

Thanks,

> Ideally you can just modify the repo build and disable the cloudsmith upload (or we just make the upload script exit if the key is not defined, and I restrict the key to master)

I think the 'or' bit is best, I don't know how this works for github; but it gitlab, you have 'protected' branches (e.g. usually master, or release branches; and tags), where only these variables are accessable (for good reason). So if the branch is not protected, you do not get the variable, thus you cannot upload. It's far the cleanest solution and much safer, if this is possible at all.

th0ma7:
> > > This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.
> > 
> > 
> > For your awareness, I'm maintaining tvheadend on Synology through the spksrc Synocommunity framework. And I am cross-compiling from x86_64 to armv5, armv7, armv7l, armv8 and i686 and ppc...
> 
> I suggest you provide a file for GitHub actions that does cross compiling for various targets on pull requests to make sure it works properly. Otherwise it might break at any time (just look at latest master, it's completely broken on many architectures right now, part of it is due to this PR, part of it due to another).

We already have a github build process for package updates (see example https://github.com/SynoCommunity/spksrc/actions/runs/5761670907/job/15620165746?pr=5820).  So when ever I create a PR to update to a newer commit it gets triggerred automatically.

This uses spksrc framework which downloads Synology toolchains to cross-compile using a Debian docker environment.  While all the hooks are there I'm unsure how this could be ported in here? (which would be great btw).

th0ma7:
> > > This will break cross-compilation, as the configure script is run on the host, but since we do not seem to support cross-builds anyway (all archs are build on emulated or native targets), this should be acceptable.
> > 
> > 
> > For your awareness, I'm maintaining tvheadend on Synology through the spksrc Synocommunity framework. And I am cross-compiling from x86_64 to armv5, armv7, armv7l, armv8 and i686 and ppc...
> 
> Hey @th0ma7
> 
> How are you cross-compiling? From what I can see, the build-system doesn't really support it? Unless you setup a fully isolated (ch)root, which of course it will work as before. The reason I (probably slightly incorrectly) mentioned cross-compiling, as normally you tell your configure/meson build system to cross-compile for you. In this case, you can't, you have to convince the entire repo, including configure what your target arch is. This because quite a lot of test check the host system/host-kernel for 'stuff'. Or, all these tests use the local x86_64 compiler to do their tests, which technically leads to incorrect results.
> 
> So unless you manually set a lot of flags, the configure script will miss-detect a few things. For the time stuff, I've actually switched from doing a static_assert on sizeof(time_t), to a printf("%ld", (time_t)1) _compilation-only_ test. So it does no longer rely on runtime behavior. It does however rely on the target sys-root, as otherwise it will use your local environment and iss-detect, as I wrote above. I'll should have updated the commit/MR message to indicate that, sorry about that.
> 
> What is your concern? And are you also already triggering this issue? If so, it would be nice to help debugging more failure case.
> 
> Thanks

Long story short, spksrc (https://github.com/SynoCommunity/spksrc) is cross-compiling all the necessary bits using a Debian image with Synology toolchains.  All archs gets built PR trigerred automatically using github-action.

As for the flags, this gets all automagically set for either using automake, cmake or meson or whatever depending of what the source code tar-ball uses.

Currently using:
```
# git hash from April 8th 2023
PKG_GIT_HASH = f32c7c59a19a276648d7b068041738e4e8337638
```
But if that can help, I could create a temporary PR against a commit to see how things behave on my end and if it does still honor the arm5-7-ppc-i686 32-bit flags?

Flole998:
> > Ideally you can just modify the repo build and disable the cloudsmith upload (or we just make the upload script exit if the key is not defined, and I restrict the key to master)
> 
> I think the 'or' bit is best, I don't know how this works for github; but it gitlab, you have 'protected' branches (e.g. usually master, or release branches; and tags), where only these variables are accessable (for good reason). So if the branch is not protected, you do not get the variable, thus you cannot upload. It's far the cleanest solution and much safer, if this is possible at all.

I have created the "Cloudsmith" environment now and only allowed access in master. I am not sure what happens if that env is attempted to be used "outside" of master. So feel free to create a PR that uses it so we can see what happens (see the coverity file for an example).

th0ma7:
So if that can help, I opened-up a PR using spksrc against latest commit tvheadend https://github.com/SynoCommunity/spksrc/pull/5834

oliv3r:
> But if that can help, I could create a temporary PR against a commit to see how things behave on my end and if it does still honor the arm5-7-ppc-i686 32-bit flags?

Anything that produces some output would be useful yes, the louder the better. Right now, my hunch is that `-Werror` isn't being honored, so I'm looking if gcc4 supported `#pragma GCC diagnostic error "-Wformat"` (and/or if I can just supply `-Wformat=error`)?

> This uses spksrc framework which downloads Synology toolchains to cross-compile using a Debian docker environment. While all the hooks are there I'm unsure how this could be ported in here? (which would be great btw).

Are you certain that you are cross-compiling with those containers, and not using binfmt  emulation (which is what I have come to rely uppon). It makes cross-target compilation MUCH easier (at the cost of compile-speed), as it ends up being just a native compilation. Setting up a cross-environment to compile tvh is quite painful, as you need a cross-sysroot with all the libs that tvheadend needs...

th0ma7:
> > But if that can help, I could create a temporary PR against a commit to see how things behave on my end and if it does still honor the arm5-7-ppc-i686 32-bit flags?
> 
> Anything that produces some output would be useful yes, the louder the better. Right now, my hunch is that `-Werror` isn't being honored, so I'm looking if gcc4 supported `#pragma GCC diagnostic error "-Wformat"` (and/or if I can just supply `-Wformat=error`)?

My current configuration on master is https://github.com/SynoCommunity/spksrc/blob/master/cross/tvheadend/Makefile
Just let me know if you want to test some specifics flags on gcc-4.  I currently cross-compile for Synology DSM6 using gcc-4 and DSM7 using gcc-8.

> > This uses spksrc framework which downloads Synology toolchains to cross-compile using a Debian docker environment. While all the hooks are there I'm unsure how this could be ported in here? (which would be great btw).
> 
> Are you certain that you are cross-compiling with those containers, and not using binfmt emulation (which is what I have come to rely uppon). It makes cross-target compilation MUCH easier (at the cost of compile-speed), as it ends up being just a native compilation. Setting up a cross-environment to compile tvh is quite painful, as you need a cross-sysroot with all the libs that tvheadend needs...

I am cross-compiling, no worries.  Feel free to install it yourself using either docker or lxc/lxd, instructions are available on the frontpage.  you then only have to:
```
$ cd spksrc
$ make setup
$ cd spk/ffmpeg5
$ make all-supported
$ cd ../spk/tvheadend
$ make all-supported
```
Depending of your CPU type you may whish to do something like `PARALLEL_MAKE=max make -j4 make all-supported` ... works best on my 12 core AMD.

Jumper78:
> @Jumper78 Can you give me the output of `grep TIME_L build.linux/build.h`? This should have generated the correct printf type config. Also the output of 'configure' should be useful, the lines regarding the time detections. You could also remove the `> /dev/null` redirect from `check_cc`'s compile line, so that we get output thereof.

I am compiling on my two different TvHeadend systems in parallel. That should give more consistent results. The systems are:
1. RasPi 4: Debian 11.7, kernel 6.1.21-v8+ - aarch64, compiler flag `arch=arm64`
2. RasPi 3: OSMC (Debian 11.7 based), kernel 5.15.83-3-osmc - armv7l, compiler flag `--arch=armhf`

### Results from config script:
1. RasPi 4
```
  checking for cc time_ld ...                       ok
  checking for cc time_lld ...                      ok
    ^ using time_t format 'lld'
  checking for cc stime ...                         fail
```
2. RasPi 3
```
  checking for cc time_ld ...                       ok
  checking for cc time_lld ...                      ok
    ^ using time_t format 'lld'
  checking for cc stime ...                         fail
```

### Results from `grep TIME_L build.linux/build.h`:
1. RasPi 4
```
#define ENABLE_TIME_LD 1
#define CONFIG_TIME_LD 1
#define ENABLE_TIME_LLD 1
#define CONFIG_TIME_LLD 1
```
3. RasPi 3
```
#define ENABLE_TIME_LD 1
#define CONFIG_TIME_LD 1
#define ENABLE_TIME_LLD 1
#define CONFIG_TIME_LLD 1
```

### Result from compiling (`make`)
1. RasPi 4
```
CC              src/rtsp.o
src/rtsp.c: In function ‘rtsp_describe_decode’:
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~                           ~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                         |
      |                                                         time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 4 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~
  293 |             &hc->hc_rtsp_range_end);
      |             ~~~~~~~~~~~~~~~~~~~~~~
      |             |
      |             time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c: In function ‘rtsp_set_position’:
src/rtsp.c:333:30: error: format ‘%lld’ expects argument of type ‘long long int’, but argument 4 has type ‘time_t’ {aka ‘long int’} [-Werror=format=]
  333 |   snprintf(buf, sizeof(buf), "npt=%" PRItime_t "-", position);
      |                              ^~~~~~~                ~~~~~~~~
      |                                                     |
      |                                                     time_t {aka long int}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
cc1: all warnings being treated as errors
make: *** [Makefile:720: /home/jens/tvheadend-temp/tvheadend/build.linux/src/rtsp.o] Error 1
```
3. RasPi 3
```
CC              src/rtsp.o
src/rtsp.c: In function ‘rtsp_describe_decode’:
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~                           ~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                         |
      |                                                         time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 4 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~
  293 |             &hc->hc_rtsp_range_end);
      |             ~~~~~~~~~~~~~~~~~~~~~~
      |             |
      |             time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c: In function ‘rtsp_set_position’:
src/rtsp.c:333:30: error: format ‘%lld’ expects argument of type ‘long long int’, but argument 4 has type ‘time_t’ {aka ‘long int’} [-Werror=format=]
  333 |   snprintf(buf, sizeof(buf), "npt=%" PRItime_t "-", position);
      |                              ^~~~~~~                ~~~~~~~~
      |                                                     |
      |                                                     time_t {aka long int}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
cc1: all warnings being treated as errors
make: *** [Makefile:720: /home/osmc/tvheadend-temp/tvheadend/build.linux/src/rtsp.o] Error 1
```

Jumper78:
Going back to the status after #1547, I do get the following results on my two systems:
1. RasPi 4
Code compiles through
2. RasPi 3
```
CC              src/rtsp.o
src/rtsp.c: In function ‘rtsp_describe_decode’:
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 3 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~                           ~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                         |
      |                                                         time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c:292:23: error: format ‘%lld’ expects argument of type ‘long long int *’, but argument 4 has type ‘time_t *’ {aka ‘long int *’} [-Werror=format=]
  292 |         sscanf(p + 8, "npt=%" PRItime_t "-%" PRItime_t, &hc->hc_rtsp_range_start,
      |                       ^~~~~~~
  293 |             &hc->hc_rtsp_range_end);
      |             ~~~~~~~~~~~~~~~~~~~~~~
      |             |
      |             time_t * {aka long int *}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
src/rtsp.c: In function ‘rtsp_set_position’:
src/rtsp.c:333:30: error: format ‘%lld’ expects argument of type ‘long long int’, but argument 4 has type ‘time_t’ {aka ‘long int’} [-Werror=format=]
  333 |   snprintf(buf, sizeof(buf), "npt=%" PRItime_t "-", position);
      |                              ^~~~~~~                ~~~~~~~~
      |                                                     |
      |                                                     time_t {aka long int}
In file included from src/rtsp.c:21:
src/tvheadend.h:338:23: note: format string is defined here
  338 | # define PRItime_t "lld"
cc1: all warnings being treated as errors
make: *** [Makefile:720: /home/xxx/tvheadend-temp/tvheadend-d50105999522cc7c35909f7c0f2a504fc40c2e1b/build.linux/src/rtsp.o] Error 1
```

Jumper78:
Both of my systems (RasPi 4 aarch64 and RasPi 3 armv7l) seem to want to have `#define PRItime_t "ld"` in `src/tvheadend.h` in order to work. If I compile them with `lld`, I do get errors during compiling.

oliv3r:
@Jumper78 can you check my latest patch in #1551 I think this should fix it and prevent from BOTH LD options to be set.

Jumper78:
> @Jumper78 can you check my latest patch in #1551 I think this should fix it and prevent from BOTH LD options to be set.

Compiling is still ongoing, but the configuration run looks already promising:
1. RasPi 4
```
  checking for cc time_ld ...                       ok
  checking for cc time_lld ...                      fail
    ^ using time_t format 'ld'
```
2. RasPi 3
```
  checking for cc time_ld ...                       ok
  checking for cc time_lld ...                      fail
    ^ using time_t format 'ld'
```

Jumper78:
@oliv3r, the patch #1551 works for me! I can compile and run on both my machines again.
Thank you very much!

PR #1550, opened 2023-08-04
Debian 64-bit time compliance

johnrosshunt:
This patch will insure 64-bit time_t is used on 64-bit systems and GNU/Linux systems where glibc >=2.34.

Older 32-bit GNU/Linux systems (where glibc < 2.34) as well as 32-bit Darwin/BSD will continue to work as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ CFLAGS  += -Wmissing-prototypes
 CFLAGS  += -fno-strict-aliasing
 CFLAGS  += -fms-extensions -funsigned-char
 CFLAGS  += -D_FILE_OFFSET_BITS=64
+CFLAGS  += -D_TIME_BITS=64
 CFLAGS  += -I${BUILDDIR} -I${ROOTDIR}/src -I${ROOTDIR}
 ifeq ($(CONFIG_ANDROID),yes)
 LDFLAGS += -ldl -lm
--- a/configure
+++ b/configure
@@ -76,6 +76,7 @@ OPTIONS=(
   "cclang_threadsan:no"
   "gperftools:no"
   "execinfo:auto"
+  "year2038:no"
 )
 
 #
@@ -802,6 +803,25 @@ if enabled_or_auto libsystemd_daemon; then
   elif enabled libsystemd_daemon; then
     die "libsystemd-daemon development support not found (use --disable-systemd_daemon)"
   fi
+fi
+
+#
+# year2038
+#
+check_cc_snippet year2038 '
+#define _FILE_OFFSET_BITS 64
+#define _TIME_BITS 64
+#include <time.h>
+int test(void){
+  #define LARGE_TIME_T ((time_t) (((time_t) 1 << 30) - 1 + 3 * ((time_t) 1 << 30)))
+  int verify_time_t_range[(LARGE_TIME_T / 65537 == 65535 && LARGE_TIME_T % 65537 == 0) ? 1 : -1];
+  return 0;
+}
+' -Werror
+
+if disabled year2038; then
+  CFLAGS="-D_USE_32BIT_TIME_T ${CFLAGS}"
+  printf "    ^ WARNING: The 'time_t' type stops working after January 2038\n"
 fi
 
 # ###########################################################################
--- a/src/epggrab/otamux.c
+++ b/src/epggrab/otamux.c
@@ -723,7 +723,7 @@ epggrab_ota_start_cb ( void *p );
 static void
 epggrab_ota_next_arm( time_t next )
 {
-  tvhtrace(LS_EPGGRAB, "next ota start event in %li seconds", next - time(NULL));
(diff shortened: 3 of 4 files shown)


src/epggrab/otamux.c
```diff
@@ -723,7 +723,7 @@ epggrab_ota_start_cb ( void *p );
 static void
 epggrab_ota_next_arm( time_t next )
 {
-  tvhtrace(LS_EPGGRAB, "next ota start event in %li seconds", next - time(NULL));
```
oliv3r, line 726: This is an related, unrelated patch which already has an MR open #1549

src/tvheadend.h
```diff
@@ -334,10 +334,16 @@ void tvh_qsort_r(void *base, size_t nmemb, size_t size, int (*compar)(const void
 # endif /* ULONG_MAX */
 #endif /* __WORDSIZE */
 
-#if __WORDSIZE == 32 && defined(PLATFORM_FREEBSD)
-#define PRItime_t       "d"
```
oliv3r, line 338: I think you based your MR not on master. And the problem is a bit more difficult then this, right now, we still see breakage on at least debian on raspberry pi.

Feel free to so share thoughts and insights in #1548

configure
```diff
@@ -804,6 +805,25 @@ if enabled_or_auto libsystemd_daemon; then
   fi
 fi
 
+#
+# year2038
```
oliv3r, line 809: I think from an ordering pov, it would be nicer to have this before all other time functions, no?

configure
```diff
@@ -804,6 +805,25 @@ if enabled_or_auto libsystemd_daemon; then
...
+#include <time.h>
+int test(void){
+  #define LARGE_TIME_T ((time_t) (((time_t) 1 << 30) - 1 + 3 * ((time_t) 1 << 30)))
+  int verify_time_t_range[(LARGE_TIME_T / 65537 == 65535 && LARGE_TIME_T % 65537 == 0) ? 1 : -1];
+  return 0;
+}
+' -Werror
+
+if disabled year2038; then
+  CFLAGS="-D_USE_32BIT_TIME_T ${CFLAGS}"
```
oliv3r, line 823: I don't think you are supposed to set a cflag, as `CONFIG_YEAR238` and `ENABLE_YEAR238` now exist in build.h

configure
```diff
@@ -76,6 +76,7 @@ OPTIONS=(
   "cclang_threadsan:no"
   "gperftools:no"
   "execinfo:auto"
+  "year2038:no"
```
oliv3r, line 79: I'm not sure how exactly the build-system works, but afaik this allows you to set a default and manual override? Is this actually needed at all?

oliv3r:
I'm not sure your patch does what it says it does. The year2038 test seems to do some bounds checking and based on that, decide whether to use `ld` or `lld`. If you look at master, my current attempt is to actually just try to compile a test snippet (#1548), but even then it seems to cause issues (which I'm trying to understand, as there's no reason that shouldn't work ...).

So I'd recommend to first rebase, then figure make sure your solution works for all supported architectures properly.

Makefile
```diff
@@ -54,6 +54,7 @@ CFLAGS  += -Wmissing-prototypes
 CFLAGS  += -fno-strict-aliasing
 CFLAGS  += -fms-extensions -funsigned-char
 CFLAGS  += -D_FILE_OFFSET_BITS=64
+CFLAGS  += -D_TIME_BITS=64
```
oliv3r, line 57: this is already in master ...

johnrosshunt:
This patch was made against commit [c531383](https://github.com/tvheadend/tvheadend/commit/c531383ca6654639dc112db67fd8dc893c1f5272).  That's because it's mutually exclusive to your later patches -- which are broken.  So, if this patch doesn't meet your needs, please close this PR.  I'm done.  I'm out.

oliv3r:
> This patch was made against commit [c531383](https://github.com/tvheadend/tvheadend/commit/c531383ca6654639dc112db67fd8dc893c1f5272). That's because it's mutually exclusive to your later patches -- which are broken. So, if this patch doesn't meet your needs, please close this PR. I'm done. I'm out.

But that commit is 2 weeks old. Yes, my patches where not perfect, they where not broken generally, they are broken on old systems, which we didn't trigger on. We are now trying to get the pipeline to trigger on any MR, to avoid this in the future.

Btw, my patches have been fixed since, and should be working according #1551 so once the pipeline can prove that, we're good.

As for basing it on [c531383](https://github.com/tvheadend/tvheadend/commit/c531383ca6654639dc112db67fd8dc893c1f5272), that's not master, so you'll end up with merge conflicts preventing the merge. So a rebase before opening the MR would have been very helpful. But your patches are certainly not mutually exclusive! I suppose if you would have rebased and checked, you would have seen that.

oliv3r:
@johnrosshunt we've now merged proper support for 64 and 32bit time for old and new systems the like. So old 32bit environments that do support 64bit time, can now happily use it.

Hope this helps.

PR #1551, opened 2023-08-05
Use explicitly on format warnings for Time test

oliv3r:
It looks like the compile check doesn't work properly on some architectures, which appears to eat or ignore the -Werror flag.

Instead, be more specific in that we state that printf formatting errors are to be triggered on, and treated as an error explicitly, which in the end is exactly what we are after.

### DIFF
--- a/configure
+++ b/configure
@@ -291,7 +291,7 @@ int test(void)
   printf("%ld", (time_t)1);
   return 0;
 }
-' -Werror
+' '-Wformat -Werror=format'
 
 check_cc_snippet time_lld '
 #define _FILE_OFFSET_BITS 64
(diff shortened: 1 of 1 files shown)


oliv3r:
@Flole998 this should REALLY fix it this time :p but lets wait for the CI MR first this time ;)

You can of course start to review it :D

Jumper78 (commented):
These changes fix the compile issue on both of my machines (RasPi 3 with 32 bit and RasPi 4 with 64 bit).

oliv3r:
@Flole998 looks like we have a winner; lets get the pipeline check in, and see what's up

PR #1552, opened 2023-08-05
Always run CI

oliv3r:
The CI currently does not run our huge amount of targets as this is only enabled on master.

This MR refactors the pipeline to always run the pipeline, but only upload cloudsmith data from special/protected branches (e.g. where the secret is available).


oliv3r:
@Flole998 this is a bit of a WIP, as I can't test this locally. I might just burn my own repo a bit though, as I might be able to do a little more without affecting master?

Anyway, I already see it's not working :p so there's that ...

.github/workflows/build.yml
```diff
@@ -12,9 +18,6 @@ jobs:
     runs-on: ubuntu-latest
     continue-on-error: true
     name: Build on Raspberry Pi ${{ matrix.arch }}
-    env:
-      GITHUB_ENV: CLOUDSMITH_API_KEY
-      CLOUDSMITH_API_KEY: ${{ secrets.CLOUDSMITH_KEY }}
```
Flole998, line 17: At least this line needs to stay, it makes the API key available in the container

.github/workflows/build.yml
```diff
@@ -48,12 +51,16 @@ jobs:
           sudo apt-get install --force-yes -y cmake git build-essential pkg-config gettext libavahi-client-dev libssl-dev zlib1g-dev wget bzip2 git-core liburiparser-dev libdvbcsa-dev python3 python3-requests debhelper ccache lsb-release
           DEBIAN_FRONTEND=noninteractive apt-get install --force-yes -y libpcre3-dev || DEBIAN_FRONTEND=noninteractive apt-get install --force-yes -y libpcre2-dev
           AUTOBUILD_CONFIGURE_EXTRA=--enable-ccache\ --enable-ffmpeg_static\ --enable-hdhomerun_static\ --python=python3 ./Autobuild.sh -p raspios
-          support/cloudsmith.sh -p raspios -f '../tvheadend*.deb'
     - uses: actions/upload-artifact@v3
+      env:
```
Flole998, line 55: This won't work I guess

Flole998:
I suggest you revert all changes and only add `environment: Cloudsmith` to each job, let's see what happens. Oh and also edit the cloudsmith.sh file to show something like "no upload needed" and then it exits

oliv3r:
> I suggest you revert all changes and only add `environment: Cloudsmith` to each job, let's see what happens. Oh and also edit the cloudsmith.sh file to show something like "no upload needed" and then it exits

I've added a variable check to the script instead, it's ugly; but should at least do what we want ;)

.github/workflows/build.yml
oliv3r: the line was moved to the 'upload' job

.github/workflows/build.yml
oliv3r: I think the 'with text' failure makes the whole job fail, I thought just the step; idk.

Flole998:
I think you still need to add the "environment" entry, otherwise the normal builds won't have access to the key either

oliv3r:
@Flole998 done; not sure why the whole CI refuses to run now though ...

oliv3r:
Hmm, what's this now? All jobs 'pass', no logs. Also its done way to fast. I enjoy everything being green, but that can't be right?

Flole998:
I guess this is because only master is allowed to run in this environment.... This seems to be more complicated than expected

oliv3r:
@Flole998 I suppose my original thought makes more sense, only run the upload job when the environment/is set.

I suppose there's no way to make the variable only available on protected/certain branches, while not touching the environment?

oliv3r:
@Flole998 so there is a way to limit secret exposure from the yaml file. The odd thing is of course that this is in the yaml file, which is ... strange. But this should work. I still don't see the pipelines though; but could be again that I'm impatient :p

Flole998:
It's complaining about illegal syntax, so this is not it. I guess we could just add a condition in the script to check if we are on master, and abort if we aren't.

oliv3r:
Where do you see the syntax error though? All checks seem to be green :S 
but yeah, just adding the master check in the script is easier, but so ugly :p

Flole998:
That syntax error is only visible when using GitHubs Editor.

According to the documentation secrets aren't available to pull requests, so this might actually be fine.... Let's remove the tags though as we only want to build on master for now, otherwise cloudsmith would mark the tag as latest release and everyone would install that. It's not setup for that at all.

oliv3r:
> That syntax error is only visible when using GitHubs Editor.
> 
> According to the documentation secrets aren't available to pull requests, so this might actually be fine.... Let's remove the tags though as we only want to build on master for now, otherwise cloudsmith would mark the tag as latest release and everyone would install that. It's not setup for that at all.

Heh, I asked google bard if this was valid syntax and it was sure. I used the github editor, and it says 'nope'.

I reverted everything back as it was, with the pipeline running for everything reasonable, and added the checks to the script for now, we can experiment more later; when the pipeline script gets cleaned up/re-written potentially. Lets see how this now fares ...

oliv3r:
Sadly, this also doesn't work; `Error: run-on-arch: env CLOUDSMITH_API_KEY value must be flat.`

Whaddamess :(

oliv3r:
@Flole998 I"m giving up for now; unless you have some magical insight. It's almost as if the CLOUDSMITH_KEY no longer exists, causing it to return an (empty) object, which causes run-on-arch to fail with this nestedness crap.

Flole998:
It does exist, it's just inaccessible from PRs as mentioned in the docs. I'm not really sure how to solve that though. Maybe we should just copy that file for PRs and just remove the cloudsmith script. Then we can also choose to enable additonal options, like ffmpeg and whatnot so the tests cover more code.

oliv3r:
> It does exist, it's just inaccessible from PRs as mentioned in the docs. I'm not really sure how to solve that though. Maybe we should just copy that file for PRs and just remove the cloudsmith script. Then we can also choose to enable additonal options, like ffmpeg and whatnot so the tests cover more code.

Yeah, but other actions run fine, so it's that check that I linked above; and 'duh' about not being available on PR's.

So yeah, lets do that for now, lets copy the script. The whole CI needs to cleaned up a bit at some point; but that's for a nother time :p

oliv3r:
@Flole998 this is looking quite good so far!

I think in the future, we can make the builds faster though, by disabling some external builds maybe? I see the builds take a long time because it's building like ffmpeg n stuff. So we can rely on external libs. And of course as you mentioned, enabling more code to get higher code coverage.

It's not finished yet, but I feel confident now :)

Flole998:
It's no problem that the builds are taking a long time, usually there's no need to rush as I'm not going to review a PR within an hour or so anyways. It's not like we need to pay for the compute time or something like that ;)

oliv3r:
@Flole998 true, but when doing the actual work, and triggering a bug, it's nice if its somewhat fast :)

Flole998:
I will do additional changes later/tomorrow but now you can at least rebase the other PR to see if we get more green then. It still won't be 100% green due to other issues, but we're slowly getting there 😉

PR #1556, opened 2023-08-06
satip server: Rename nosatip to nosatipcli CLI flag

oliv3r:
The satip_boot function was added to config_boot in commit 442a871839bb ("satip server: fixed the RTP TCP size initialization, fixes #4517"), which doesn't tell us at all why it is there.

Whats worse, the nosatip flag seems to be ignored and the satip server still gets loaded.

Move things around a little so we can put them together under the nosatip flag to honor the intend better.

### DIFF
--- a/docs/markdown/cmdline_options.md
+++ b/docs/markdown/cmdline_options.md
@@ -27,7 +27,7 @@ Usage: `tvheadend [OPTIONS]`
       --satip_bindaddr        Specify bind address for SAT>IP server
       --satip_rtsp            SAT>IP RTSP port number for server
                               (default: -1 = disable, 0 = webconfig, standard port is 554)
-      --nosatip               Disable SAT>IP client
+      --nosatipcli            Disable SAT>IP client
       --satip_xml             URL with the SAT>IP server XML location
 ```
 ### Server connectivity
--- a/man/tvheadend.1
+++ b/man/tvheadend.1
@@ -66,7 +66,7 @@ Only use specified DVB adapters (comma separated, -1 == none).
 SAT>IP RTSP port number for server (default: -1 = disable, 0 = 
 webconfig, standard port is 554).
 .TP
-\fB\-\-nosatip\fR
+\fB\-\-nosatipcli\fR
 Disable SAT>IP client.
 .TP
 \fB\-\-satip_xml\fR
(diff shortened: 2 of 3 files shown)


oliv3r:
Note, that I have not been able to properly test this, as I lack infrastructure to do so. Disabling satip with -nosatip seems to work now though according to debug prints.

Flole998:
Let me fix the CI first, we still don't get all 45 checkmarks green.

Flole998:
That flag is for the sat>ip client, you are now disabling the server. That seems not what is intended.

oliv3r:
> That flag is for the sat>ip client, you are now disabling the server. That seems not what is intended.

You sure? nasty double negations :)

The flag is 'nosatip' which would indicate to me, that it DISABLES the sat->ip server. So we check if nosatip is NOT set, and if it is not set, we start the server. Seems ok?

Flole998:
That flag isn't related to the server at all. It is only for the client, and for that it works.

oliv3r:
Ohhh! server and CLIENT. Ok I'll rename the 'nosatip' flag to 'nosatipclient' for internal use, and see if it makes sense to add a flag for external.

The 'move' of the boot function should probably be in a better place though; not in the config section

oliv3r:
This is a brainfart really. I cleaned up the client stuff a little, and introduced the server flag.

If we feel the server flag is a bad idea, I can remove it, but keep the rest of this MR still.

Flole998:
Isn't configuring the port as 0 already disabling the server?

Flole998:
I think this is now also accessing an uninitialized variable....

oliv3r:
@Flole998 

I don't know, I am no satip user/expert. Can you elaborate on the uninitialized variable?

This is becoming more complicated then anticipated :D though the uninitialized variable sounds like a nastyness to begin with tbh

Flole998:
I am not sure about the uninitialized variable, however you are overwriting the config with the default values apparently: In `satip_server_boot` there is
```
satip_server_conf.satip_deviceid = 1;
satip_server_conf.satip_rtptcpsize = 7896/188;
```
which would overwrite the settings previously loaded.

oliv3r:
Hmm, maybe that's why the boot function is in the middle of the config block? If we only had a commit message that explained what they where doing and _why_ :p

Flole998:
> Hmm, maybe that's why the boot function is in the middle of the config block? If we only had a commit message that explained what they where doing and _why_ :p

To fix a bug, seems like the bug that you were about to reintroduce ;)

oliv3r:
lol, to work-around a bug :p ok fair nuff, not sure if I want to go through this effort right now :D

PR #1557, opened 2023-08-07
CI: drop support for jessie and stretch

oliv3r:
Jessie and Stretch have been End-of-Life for a long time. They do not get security updates etc. Also, it requires extra effort to get them to actually work again, due to the apt-repo's being archived.

Unless we have statistics that prove people are still downloading/using jessie and stretch, I think it's best to remove the maintenance burden on our side.

### DIFF
--- a/.github/workflows/build-cloudsmith.yml
+++ b/.github/workflows/build-cloudsmith.yml
@@ -64,18 +64,12 @@ jobs:
     strategy:
       matrix:
        arch: [ armv7, aarch64]
-       distro: [ stretch, buster, bullseye, ubuntu16.04, ubuntu18.04, ubuntu20.04, ubuntu22.04 ]
+       distro: [ buster, bullseye, ubuntu16.04, ubuntu18.04, ubuntu20.04, ubuntu22.04 ]
        include:
-         - arch: armv6
-           distro: jessie
-         - arch: armv6
-           distro: stretch
          - arch: armv6
            distro: buster
          - arch: armv6
            distro: bullseye
-         - arch: armv7
-           distro: jessie
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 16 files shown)


oliv3r:
See https://tvheadend.org/boards/5/topics/50272 to add to the conversation

oliv3r:
@Flole998 it's not done running yet, but its looking good so far. Sid is not possible right now, because its to new (trixie support for the most part).

Trusty is failing because of the 'for' loop thing.  But I'm actually inclined to add trusty to the list, it's the LTS of 2014! 10 years ago. Dropping trusty also allows us to actually use for loop initializers ;)

Flole998:
Sid/Trixie is failing due to ffmpeg having an issue, it will be fixed in the next ffmpeg release.

oliv3r:
@Flole998 dropping ancient debians makes all but trixie builds clean and green.

So you decide if you want to keep supporting these ancient and unsupported debians or drop them from the CI :)

Flole998:
If you want we can drop all the builds that are failing upload from the repo build. For the CI it is fixed now, only sid/trixie is now failing.

mralapete:
Possibly a naive question but would drawing a line for support for anything greater than 5 years old be a good yardstick to use.

Flole998:
We can still make sure it compiles (that's not a huge effort), just no longer provide packages for it. If someone really wants to use it they can self compile it.

oliv3r:
I'll rebase it tomorrow, I think it still makes sense to drop anything that is no longer maintained upstream.

If anybody still needs it for old architectures, they can compile it themselves of course.

I'd also argue, we should be open to allow modern gcc features `for (int i = 0...` comes to mind, because all systems (except the really old ones) support it already as well.

As for 'we can still compile it' yeah we could, but I don't think it's nice/fair to debian to burn their archive bandwidth.

And as said 'we should draw the line somewhere'. And again, it's not like it'll be impossible :) One solution could be though, (next to the docker solution, which also solves this problem) is to offer a statically build binary instead. That'll pretty much work always.

Flole998:
They can compile it if we don't break it. That's why I want to keep it active for CI and no longer build all the stuff for the repo that didn't upload anyways.

oliv3r:
But how far back do you want to keep support; and thus keep maintaining things, and also not evolving? Even Linus switch the kernel to modern set of features. (maybe even these old distro's support -cstd=GNU10 or whatever, but that's a secondary discussion :D)

https://tvheadend.org/boards/5/topics/50272 can be used to have the technical discussion on the 'cutoff' date, or here :)

Where do we draw the line? Even if users build it themselves, and we have a CI, it does taking into account said users.

Flole998:
We are still hitting issues on supported builds with recently introduced changes (see the latest bug report), so reducing tests is not a good idea. I don't see the advantage from dropping support for older versions at this point. I know that people run Tvheadend on old NAS devices with ancient build-environments, and people will complain it we cut them off.

Flole998:
All builds have been fixed.

PR #1559, opened 2023-08-11
Fix for bug #6297

DeltaMikeCharlie:
Change params[] index from 25 to 24 in dvr.js module.

### DIFF
--- a/src/webui/static/app/dvr.js
+++ b/src/webui/static/app/dvr.js
@@ -73,7 +73,7 @@ tvheadend.dvrDetails = function(grid, index) {
         var genre = params[21].value;
         /* channelname is unused param 22 */
         var fanart_image = params[23].value;
-        var age_rating = params[25].value;
+        var age_rating = params[24].value;
         var content = '<div class="dvr-details-dialog">' +
         '<div class="dvr-details-dialog-background-image"></div>' +
         '<div class="dvr-details-dialog-content">';


DonKult:
While this resolves the TypeError at least, it uses the wrong value: 24 is not the age rating – that value is never requested. 24 is a value dubbed "Broadcast". That value isn't displayed in the dialog nor in most grids, only the upcoming tab has the possibility to enable that column and I have no idea what that number represents but as the lowest for me seems to be 115 while the highest is 898 I somehow doubt it is related to age :wink: On the "upside", the dialog isn't showing the age rating at all currently, so what value isn't showing is perhaps not that important…

btw: You don't have to trust me: In your browser, check the networking tab and inspect the '/api/idnode/load` POST request that is backing the dialog. The returned JSON contains beside the actual values also details about the individual params which I think in an ideal world should be used to make this API at least a bit less evil. The epg dialog does it somehow at least. Oh well, the dvr c-code isn't all pretty either…

Anyhow, I think I got a better fix in the pipeline as I stumbled over the problem today, so lets see how that goes.

DeltaMikeCharlie:
@DonKult - Thanks for your feedback.

The initial problem that stopped the details windows from loading was that that the index was out of bounds.  By adjusting the index accordingly, I have resolved that specific issue, at least on my system.  A quick bug fix to restore functionality was my first priority, as the value is not actually displayed, it has not impact on the user.

I obviously need to look into this a lot deeper as it was clearly not in a state to be released.

Flole998:
I believe #1560 made this obsolete?

DeltaMikeCharlie:
> I believe #1560 made this obsolete?

Agreed.

PR #1561, opened 2023-08-13
Warn missing config

oliv3r:
When using the `--fork` flag, and no user or config arguments are
supplied, the configuration folder will end up with whatever the default
`daemon` user has set, which is often `/sbin` set as the homedir.
    
This is weird, but not 'wrong' per say. Lets warn the user that forking
can have an unexpected side effect.

Contributes to https://tvheadend.org/issues/6296

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1114,6 +1114,9 @@ main(int argc, char **argv)
   signal(SIGPIPE, handle_sigpipe); // will be redundant later
   signal(SIGILL, handle_sigill);   // see handler..
 
+  if (opt_fork && !opt_user && !opt_config)
+    tvhwarn(LS_START, "Forking without --user or --config may use unexpected configuration location");
+
   /* Set privileges */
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1114,6 +1114,9 @@ main(int argc, char **argv)
   signal(SIGPIPE, handle_sigpipe); // will be redundant later
   signal(SIGILL, handle_sigill);   // see handler..
 
+  if (opt_fork && !opt_user && !opt_config)
+    tvhwarn(LS_START, "Forking without --user or --config may use unexpected configuration location");
```
chewitt, line 1118: I'd try to use wording that's more instructional than informational, e.g. "Fork with --user and --config to avoid configuration storage in system folders"

oliv3r: Hey Chris, missed your comment and of course not a native english speaker like yourself :) As the MR was open since august, I kind of lost track of it :)

PR #1563, opened 2023-08-21
Update tvhdhomerun_frontend.c

github-user1979:
To stop the tuner set the channel to none, otherwise the tuner LEDs will flash on some models (e.g. HDHR3-4DC) from firmware release 20200907 and later

References:
https://info.hdhomerun.com/info/hdhomerun_config#tuning_a_physical_channel

https://info.hdhomerun.com/info/prime#led_status

https://www.silicondust.com/support/downloads/firmware-changelog/

### DIFF
--- a/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
+++ b/src/input/mpegts/tvhdhomerun/tvhdhomerun_frontend.c
@@ -485,6 +485,12 @@ tvhdhomerun_frontend_stop_mux
   hfe->hf_status = 0;
   hfe->hf_ready = 0;
 
+/* To stop the tuner set the channel to none, otherwise the tuner LEDs will flash on some models (e.g. HDHR3-4DC) from firmware release 20200907 and later */
+  tvhdebug(LS_TVHDHOMERUN, "setting channel to none");
+  tvh_mutex_lock(&hfe->hf_hdhomerun_device_mutex);
+  hdhomerun_device_set_tuner_channel(hfe->hf_hdhomerun_tuner, "none");
+  tvh_mutex_unlock(&hfe->hf_hdhomerun_device_mutex);
+  
   mtimer_arm_rel(&hfe->hf_monitor_timer, tvhdhomerun_frontend_monitor_cb, hfe, sec2mono(2));
 }
 


github-user1979:
Can you please add this commit to the base branch? Thank you!

Flole998:
Please explain how/why this doesn't introduce a race condition. Isn't it possible with the way this PR is done that a tuner is already tuned to a new channel and then your new code tunes to "none" afterwards?

github-user1979:
It works like a charm, I already did several tests. And now the LED tuner isn't blinking anymore when the stream stops.
It's the same behaviour like the original BDA driver from Silicondust for Windows and internal DLNA server of the HDHomeRun.
The code in Tvheadend stops the mux at the end, and this is only an addition to it. I didn't notice any difference.

This is what Silicondust told me, see last line in bold.

Silicondust Support:
If the application is using tuner locking and sets the tuner target to none but does not release the lock on the tuner, the tuner LEDs will blink. The HDHomeRun will automatically release the lock if the tuner isn't streaming and the application doesn't send any commands for 30 seconds. If it's going longer than 30 seconds, TVHeadend is probably still trying to check the signal stats for some reason, which doesn't really make a lot of sense to us. Normally an application would tune the channel, start up a recurring status monitor, and then when it's done streaming, turn off that monitor, because that status info no longer matters. Without knowing exactly how TVHeadend works internally, we'd suggest that at least as a place to start to look. Maybe it's waiting for the stats to return to nothing, which won't happen if it sets the target to none **but doesn't set the channel to none**.

Flole998:
"Works for me" is not enough to show that code is deterministic and won't introduce a race condition. That's what's needed before merging this, and I'm pretty sure that the way you added this could introduce such a race condition.

github-user1979:
Maybe someone else with HDHomeRun can assist?

github-user1979:
Feel free to change it in the way it should according to you ;-)

Flole998:
> Feel free to change it in the way it should be according to you as long as the tuner LEDs are off after streaming ;-)

I don't have time for that. I was under the assumption that you wanted a quick merge, so I explained what's needed to get it merged. I don't even have the hardware to test it, so even if I would have the time to look into it I couldn't even test it properly.

PR #1564, opened 2023-08-24
Fix htsstr_argsplit

skittlesvampir:
There seemed to be a flaw in the splitting logic when it comes to quotes, e.g.: --output="filename" should be one argument, but htsstr_argsplit treated it as ['--output=', '"filename"] which I think is wrong.

I don't have the setup to test this code so I would appreciate it if someone else does.

### DIFF
--- a/src/htsstr.c
+++ b/src/htsstr.c
@@ -133,6 +133,7 @@ htsstr_argsplit_add
 char **
 htsstr_argsplit(const char *str) {
   int quote = 0;
+  int quote_inbetween = 0; // when quote isn't start of argument
   int inarg = 0;
   const char *start = NULL;
   const char *s;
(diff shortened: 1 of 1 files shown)


Flole998:
At the bottom of the file there is a small test-program. Please add an example that is currently parsed incorrectly and will be parsed correctly with your patch.

skittlesvampir:
I've added a test case but I can't seem to get it running. I've executed the following command in the `src` directory:
`
gcc -g -I ../build.linux/ -o test htsstr.c
`

Then I always (even without any of my changes) this error:
`
/usr/bin/ld: /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
collect2: error: ld returned 1 exit status
`

Can you help me to properly test my testcase.

Flole998:
You need to remove the "#if 0" or change it to "#if 1"

PR #1522, opened 2023-04-09
update to ffmpeg 6.0

uknunknown:
update to ffmpeg 6.0

PR #1573, opened 2023-11-27
ffmpeg update to 6.0.1

CvH:
- updates ffmpeg to 6.0.1
- updates x264 to latest
- updates libvpx to 1.13.1

supersedes #1522 

it builds, otherwise I can't really test it


CvH:
@Flole998 you need to drop the EOL ubuntu kinetic build (from my pov best solution) or add to the action another sed like done for the old debian builds to get all tests to green

taken from [askubutu](https://askubuntu.com/a/1361265)
`sed -i "s/old-releases/archive/g" /etc/apt/sources.list /etc/apt/sources.list.d/*.list`

Flole998:
I know. It should be dropped/replaced by the 2 new ubuntu releases. Just haven't found time yet.

PR #1576, opened 2023-11-29
update ci builds

CvH:
blind changes, can't test locally

- remove EOL distros besides old TLS
- reorder releases by date
- add new build targets Ubuntu 23.04, 23.10 and Debian Trixie
- left out i386 for new targets, Ubuntu already dropped 32bit

not sure if Debian 13 i386 is needed ?

### DIFF
--- a/.github/workflows/build-cloudsmith.yml
+++ b/.github/workflows/build-cloudsmith.yml
@@ -124,7 +124,7 @@ jobs:
             if [ '${{ matrix.distro }}' = 'stretch' ]; then sed -i 's/security.debian.org/archive.debian.org/g' /etc/apt/sources.list; fi
             if [ '${{ matrix.distro }}' = 'stretch' ]; then sed -i '/stretch-updates/d' /etc/apt/sources.list; fi
             case "${{ matrix.distro }}" in
-              ubuntu*|jessie|stretch|buster|bullseye|bookworm)
+              ubuntu*|jessie|stretch|buster|bullseye|bookworm|trixie)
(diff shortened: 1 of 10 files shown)


Flole998:
You're missing the files in the Autobuild directory. Also please squash your commits, this only needs to be a single one.

Flole998:
Also everything that still builds should be left in place. It can still be dropped later on if necessary.

CvH:
- added trixie i386 (its already have a autobuild so guess its wanted)
- added autobuild for lunar and mantic

> Also everything that still builds should be left in place.

just kinetic is dropped

> please squash your commits

I squash as soon you are happy with it 👍  (easier to review)

Flole998:
Alright, 2 more things: Disabling those features in the armv7l file should not be done unless it is necessary (we need to check that, on debian bookworm it doesn't seem to be necessary for example). And the same changes need to be done to the build-cloudsmith file aswell so we are providing packages for these new distros.

CvH:
- dropped kinetic + added trixi to cloudsmith (blindly, no idea if correct)
- temp removed the disable options from lunatic and mantic

CvH:
all green, good to go (i would squash everything before) ?

Flole998:
One thing I just noticed: Please don't delete the kinetic files from the Autobuild directory. If someone wants to build Tvheadend on kinetic it will still be useful for them. As it still builds on a properly setup system those files should be left in place. Other than that, please squash your changes and I will merge them.

CvH:
done

PR #1578, opened 2023-12-01
WebUI: Update donation string as a test to Transifex feed

ProfYaffle:
As well as (perhaps) a minor improvement, this changes one of the strings that ripples through to Transifex. It will confirm how automated the process is, in that direction at least (i.e. is it automated through URL origin tracking, or manual).


Flole998:
Should this be merged? Or do we already know it's working?

PR #1582, opened 2023-12-06
ci: remove the test-compile workflow

chewitt:
~Rename the workflow in `test-compile.yml` so runs can be distinguished from the main `build-yml` CI workflow. Is this supposed to be active?~

PR updated to drop the workflow following conversation in team chat.


Flole998:
I think this used to provide the status for the logo in the readme.md file aswell. The repo builds should be used there instead I think.

PR #1585, opened 2023-12-09
Makefile.ffmpeg nvenc update

Saentist:
FFNVCODEC_VER  = 11.1.5.0 -> 12.1.14.0

### DIFF
--- a/Makefile.ffmpeg
+++ b/Makefile.ffmpeg
@@ -102,11 +102,11 @@ LIBOPUS_TB     = $(LIBOPUS).tar.gz
 LIBOPUS_URL    = https://archive.mozilla.org/pub/opus/$(LIBOPUS_TB)
 LIBOPUS_SHA1   = ed226536537861c9f0f1ef7ca79dffc225bc181b
 
-FFNVCODEC_VER  = 11.1.5.0
+FFNVCODEC_VER  = 12.1.14.0
 FFNVCODEC      = nv-codec-headers-$(FFNVCODEC_VER)
 FFNVCODEC_TB   = $(FFNVCODEC).tar.gz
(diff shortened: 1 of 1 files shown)


CvH:
just saying, that breaks the NV usage at old outdated EOL distros that are activly supported (for whatever reasons) by Tvh
from a quick look just Ubuntu 20.04 supports the >=530 driver

11.1.5.0
> Minimum required driver versions:
> Linux: 470.57.02 or newer

12.1.14.0
> Minimum required driver versions:
> Linux: 530.41.03 or newer

Saentist:
FFMpeg greater then 4 broke already lot of old OS's

Maby some IF>THEN>ELSE strategy in building scenarios needed.
Linear building is not best solution for all OS's and Architectures.

CvH:
> Maby some IF>THEN>ELSE strategy in building scenarios needed.

or in general just drop support for EOL distros
no idea if there is a reasonable demand from Ubuntu 14/16/18.04 users that need beleeding edge Tvh builds and are not able to build their own

Flole998:
Old distros are provided mainly for repurposing old devices that can't really be upgraded (like NAS systems, routers, receivers and whatnot), if nvenc breaks for those it's not a big deal.

Saentist:
> > Maby some IF>THEN>ELSE strategy in building scenarios needed.
> 
> or in general just drop support for EOL distros no idea if there is a reasonable demand from Ubuntu 14/16/18.04 users that need beleeding edge Tvh builds and are not able to build their own

I personally use 14 and 16 just because newer OS kernel (4.4+) do not support driver building drivers for some tuners

Flole998:
Please fix the failing builds.

PR #1590, opened 2023-12-21
ci: rename build.yml > build-ci.yml

chewitt:
Rename `build.yml` to `build-ci.yml` so it's more distinctly named. Add pull_request exclusions for workflow and translation file changes which don't need to be automatically build tested.


Flole998:
We should still run it on workflow file changes, otherwise changes like these could theoretically break the CI and it will only be noticed on the next submitted PR.

chewitt:
I dropped the content changed - only the file rename now.

PR #1592, opened 2023-12-25
ci: fix Cloudsmith upload for distros with Python 3.5

wywywywy:
Xenial (and possibly others) builds have been failing for a while, because the Python 3.5 on it isn't compatible with the latest Pip.

This PR changes that it only upgrades Pip if Python is 3.6 or higher, otherwise rely on the system Pip.


Flole998:
Initially there was an attempt to replace the entire upload script with something better, but the effort seems to have stalled for now. Let's use this until the new script is ready.

PR #1595, opened 2024-01-08
ICAM update and fix

pimzand:


Flole998:
You are now closing the library while you're still trying to use a function from it. That can't be right and will cause crashes.

pimzand:
I guess this is safe, because dlopen has not opened a specific library, instead it has acquired the global symbol table by specifying NULL as the shared library. So closing it does not close any dynamically opened shared library. The handle to the function we need remains valid after closing the dlopen handle.

It has been working for me for over 24 hours.

This is just to keep coverity happy. Apparently it does not want the dlopen handle to remain open.

Other solutions would be:

- always dlopen/dlsym/run function/dlclose for every call to the function. That would cost performance, I guess.
- use dlsym(dlopen(NULL,...),...) without storing the dlopen handle in a variable. Assuming dlopen(NULL,..) never fails.
- declare the dlopen handle as static, telling coverity we intentionally want to keep the handle

pimzand:
Yes, excellent, this works. And really cleans up the code. Will prepare an update to the PR.

Flole998:
Can you squash your commits? Then I can merge them once all tests are finished.

pimzand:
Apparently RTLD_DEFAULT is not universal. Waiting for CI of the workaround commit. Will squash after success.

Flole998:
Hm, this approach lacks error checking and whatnot, the old approach was better. I'm not sure what the best fix for coverity is though.

pimzand:
In the unlikely event that dlopen(NULL,...) would fail, it would pass NULL as a handle to dlsym, which in turn would fail. And that error is being checked.

In the old approach, we just need to convince coverity that dlopen is run no more than once, so no resources are being leaked.

pimzand:
On my system, /usr/include/dlfcn.h contains

`#define RTLD_DEFAULT    ((void *) 0)`

It looks like that before RTLD_DEFAULT was introduced, it was custom to pass NULL to dlsym as the first argument to achieve the same. In that case we would only need to define RTLD_DEFAULT if it is not defined.

Van anyone do "man dlsym" on  a system where RTLD_DEFAULT is missing?

pimzand:
See https://github.com/jphickey/osal/blob/1cf85b2a1662d22bad70d4b9797e0c52e515eb0c/src/os/portable/os-impl-posix-dl-symtab.c#L46-L64

Flole998:
That definition seems to work well. I think we can get rid of the `ifndef`

pimzand:
Oops, I seriously messed up this PR by a squash attempt. Will close this one and create two separate ones in new branches.

PR #1596, opened 2024-01-10
descrambler: avoid dlopen()

pimzand:


pimzand:
I fail to see how the failure on ubuntu:focal is related to this PR.

Flole998:
It's not, I wanted to re-run it earlier when I saw it but I need to wait until all jobs are finished. Now I'll trigger a re-run.

PR #1598, opened 2024-01-21
update to ffmpeg 6.1.1

uknunknown:
update to ffmpeg 6.1.1


uknunknown:
I verified on two platforms with Intel CPU (with VAAPI transcoding): Atom x5-Z8350 and NUC11 (11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz).
More tests are welcome :) .

PR #1601, opened 2024-01-23
ci: fix broken cloudsmith python-pip

wywywywy:


Flole998:
Could you please check your PRs before you submit them? Will this fix it or will this need yet another PR to fix something?

wywywywy:
> Could you please check your PRs before you submit them? Will this fix it or will this need yet another PR to fix something?

The last MR wasn't really ready. I should have set it to draft. That was my bad.

wywywywy:
@Flole998 I'd like to add the Cloudsmith script to the CI workflow pipeline, but with `--dry-run` so we can test it without uploading anything. But even dry running needs a valid API token. Can I get your permission to carry on?

**If yes** - I'll replace the `MOCK_API_TOKEN` with the API token variable defined in this repo

**If no** - I'll remove cloudsmith.sh script from the CI workflow pipeline in the PR

Please let us know which way you prefer. And sorry about the last PR.

Flole998:
Sure, go ahead.

wywywywy:
@Flole998 Ok this PR is ready for review now.

I have since realised that on GitHub, forks aren't allowed to use secrets from the upstream repo anyway, so I've changed the way to dry-run the script, now without needing an API key.

I've resolved the dependencies for various Python versions to get the Cloudsmith CLI to install. Except Trusty - that's way too old and beyond saving.

Of course, without merging to main I cannot be 100% sure all artifacts will upload successfully. But at least it should be better than when we started.

Flole998:
Please squash your commits and add lunar and mantic to the OS list, then those 2 will likely upload successfully aswell.

Flole998:
One thing I don't like about this is that we've now a broken CI if this is merged, so it's no longer easily possible on other PRs to see if they break something or not.

chewitt:
> I've resolved the dependencies for various Python versions to get the Cloudsmith CLI to install. Except Trusty - that's way too old and beyond saving.

For the sake of asking: what's the specific issue with Trusty?

wywywywy:
> Please squash your commits and add lunar and mantic to the OS list, then those 2 will likely upload successfully aswell.

Lunar & Mantic added now. I'll squash the commits when there are no more changes

> One thing I don't like about this is that we've now a broken CI if this is merged, so it's no longer easily possible on other PRs to see if they break something or not.

We can revert the change and remove the cloudsmith.sh from CI if you prefer. Or we can remove Trusty from CI so that it doesn't break any more. Please let us know which direction you want to take.

> For the sake of asking: what's the specific issue with Trusty?

It's just a case of I've not found the right combination of dependencies to use on Trusty. To be honest I think all this can be avoided if the script was rewritten to use venv.

Flole998:
We could try the (risky) step to tell pip to overwrite the system python files. That may or may not work to fix trusty aswell. As this is a short-living CI container it doesn't really matter if we break it or make it unreliable 5 seconds before its stopped and deleted.

Flole998:
I just cancelled the run and I will run trusty manually so we can get results faster

wywywywy:
> I just cancelled the run and I will run trusty manually so we can get results faster

@Flole998 I think we're close now. Can you re-run the failed jobs please? They seem to be transient and unrelated to the change.

wywywywy:
I think we're ready. All jobs have passed and the commits are squashed.

Flole998:
There are still 16 commits, it should be squashed into a single one.

wywywywy:
My apologies. Not sure how I messed up the squash before, but I should have checked first. Anyway it's squashed now.

PR #1606, opened 2024-01-29
ci: use CURL for cloudsmith.sh and enable RPM upload

6007135:
Per discussion on IRC and Discord.

This change requires adding three repository variables:
`CLOUDSMITH_ORG`
`CLOUDSMITH_OWNER`
`CLOUDSMITH_REPO`

### DIFF
--- a/support/cloudsmith.sh
+++ b/support/cloudsmith.sh
@@ -49,47 +49,51 @@ case $OSPREFIX$TARGET in
         OS="ubuntu";;
     raspios*)
         OS="raspbian";;
+    37|38|39)
+        OS="fedora";;
+    40|41)
+        echo "Fedora 40 and 41 (current rawhide) are not (yet) supported by Cloudsmith" && exit;;
     *) echo "OS $TARGET could not be recognized" && exit 1;;
 esac
 
 export LC_ALL=C.UTF-8
 export LANG=C.UTF-8
-export DEBIAN_FRONTEND=noninteractive
 
-# Since Python 3.11, this flag is required to install pip packages globally
-export PIP_BREAK_SYSTEM_PACKAGES=1
-    
-apt install --force-yes -y python3.7 || true
-
-# Use ensurepip to bootstrap pip if available, else install from apt
-python3 -m ensurepip || apt install --force-yes -y python3-pip || apt install --force-yes -y python-pip
-
-# Get the major and minor version of the installed Python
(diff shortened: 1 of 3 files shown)


6007135:
I just realized this failed on buster armv7 with a CURL certificate error [~]. Interestingly the other buster runners did not have this issue. I'll look into if I need to download and pass the latest CA cert manually. I would have assumed so, but only having the issue on one buster runner confuses me.

[~] https://github.com/6007135/tvheadend/actions/runs/7691769756

6007135:
Fixed by adding a `update-ca-certificates --fresh` per Christian's suggestion.

6007135:
All CI runners succeed on my fork (different commit hash because it was pre rebase).

https://github.com/6007135/tvheadend/actions/runs/7811271207

Flole998:
You need to do something to line 36 of your script: The variable `package` is not assigned there yet (and all the other variables are always empty. Maybe move it down to the loop where the packages are actually pushed and do a `continue` (assuming that exists in bash, not sure).

6007135:
Good point. I should have caught that. Fixed.

support/cloudsmith.sh
```diff
@@ -49,47 +49,51 @@ case $OSPREFIX$TARGET in
...
 for package in "${FILEARRAY[@]}"; do
+    EXTENSION=${package##*.}
+    PKGBASENAME=$(basename $package)
+    # dryrun exit is performed as late as possible to catch any earlier potential issues
     if [ $DRYRUN = "1" ]; then
-        echo "DRYRUN MODE: Skip pushing tvheadend/tvheadend/$OS/$TARGET $package"
-    else
-        python3 /usr/local/bin/cloudsmith push deb "tvheadend/tvheadend/$OS/$TARGET" $package || python /usr/local/bin/cloudsmith push deb "tvheadend/tvheadend/$OS/$TARGET" $package || cloudsmith push deb "tvheadend/tvheadend/$OS/$TARGET" $package
+        echo "DRYRUN MODE: Skip pushing $OS $TARGET package $package"
+        exit 0
```
Flole998, line 77: exit 0 is not really the right thing here, continue would be better as there might be multiple packages.

6007135: Ah, of course. Eventually I will learn to proofread a little better after cutting and pasting.

PR #1608, opened 2024-01-30
Add missing Lithuanian string template

ProfYaffle:
We have template discrepancies between `tvheadend/intl` and the sub-directories contain the .po strings. Language strings/templates for Lithuanian (lt) are in `tvheadend/intl` and `tvheadend/intl/js`, but are missing from `tvheadend/intl/docs`.

The language exists on Transifex, so I'm assuming that it's an import error (i.e. no target into which to import the translated strings). There appears to be nothing in the code or Makefile that creates these files otherwise.

This forces the creation of a template by simply copying the master `tvheadend/intl/docs/tvheadend.doc.pot` to `tvheadend/intl/docs/tvheadend.doc.lt.po`.


chewitt (approved):
Thanks, let's see what happens :)

PR #1623, opened 2024-02-04
Add support for 12-hour custom date formats

chewitt:
Squashed and rebased version of #1528 from @pmatis

### DIFF
--- a/src/webui/static/app/tvheadend.js
+++ b/src/webui/static/app/tvheadend.js
@@ -1387,38 +1387,48 @@ tvheadend.toLocaleFormat = function()
 	return tvh_locale_lang.replace('_','-');
 };
 
-tvheadend.toCustomDate = function(date, format) //author: meizz, improvements by pablozg
-{
-    if(/([%][MmsSyYdhq]+)/.test(format)){
-        var o = {
-            "\%M+" : date.getMonth()+1, //month
-            "\%d+" : date.getDate(),    //day
-            "\%h+" : date.getHours(),   //hour
-            "\%m+" : date.getMinutes(), //minute
-            "\%s+" : date.getSeconds(), //second
-            "\%q+" : Math.floor((date.getMonth()+3)/3),  //quarter
-            "\%S" : date.getMilliseconds() //millisecond
+tvheadend.toCustomDate = function(date, format) {
+    if (/(%[MmsSyYdhHIpPq]+)/.test(format)) {
+        const o = {
+            "%M+": date.getMonth() + 1,
+            "%d+": date.getDate(),
+            "%h+": date.getHours(),
+            "%H+": date.getHours(),
+            "%I+": date.getHours() % 12 || 12,
+            "%p": date.getHours() >= 12 ? "PM" : "AM",
+            "%P": date.getHours() >= 12 ? "pm" : "am",
+            "%m+": date.getMinutes(),
+            "%s+": date.getSeconds(),
+            "%q+": Math.floor((date.getMonth() + 3) / 3),
+            "%S": date.getMilliseconds()
+        };
+
+        format = format.replace(/(%[yY]+)/, (date.getFullYear() + "").substr(5 - RegExp.$1.length))
+                       .replace(/(%MMMM)/, date.toLocaleDateString(toLocaleFormat(), { month: 'long' }))
+                       .replace(/(%MMM)/, date.toLocaleDateString(toLocaleFormat(), { month: 'short' }))
+                       .replace(/(%dddd)/, date.toLocaleDateString(toLocaleFormat(), { weekday: 'long' }))
+                       .replace(/(%ddd)/, date.toLocaleDateString(toLocaleFormat(), { weekday: 'short' }));
+
+        for (const k in o) {
+            if (new RegExp("(" + k + ")").test(format)) {
+                format = format.replace(RegExp.$1, RegExp.$1.length == 2 ? o[k] : ("00" + o[k]).substr(("" + o[k]).length));
+            }
         }
 
-        if(/(\%[yY]+)/.test(format)) format=format.replace(RegExp.$1, (date.getFullYear()+"").substr(5 - RegExp.$1.length));
-
-        if(/(\%MMMM)/.test(format)) format=format.replace(RegExp.$1, (date.toLocaleDateString(tvheadend.toLocaleFormat(), {month: 'long'})));
-
-        if(/(\%MMM)/.test(format)) format=format.replace(RegExp.$1, (date.toLocaleDateString(tvheadend.toLocaleFormat(), {month: 'short'})));
-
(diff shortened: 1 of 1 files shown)


Flole998:
AI suggests this as functional equivalent code that is better to read:

```
tvheadend.toCustomDate = function(date, format) {
    const options = {
        month: '2-digit', 
        day: '2-digit', 
        year: 'numeric', 
        hour: '2-digit', 
        minute: '2-digit', 
        second: '2-digit',
        hour12: false
    };

    if (/([%][MmsSyYdhHIq]+)/.test(format)) {
        const o = {
            "\%M+": date.getMonth() + 1, 
            "\%d+": date.getDate(),    
            "\%h+": date.getHours(),   
            "\%H+": date.getHours(),   
            "\%I+": date.getHours() % 12 || 12,   
            "\%p": date.getHours() >= 12 ? "PM" : "AM",   
            "\%P": date.getHours() >= 12 ? "pm" : "am",   
            "\%m+": date.getMinutes(), 
            "\%s+": date.getSeconds(), 
            "\%q+": Math.floor((date.getMonth() + 3) / 3),  
            "\%S": date.getMilliseconds() 
        };

        format = format.replace(/(\%[yY]+)/, (date.getFullYear() + "").substr(5 - RegExp.$1.length))
                       .replace(/(\%MMMM)/, date.toLocaleDateString(tvheadend.toLocaleFormat(), { month: 'long' }))
                       .replace(/(\%MMM)/, date.toLocaleDateString(tvheadend.toLocaleFormat(), { month: 'short' }))
                       .replace(/(\%dddd)/, date.toLocaleDateString(tvheadend.toLocaleFormat(), { weekday: 'long' }))
                       .replace(/(\%ddd)/, date.toLocaleDateString(tvheadend.toLocaleFormat(), { weekday: 'short' }));

        for (const k in o) {
            if (new RegExp("(" + k + ")").test(format)) {
                format = format.replace(RegExp.$1, RegExp.$1.length == 2 ? o[k] : ("00" + o[k]).substr(("" + o[k]).length));
            }
        }
        
        return format;
    } else {
        return date.toLocaleString(tvheadend.toLocaleFormat(), options);
    }
};

```

We need to check if it really is functional equivalent though, and then see if it increases the maintainability score.

oliv3r:
Looking purely at the change, I wouldn't use what the AI tells us for THIS merge request. Once merged, it does make sense to listen to the AI.

Keeping this Merge requested focused on the change itself (which just adds a few lines) certainly has merit in itself.

Flole998:
I think the main question is how we want to handle the sonarcloud failures/issues. If we do a "merge now, fix later" that will lead to a "merge now, fix never". So either the restructuring needs to be done first, or all in one step. Although I don't really like multiple commits in a single PR this might be a case where it makes sense and where a "merge" and then "restructure/cleaup" commit is done.

Flole998:
Why do we have `hour` in there twice now? And was that `%P` and `%p` ever tested? Because in the way it was done initially it can not work. I changed the AI suggestion to this and verified it now:

```
toCustomDateAI = function(date, format) {
    if (/([%][MmsSyYdhHIpPq]+)/.test(format)) {
        const o = {
            "%M+": date.getMonth() + 1, 
            "%d+": date.getDate(),    
            "%h+": date.getHours(),   
            "%H+": date.getHours(),   
            "%I+": date.getHours() % 12 || 12,   
            "%p": date.getHours() >= 12 ? "PM" : "AM",   
            "%P": date.getHours() >= 12 ? "pm" : "am",   
            "%m+": date.getMinutes(), 
            "%s+": date.getSeconds(), 
            "%q+": Math.floor((date.getMonth() + 3) / 3),  
            "%S": date.getMilliseconds() 
        };

        format = format.replace(/(\%[yY]+)/, (date.getFullYear() + "").substr(5 - RegExp.$1.length))
                       .replace(/(\%MMMM)/, date.toLocaleDateString(toLocaleFormat(), { month: 'long' }))
                       .replace(/(\%MMM)/, date.toLocaleDateString(toLocaleFormat(), { month: 'short' }))
                       .replace(/(\%dddd)/, date.toLocaleDateString(toLocaleFormat(), { weekday: 'long' }))
                       .replace(/(\%ddd)/, date.toLocaleDateString(toLocaleFormat(), { weekday: 'short' }));

        for (const k in o) {
            if (new RegExp("(" + k + ")").test(format)) {
                format = format.replace(RegExp.$1, RegExp.$1.length == 2 ? o[k] : ("00" + o[k]).substr(("" + o[k]).length));
            }
        }
        
        return format;
    } else {
        const options = {
            month: '2-digit', 
            day: '2-digit', 
            year: 'numeric', 
            hour: '2-digit', 
            minute: '2-digit', 
            second: '2-digit',
            hour12: false
        };
        return date.toLocaleString(toLocaleFormat(), options);
    }
};
```

We can now either fix the original commit and remove the double-hour and add the pP to the regex, or just do the AI change and hopefully satisfy the static analysis with it.

oliv3r:
> I think the main question is how we want to handle the sonarcloud failures/issues. If we do a "merge now, fix later" that will lead to a "merge now, fix never". So either the restructuring needs to be done first, or all in one step. Although I don't really like multiple commits in a single PR this might be a case where it makes sense and where a "merge" and then "restructure/cleaup" commit is done.

I'd be careful here. Yes, you are right on a proper maintained, with lots of churn. You can spend the time to refactor things.

But if you now request every improvement through a refactor cycle first, the project might loose even more steam.

Obviously, if someone comes a long and wants to refactor improve things, that's of course great and super helpful.

With project management, there's always a balance to be struck, with reducing techdept/refactor things, and moving forward.

Flole998:
Sonarqube is right again, the unnecessary escape characters should be removed. So should that single character class. And that should yield to the required maintainability rating. It might be worth checking the deprecation warnings though. And to be honest: The high cognitive complexity is also a valid point, especially that for-loop could probably be made easier to read (I'm fine with leaving that one as is though for now).

src/webui/static/app/tvheadend.js
```diff
@@ -1387,38 +1387,48 @@ tvheadend.toLocaleFormat = function()
...
-            "\%S" : date.getMilliseconds() //millisecond
-        }
-
-        if(/(\%[yY]+)/.test(format)) format=format.replace(RegExp.$1, (date.getFullYear()+"").substr(5 - RegExp.$1.length));
-
-        if(/(\%MMMM)/.test(format)) format=format.replace(RegExp.$1, (date.toLocaleDateString(tvheadend.toLocaleFormat(), {month: 'long'})));
-
-        if(/(\%MMM)/.test(format)) format=format.replace(RegExp.$1, (date.toLocaleDateString(tvheadend.toLocaleFormat(), {month: 'short'})));
+tvheadend.toCustomDate = function(date, format) {
+    if (/([%][MmsSyYdhHIpPq]+)/.test(format)) {
```
Flole998, line 1391: Replace the `[%]` with `%`, no need for a character class when there is only a single character.

Flole998:
Looks great, lets squash, I'll test again and then we can merge it.

chewitt:
Squashed.

PR #1627, opened 2024-02-06
dvr: Fix incorrect usage of `strerror`

i-Pear:
`strerror` takes the `errno` directly as its argument, negating it will result in an "Unknown error".


Flole998:
But it is negated before in https://github.com/tvheadend/tvheadend/blob/6409a6382f1ded18cd6f21649519879c410eb8ab/src/utils.c#L949?

i-Pear:
> But it is negated before in
> 
> https://github.com/tvheadend/tvheadend/blob/6409a6382f1ded18cd6f21649519879c410eb8ab/src/utils.c#L949
> 

Yes, r = -errno; But the modified line was `strerror(-errno)`, which is unrelated to `r`.
Since `errno` is never negative, `strerror(-errno)` will return "Unknown error" all the time.

Flole998:
Ohh,  I think the intention was to use `-r` here, but I guess this also works.

PR #1639, opened 2024-02-11
CI: Ensure we clone the whole repo

oliv3r:
We have to make sure we clone the whole repo, so that `git describe` works as expected. Without it, we get version 0.0.0, not what we want.

### DIFF
--- a/.github/workflows/container-build.yaml
+++ b/.github/workflows/container-build.yaml
@@ -32,6 +32,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


oliv3r:
Haven't tested this yet, but then again, can't test it locally either :p

Flole998:
> Haven't tested this yet, but then again, can't test it locally either :p

I've skipped all pre-merge tests aswell, we test in production! 🌚😂

oliv3r:
It's a tiny change, and the risk is extremely low. My indentation could be wrong, or tabs/spaces. YAML loves that stuff :p but overal I'm not too scared :) Thanks @Flole998 !

PR #1641, opened 2024-02-17
ci: update upload-artifact@v3 to upload-artifact@v4

chewitt:
This should silence the following warnings during workflow runs:

 `Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -51,7 +51,7 @@ jobs:
           DEBIAN_FRONTEND=noninteractive apt-get install --force-yes -y libpcre3-dev || DEBIAN_FRONTEND=noninteractive apt-get install --force-yes -y libpcre2-dev
           AUTOBUILD_CONFIGURE_EXTRA=--enable-ccache\ --enable-ffmpeg_static\ --enable-hdhomerun_static\ --python=python3 ./Autobuild.sh -p raspios
           support/cloudsmith.sh -n -p raspios -f '../tvheadend*.deb'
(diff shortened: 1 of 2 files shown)


Flole998:
IIRC I saw these (or similar) issues when initially setting up GH actions, so I went with the older version. It needs to be addressed at some point though.

chewitt:
One of the v3 > v4 differences is not allowing the same file to be uploaded more than once, so I think we might have some name collisions taking place. If yes, the solution will be to tweak the artefact naming.

Flole998:
If it's the `name` field, then yes, we have a huge deb and a not so huge rpm zip artifact. That should probably be improved/changed.

Flole998:
> IIRC I saw these (or similar) issues when initially setting up GH actions, so I went with the older version. It needs to be addressed at some point though.

I just looked it up again: That was for the checkout action, not for upload artifact.

chewitt:
I'll investigate in the week. I still have relatives in-town so not much time for reading GitHub docs at the moment :)

Flole998:
I'd guess all that's needed is adding a variable to the name for the architecture. Just a guess though and the docs should clear that up. No need to rush though.

Flole998:
You'll probably need to add the architecture to it somehow, otherwise i386 and x64 are conflicting again.

Flole998:
This is now part of #1797

PR #1642, opened 2024-02-17
ci: updates for build-container action

chewitt:
This updates the action versions to resolve the following warning:

`Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, docker/setup-qemu-action@v2, docker/setup-buildx-action@v2, docker/login-action@v2, docker/metadata-action@v4, docker/build-push-action@v4. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`

It also adds concurrency so build runs are cancelled if the PR is force-push updated before they complete, and it renames the file from .yaml to .yml for consistency with other actions.


Flole998:
The concurrency change will likely have unintended side-effects: Here we have a single file for PRs and Repo publishing, when there are fast merges not all builds will be published to the repo. That's not intended, the repo should contain all builds.

chewitt:
The concurrency criteria includes the PR: `group: ${{ github.workflow }}-${{ github.event.pull_request.number }}` so force-pushing to a PR will cancel existing jobs for that PR and requeue with the latest changes. If you have other PR's building at the same time these won't be impacted as they're tracked under a separate identifier.

Flole998:
Yes I understand that, but this job is special: It does not only run on PRs, but also on master (with the additional step of publishing the container to the repo). So what will happen when 2 PRs are quickly merged into master? Will the second one cancel the first one? Or do they run one after the other?

chewitt:
Closing as superseded by recent CI updates

Issue #1638, opened 2024-02-10
Add ERRNO to Log Entries Regarding File System Errors


DeltaMikeCharlie:
### Description

Add the ERRNO to log messages regarding file system error.

Triggered by this forum post:
https://tvheadend.org/d/8520-change-default-folders-for-tv-recordings-and-timeshift-on-ubuntu-2204

https://github.com/tvheadend/tvheadend/blob/b91587037c6099e77d233877162e36138c62e5b2/src/dvr/dvr_rec.c#L127

```
  if(stat(de->de_config->dvr_storage, &st) || !S_ISDIR(st.st_mode)) {
    tvherror(LS_DVR, "the directory '%s' is not accessible", de->de_config->dvr_storage);
    ret = -EIO;
    goto _return;
  }
```

Adding the ERRNO plus the ERRNO description would have provided the user with more troubleshooting information.

### Help and Guidance

This is a placeholder following a discussion with @Flole998.

### Acceptance

- [X] The proposal will be developed and maintained by me

DeltaMikeCharlie:
Note: The strerror_r() function is similar to strerror(), but is thread safe.

Flole998:
strerror is used everywhere else, we should continue using it even though it could theoretically cause incorrect log messages when multiple error messages are printed at the exact same time and when the implementation uses a single, static buffer. Usually that is not an issue though.

PR #1648, opened 2024-02-19
Add some ERRNOs for DVR

DeltaMikeCharlie:
Discussed in Issue: https://github.com/tvheadend/tvheadend/issues/1638

Log some error numbers and error messages for file system errors in the DVR module.

Modified error messages:

```
2024-02-19 13:11:44.594 [  ERROR] dvr: '/home/tvhtest/text.txt' is not a directory. (ERRNO: 2)
2024-02-19 13:12:27.654 [  ERROR] dvr: Directory '/home/tvhtest/not_found' not accessible.  ERRNO: 2, 'No such file or directory'
2024-02-19 13:30:41.169 [  ERROR] dvr: Directory '/home/dmc/recordings' not accessible.  ERRNO: 13, 'Permission denied'
2024-02-19 13:39:03.183 [  ERROR] dvr: unable to stat file '/home/dmc/Jazz Lunch.ts' (ERRNO: '13')
```
This may help users more easily identify the cause of their error.

Fixes: #1638

### DIFF
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -82,6 +82,7 @@ dvr_rec_subscribe(dvr_entry_t *de)
   uint32_t rec_count, net_count;
   int ret = 0, pri, c1, c2;
   struct stat st;
+  int stat_ret;
 
   assert(de->de_s == NULL);
   assert(de->de_chain == NULL);
@@ -124,8 +125,18 @@ dvr_rec_subscribe(dvr_entry_t *de)
     }
   }
 
-  if(stat(de->de_config->dvr_storage, &st) || !S_ISDIR(st.st_mode)) {
-    tvherror(LS_DVR, "the directory '%s' is not accessible", de->de_config->dvr_storage);
+  stat_ret = stat(de->de_config->dvr_storage, &st);
+
+  //If the stat() failed, show the error message.
+  if(stat_ret != 0) {
+    tvherror(LS_DVR, "Directory '%s' not accessible: %s", de->de_config->dvr_storage, strerror(errno));
+    ret = -EIO;
+    goto _return;
+  }
+
+  //If the stat() worked, but the path is not a directory.
+  if(!S_ISDIR(st.st_mode) && stat_ret == 0) {
+    tvherror(LS_DVR, "'%s' is not a directory.", de->de_config->dvr_storage);
     ret = -EIO;
     goto _return;
   }
--- a/src/config.c
+++ b/src/config.c
@@ -135,7 +135,7 @@ config_migrate_v1_dvb_svcs
         htsmsg_add_str(svc, "svcname", str);
       if ((str = htsmsg_get_str(e, "provider")))
         htsmsg_add_str(svc, "provider", str);
-      if (!(htsmsg_get_u32(e, "type", &u32))) 
+      if (!(htsmsg_get_u32(e, "type", &u32)))
         htsmsg_add_u32(svc, "dvb_servicetype", u32);
       if (!htsmsg_get_u32(e, "channel", &u32))
         htsmsg_add_u32(svc, "lcn", u32);
@@ -143,9 +143,9 @@ config_migrate_v1_dvb_svcs
         htsmsg_add_u32(svc, "enabled", u32 ? 0 : 1);
       if ((str = htsmsg_get_str(e, "charset")))
         htsmsg_add_str(svc, "charset", str);
-      if ((str = htsmsg_get_str(e, "default_authority"))) 
+      if ((str = htsmsg_get_str(e, "default_authority")))
         htsmsg_add_str(svc, "cridauth", str);
-  
+
       // TODO: dvb_eit_enable
 
       hts_settings_save(svc, "input/linuxdvb/networks/%s/muxes/%s/services/%s",
@@ -185,7 +185,7 @@ config_migrate_v1_dvb_network
     "fec_lo",
     "fec"
   };
-    
+
 
   /* Load the adapter config */
   if (!(tun = hts_settings_load("dvbadapters/%s", name))) return;
@@ -293,7 +293,7 @@ config_migrate_v1_dvr ( const char *path, htsmsg_t *channels )
   htsmsg_t *c, *e, *m;
   htsmsg_field_t *f;
   const char *str;
-  
+
   if ((c = hts_settings_load_r(1, path))) {
     HTSMSG_FOREACH(f, c) {
       if (!(e = htsmsg_field_get_map(f))) continue;
@@ -323,7 +323,7 @@ config_migrate_v1_epggrab ( const char *path, htsmsg_t *channels )
   htsmsg_field_t *f, *f2;
   const char *str;
   uint32_t u32;
-  
+
   if ((c = hts_settings_load_r(1, path))) {
     HTSMSG_FOREACH(f, c) {
       if (!(e = htsmsg_field_get_map(f))) continue;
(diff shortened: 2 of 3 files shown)


src/dvr/dvr_rec.c
```diff
@@ -83,6 +83,10 @@ dvr_rec_subscribe(dvr_entry_t *de)
   int ret = 0, pri, c1, c2;
   struct stat st;
```
Flole998, line 85: No need for a space character here

src/dvr/dvr_rec.c
```diff
@@ -124,8 +128,21 @@ dvr_rec_subscribe(dvr_entry_t *de)
...
+  if(statRet != 0) {
+    strerror_r(errSaved, errText, sizeof(errText));
+    tvherror(LS_DVR, "Directory '%s' not accessible.  ERRNO: %d, '%s'", de->de_config->dvr_storage, errSaved, errText);
+    ret = -EIO;
+    goto _return;
+  }
+
+  //If the stat() worked, but the path is not a directory.
+  if(!S_ISDIR(st.st_mode) && statRet == 0) {
+    tvherror(LS_DVR, "'%s' is not a directory. (ERRNO: %d)", de->de_config->dvr_storage, errSaved);
```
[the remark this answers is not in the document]
Flole998: It should be removed, as errno will not contain any relevant info in this case anyways.

src/dvr/dvr_rec.c
```diff
@@ -124,8 +128,21 @@ dvr_rec_subscribe(dvr_entry_t *de)
...
 
-  if(stat(de->de_config->dvr_storage, &st) || !S_ISDIR(st.st_mode)) {
-    tvherror(LS_DVR, "the directory '%s' is not accessible", de->de_config->dvr_storage);
+  errSaved = 0;
+  statRet = stat(de->de_config->dvr_storage, &st);
+  errSaved = errno;
+
+  //If the stat() failed, show the ERRNO.
+  if(statRet != 0) {
+    strerror_r(errSaved, errText, sizeof(errText));
```
Flole998, line 137: Use `strerror` so you don't need those 1024 byte buffer.

src/dvr/dvr_rec.c
```diff
@@ -124,8 +128,21 @@ dvr_rec_subscribe(dvr_entry_t *de)
...
+    tvherror(LS_DVR, "Directory '%s' not accessible.  ERRNO: %d, '%s'", de->de_config->dvr_storage, errSaved, errText);
```
Flole998, line 138: I'd do something like `"Directory '%s' not accessible (error %d): %s"`. Although there is no need to print out the error number if it has been translated, so a simple `"Directory '%s' not accessible: %s"` would be sufficient, which would then lead to a log message like `...not accessible: No such file or directory`.

Flole998:
Adding a "Fixes: <Issue Number>" like I did now will make sure the issue is automatically closed once the PR is merged.

src/dvr/dvr_rec.c
DeltaMikeCharlie: > Use `strerror` so you don't need those 1024 byte buffer.

True, however, `strerror` appears not to be thread-safe.  This is why I used `strerror_r` instead.

https://man7.org/linux/man-pages/man3/strerror.3.html

Do you still want to use `strerror`?

Flole998: It's not guaranteed to be thread-safe, it actually depends on the implementation. The "worst" thing that can happen is, that 2 error messages are logged at the same time from multiple threads and one message is not shown and the other one is shown twice (so strerror returns the wrong string). It's pretty unlikely to have a scenario where 2 errors are logged at the same time, so that's fine.

Also some implementations return a pointer while holding all the strings in memory, in that case there is no issue with regards to concurrency.

src/dvr/dvr_rec.c
DeltaMikeCharlie: This is a debugging message in a log file and not intended for general consumption.

The ERRNO is for technical folk to know exactly which error occurred and the text description is to give the user a better idea of what is going on and where to look should they be so inclined.  Both useful pieces of information.

src/dvr/dvr_rec.c
```diff
@@ -124,8 +128,21 @@ dvr_rec_subscribe(dvr_entry_t *de)
     }
   }
 
-  if(stat(de->de_config->dvr_storage, &st) || !S_ISDIR(st.st_mode)) {
-    tvherror(LS_DVR, "the directory '%s' is not accessible", de->de_config->dvr_storage);
+  errSaved = 0;
+  statRet = stat(de->de_config->dvr_storage, &st);
+  errSaved = errno;
```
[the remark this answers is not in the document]
DeltaMikeCharlie: Where it can be easily done, I usually save the ERRNO to a variable immediately after the operation in question.  I do this because I have personally encountered consecutive errors masking the original cause of the error.

However, if this is a mandatory requirement, I will remove the variable.

src/dvr/dvr_rec.c
DeltaMikeCharlie: I will use `strerror` as required.

src/dvr/dvr_rec.c
Flole998: The errno contains exactly the same information as the error string, there is no additional information in it. To make things worse: The errno is platform/architecture dependent, so you need to know which architecture the issue appeared on.

Usually what is done is using `perror` when no "special" logging functions are used, which will lead to the error message (without the errno)  being printed to stderr. There's no need for the actual error code, so it isn't shown.

If the error code would contain additional information that would be a different story.

src/dvr/dvr_rec.c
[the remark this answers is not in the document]
Flole998: In this case there is no chance for consecutive errors as there is no operation accessing errno in between. I'm a little hesitant as such copying always uses additional resources, and I am not sure if the compiler would optimize that out again in this case (which would be alright again).

We should attempt to match the coding style of the existing project, so please have a look and then let's do it exactly like it's done everywhere else. We can change it everywhere in case we don't like it in another PR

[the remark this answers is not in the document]
DeltaMikeCharlie: I will remove the variable.

src/dvr/dvr_rec.c
DeltaMikeCharlie: I will use your format.

DeltaMikeCharlie:
I have made the updates as required.

I also encountered an issue where the config backup failed so I added the error text there too using the same formatting as the DVR.

```
2024-02-20 08:14:35.336 [  ERROR] config: unable to change directory to '/home/dmc/development/TVH/dvr_errno/tvheadend/build.linux': Permission denied
2024-02-20 08:22:34.315 [  ERROR] dvr: '/home/tvhtest/text.txt' is not a directory.
2024-02-20 08:23:49.761 [  ERROR] dvr: Directory '/home/tvhtest/not_found' not accessible: No such file or directory
2024-02-20 08:27:04.979 [  ERROR] dvr: Directory '/home/dmc/recordings' not accessible: Permission denied
2024-02-20 08:39:02.339 [  ERROR] dvr: unable to stat file '/home/dmc/Today.ts': Permission denied
```

src/dvr/dvr_rec.c
```diff
@@ -82,6 +82,7 @@ dvr_rec_subscribe(dvr_entry_t *de)
   uint32_t rec_count, net_count;
   int ret = 0, pri, c1, c2;
   struct stat st;
+  int statRet;
```
Flole998, line 85: camelCase is not usually used for c programming, and doesn't really fit here either, use stat_ret instead.

Flole998:
Just one more thing before this is good to go. Sorry, I didn't catch it last time.

src/dvr/dvr_rec.c
DeltaMikeCharlie: dONE

PR #1654, opened 2024-02-20
Unify coding style of the entire codebase

Flole998:
This needs additional discussion and some things are not quite right yet:

- [ ] Imports are reordered (this could potentially lead to issues)
- [ ] #define-Functions are not interpreted properly (TAILQ_FOREACH for example, the lines after that should be indented.


src/dbus.h
```diff
@@ -46,23 +49,27 @@ void dbus_server_done(void);
...
 static inline void
-dbus_emit_signal(const char *obj_name, const char *sig_name, htsmsg_t *msg) { htsmsg_destroy(msg); }
-static inline void
-dbus_emit_signal_str(const char *obj_name, const char *sig_name, const char *value) { }
-static inline void
-dbus_emit_signal_s64(const char *obj_name, const char *sig_name, int64_t value) { }
+dbus_emit_signal(const char* obj_name, const char* sig_name, htsmsg_t* msg) {
+  htsmsg_destroy(msg);
+}
+static inline void dbus_emit_signal_str(const char* obj_name,
```
Flole998, line 55: This looks odd....

Flole998:
We should allow longer lines, the current settings add many line breaks, also in the existing codebase

Flole998:
The failing builds are due to the include re-ordering. I'll need to read on how to do it properly so that clang-format doesn't break everything.

PR #1572, opened 2023-11-26
hdhomerun: Add HDHomeRun server emulation support for LiveTV only

chkuendig:
Closes https://tvheadend.org/issues/4461

This is an attempt to cleanup the previously rejected #1348:
- I removed all code related to the authentication bypass 
- harmonized the comment format.
- Successfully tested by using IP based authentication and Plex

I kept the previous commit to highlight the new changes, but happy to squash if that's preferred.

PR #1679, opened 2024-03-26
Add SSDP Discovery for HDHomerun Server Emulation

chkuendig:
This is a follow-up to https://github.com/tvheadend/tvheadend/pull/1572. 

Without this PR, Plex eventually (I think after a while or when restarting either plex or tvh) loses the connection and it's not possible to tune to any channels. This was also discussed in https://github.com/tvheadend/tvheadend/pull/1348.

While maintaining [tvhProxy](https://github.com/chkuendig/tvhProxy) I added SSDP to make sure Plex doesn't drop the receiver when it's restarted. The code here is mostly based on the existing discovery for SAT>IP in [src/satip/server.c](https://github.com/tvheadend/tvheadend/blob/master/src/satip/server.c) and what I implemented to get tvhProxy working.

As a neat bonus, this also now allows for automatic configuration in Plex.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2552,9 +2552,10 @@ const idclass_t config_class = {
       .name   = N_("Local IP Address"),
       .desc   = N_("IP of the Docker host. Each HDHomeRun tuner sends data "
                    "to TVheadend through a socket. This lets you define the "
-                   "IP address that HDHomeRun needs to send to. Leave this "
-                   "blank if you want TVheadend to automatically pick an "
-                   "address."),
+                   "IP address that HDHomeRun needs to send to. This also used "
+                   "to set the IP of the server if HDHomeRun Server Emulation " 
+                   "is enabled. Leave this blank if you want TVheadend to "
+                   "automatically pick an address."),
       .off    = offsetof(config_t, local_ip),
       .opts   = PO_HIDDEN | PO_EXPERT,
       .group  = 6
--- a/src/main.c
+++ b/src/main.c
@@ -1316,10 +1316,10 @@ main(int argc, char **argv)
   tvhftrace(LS_MAIN, timeshift_init);
 #endif
   tvhftrace(LS_MAIN, tcp_server_init);
-  tvhftrace(LS_MAIN, webui_init, opt_xspf);
 #if ENABLE_UPNP
   tvhftrace(LS_MAIN, upnp_server_init, opt_bindaddr);
 #endif
+  tvhftrace(LS_MAIN, webui_init, opt_xspf);
   tvhftrace(LS_MAIN, service_mapper_init);
   tvhftrace(LS_MAIN, epggrab_init);
   tvhftrace(LS_MAIN, epg_init);
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -52,6 +52,10 @@
 #include <sys/socket.h>
 #endif
 
+#if ENABLE_HDHOMERUN_SERVER
+#include "upnp.h"
+#endif
+
 enum {
   PLAYLIST_M3U,
   PLAYLIST_E2,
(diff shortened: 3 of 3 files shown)


chkuendig:
Not sure what to make of the SonarCloud issues, I think they all only highlight false positives from what I can tell.

Flole998:
The cognitive complexity one is kinda real, that can be solved better. A while loop with a if statement that results in a break is not really a good idea. Also the nested ifs when checking if ptr is not null aren't necessary, if ptr is null you can simply return, no need to continue in parsing just to fail later on (searchport will be null, so it will return there).

Also I don't like the introduction of the mandatory IP address parameter, I've made an approach to auto-detect it in #1637, if that works something like that should be used instead.

chkuendig:
Fair point on the complexity, this was bastardized from the SAT>IP code. I'll try to clean it up a bit.

Regarding the local IP setting: This isnt mandatory, I copied the code to detect the local IP also over from the SAT>IP server. But since we already have an option to manually set it for HDHR, I thought it makes sense to use that as an option to override the detected IP in scenarios where the automated detection might fail (tunnels, reverse proxies, docker etc).

chkuendig:
@Flole998 hey, I looked at https://github.com/tvheadend/tvheadend/pull/1637 but I'm not sure how you set the Location in `MSG` unless `satip_server_bindaddr` is set. Is it just null in that case? 

-  both https://github.com/tvheadend/tvheadend/blob/1134ca16a7acfc76c5a5c2c046a80fb771e5ad48/src/satip/server.c#L433 and https://github.com/tvheadend/tvheadend/blob/1134ca16a7acfc76c5a5c2c046a80fb771e5ad48/src/satip/server.c#L389 take it from `http_server_ip` to fill in `MSG`

- https://github.com/tvheadend/tvheadend/blob/1134ca16a7acfc76c5a5c2c046a80fb771e5ad48/src/satip/server.c#L1005 copies`satip_server_bindaddr`  to `http_server_ip`

Wouldn't that mean the announce message would lack the critical full location address?

chkuendig:
I'm fine with holding this until #1637 is decided on. I initially wanted to try to implement this change on top of #1637 and test it thoroughly, but I haven't even gotten to fixing the code quality issue due to various private and professional things keeping my busy. So best to probably just hold this for now. 

Once I get to fix the complexity problems, I might as well also adapt it for #1637 and we can decide on how to continue then.

chkuendig:
I had some time to update this. Two changes:
- I noticed a few services that were very flaky in Plex. This is a known issue (see e.g. ttps://github.com/xteve-project/xTeVe/issues/331 ) as HDR usually already does some pre-processing of the streams. Respecting the users settings for the streaming profile let's one to work around this (I just pick the first enabled profile).
- I tried reducing the code complexity in `hdhr_server_upnp_discovery_received` by removing the nested `if`s and the `while(1)`. Didn't get it below the threshold but it's much lower than before.

chkuendig:
> I think I'd prefer to hold this until https://github.com/tvheadend/tvheadend/pull/1637 is done or given up on (in case that idea isn't feasible for some reason, it's still not really tested). Otherwise this would need to be refactored again afterwards. If you agree to do the refactoring so the new functions are used here once https://github.com/tvheadend/tvheadend/pull/1637 is merged, I'd agree to merge this once the complexity is reduced a little. There is no ETA for https://github.com/tvheadend/tvheadend/pull/1637.

@Flole998 Given the recent discussion in #1637 and the still WIP nature of it, I'd propose to merge this PR as is - if okay from a code quality point of view of course. 

Regarding your ask - I agree to do the refactoring so the new functions are used - I also ported these changes already on top of #1637 for future reference: https://github.com/tvheadend/tvheadend/compare/master...chkuendig:tvheadend:ssdp-remove8888 (which is what I'm currently running on locally)

chkuendig:
@Flole998 Is there anything left to be done before this one can be merged?

Flole998:
Not sure, I haven't found time to review this as I thought it would be redone after #1637, I'll see if I find some time over christmas to have a look at it.

PR #1681, opened 2024-04-07
Let tvheadend recognize avs video format

yzhj:
Includes avs, avs2 and avs3, usually used for DTV and IPTV in China.
The code just let tvheadend know that is a video stream.


Flole998:
Unfortunately this is incomplete, it is necessary to include it in the matroska muxer so it can get copied into mkv-streams here: https://github.com/tvheadend/tvheadend/blob/4874aaa3161fbdd8b9d3abe50fd3fa20b18f8b0b/src/muxer/muxer_mkv.c#L253

PR #1685, opened 2024-04-11
Add subsystems to JSON API.

DeltaMikeCharlie:
Add the ability to extract the available subsystems and their descriptions via the JSON API.

https://tvheadend.org/d/8597-extracting-tvh-subsystems-via-the-json-api

Sample:
```
        {
            "id": 12,
            "subsystem": "fsmonitor",
            "description": "Filesystem monitor",
            "trace": false,
            "debug": false
        },
```
**id** - The internal ENUM for this subsystem.
**subsystem** - The name of the subsystem.
**description** - The locale-aware description of the subsystem.
**trace** - True if this subsystem is currently being traced.
**debug** - True if this subsystem is currently being debugged.

PR #1686, opened 2024-04-17
Update WebUI to allow debug/trace subsystem selection from a list.

DeltaMikeCharlie:
Update the WebUI to utilise the API subsystem list introduced in PR https://github.com/tvheadend/tvheadend/pull/1685

Also see [forum discussion thread](https://tvheadend.org/d/8597-extracting-tvh-subsystems-via-the-json-api) on the matter.


Flole998:
Try to avoid using `var`, use `let` or `const`, especially since some . That is also what sonarqube flagged.

DeltaMikeCharlie:
> Try to avoid using `var`, use `let` or `const`, especially since some of the variables are re-declared in a different scope. That is also what sonarqube flagged.

Done.

PR #1693, opened 2024-04-26
Fix echo target for superuser file in Debian postinst

mamarley:
aba5e60792177d6a2a867445559f4806973b3258 was causing the username and password to get printed to the console instead of being put in the correct file.  Also, use the modern $() syntax instead of `` and quote all variable assignments.


Flole998:
Thanks, that was still the test code commited.... We REALLY need a good testing environment so that code can actually be tested before merging....

PR #1705, opened 2024-06-02
Allow setting a custom grace period for LinuxDVB adapters

xavery:
When using Astrometa to tune to DVB-T2 muxes in Poland, the scans are reported as complete but the found services have zero elementary services due to the scan period being too short in order to fetch PMTs.

This change allows overriding the default grace period of 5 seconds with a custom value. I successfully scanned all services with this setting changed to 15 for this particular adapter/mux combination.


Flole998:
Please revert the merge commit, it can't be merged like this.

Flole998:
Now you have 3 commits, it should be a single commit with your changes only, no merge commit, no undoing of other commits. You probably want to hard-reset to HEAD~2 and force-push to fix this.

Flole998:
We do rebasing instead, which can't be combined with squashing on merge.

PR #1707, opened 2024-06-03
Docker/Alpine: Remove USB group

oliv3r:
The USB group has been removed from upstream alpine in commit bb00d0e4f345 ("main/alpine-baselayout: remove mem and usb group") which was a fixup on commit
f16d0754d601 ("main/alpine-baselayout: remove unused/moved users and groups")

Lets remove it here as well as we cannot join the group any longer.

Besides, device access is probably better managed with host specific udev rules.


Flole998:
This is (also) part of #1696. I will remove it from the other PR as this one has a better commit message.

The failing ubuntu builds are caused by a staged rollout which causes the dependencies to not be available.... 10% is the current status on that one (https://launchpad.net/ubuntu/xenial/amd64/software-properties-common/0.96.20.13), once it's rolled out we will get all green and it will be automatically merged.

PR #1720, opened 2024-06-23
Replace some deprecated code

a-andre:


Flole998:
Wow, thanks a lot!

a-andre:
What's the minimal version of ffmpeg tvheadend needs to work with?

Issue #1714, opened 2024-06-17
[BUG]: Event descripton of channels in EPG missing


Haxe18:
### Tvheadend Version

v4.3 (development)

### Build Number

4.3-2336~g366e56290

### Platform

Debian 12 (Bookworm)

### Architecture

ARMv6 (arm)

### Category

EPG

### Description

On some Vodafone West BW muxes (all muxes with Network-ID 133 - 362 MHz, 378 MHz, 386 MHz, 522 MHz, 530 MHz, 554 MHz) on all events of the channel the description is missing, but only if another mux is scanned after.

If the mux only is grabbed, the description to all events are present as expected.

This bug is only about the detailed description, the event name and subtile is grabbed as expected.
On all other muxes the description of the events is always fetched as expected.

On my enigma2 Dreambox via FritzBox Sat>IP the description of the event is set for all channels, so this seems to be an TVH Bug.

### Logfile

For further Screenshots see: https://tvheadend.org/d/8670-problems-with-epg-description

### Reproduction

The bug is reproducible for me via FritzBox Sat>IP and also via Sundtek Tuner, also in different rooms/Cable connections on an empty tvheadend Instanace.

Bug occures also on x86 or armv6.

Flole998:
This issue report is missing debug logs.

Haxe18:
I hope logs of the levels _tbl-eit,epg,epgdb,epggrab_ are enough:

[386MHz_only.log](https://github.com/user-attachments/files/15877634/386MHz_only.log)
[386MHz+378MHz.log](https://github.com/user-attachments/files/15877644/386MHz%2B378MHz.log)

As you see in 386MHz+378MHz.log for Sky Sport Golf HD:
```
2024-06-17T23:44:09.687630+02:00 raspberrypi4 tvheadend[3027848]: tbl-eit: 00 90 CD 00 01 00 03 00 85 01 4F 8C A3 EC 3E 21 ..........O...>!
2024-06-17T23:44:09.687720+02:00 raspberrypi4 tvheadend[3027848]: tbl-eit: 30 00 00 30 00 91 43 4D 59 44 45 55 23 05 47 6F 0..0..CMYDEU#.Go
2024-06-17T23:44:09.687791+02:00 raspberrypi4 tvheadend[3027848]: tbl-eit: 6C 66 3A 20 45 44 47 41 20 2D 20 4C 69 76 69 6E lf: EDGA - Livin
[...]

PR #1728, opened 2024-07-04
Keep event description on epg entry update

Haxe18:
Fixes https://github.com/tvheadend/tvheadend/issues/1714

Added fallback to reuse the description of the already existing epg event details.

### DIFF
--- a/src/epggrab/module/eit.c
+++ b/src/epggrab/module/eit.c
@@ -749,8 +749,11 @@ static int _eit_process_event_one
     if (short_target != 0 ||
         (ev->subtitle && lang_str_compare(ev->summary, ev->subtitle)))
       *save |= epg_broadcast_set_summary(ebc, ev->summary, &changes);
-  if (ev->desc)
+  if (ev->desc) {
     *save |= epg_broadcast_set_description(ebc, ev->desc, &changes);
+  } else if (ebc->description) {
+    *save |= epg_broadcast_set_description(ebc, ebc->description, &changes);
+  }
 
   /* Broadcast Metadata */
   *save |= epg_broadcast_set_is_hd(ebc, ev->hd, &changes);


DeltaMikeCharlie:
@Haxe18 - Some random thoughts....

If the broadcaster truly intends to erase the description, this proposed change will defeat that intention.

If a channel has multiple EPG sources, some with and some without descriptions, those grabber sources with descriptions could be prioritised instead.  As mentioned in the issue that you raised a few weeks ago, this is not really a bug.

If this really needs to be implemented, perhaps it should be parametrised and not applied universally.  Enable at either the global, mux, service or channel level.  Disabled by default.

Perhaps you could consider testing to see if the title/sub-title descriptors for the event are still the same too.  If the title/sub-title have changed and the description is null, then allow a null description.  Perhaps a broadcaster could reuse the same EventID, but change the actually programme being broadcast due to a last-minute change.

Flole998:
This should really be fixed at the provider level.... If a workaround in Tvheadend is implemented it needs to look different, as otherwise the next issue might be just around the corner: What about all the other data the provider transmits on the wrong frequency? I would agree with an option on the network level to enforce that the EIT must be transmitted on the same frequency as the mux it contains data for, which is off by default. Something like "Ignore EIT on other muxes". I don't think this needs to be taken to a mux-level, if a provider messes it up in such a way that some EIT entries are transmitted on a different frequency while other, bogus, EIT entries are also transmitted on a different frequency with the actual EIT being transmitted on the same frequency, I will personally reach out to them and ask them if they have any idea what they are doing there.....

Such a feature would be especially important if you live in a house with a DVB-C headend which injects channels into the network, for example from a satellite source. If the service IDs collide there, you'd be seeing this issue I guess and this option could be turned on as a fix.

DeltaMikeCharlie:
Perhaps the 'EPG Grabber Channel' concept could be expanded.

image

The `module` and `path` properties could be set to be some combination of `EIT` and the mux/service (ONID/TSID/SID).  Multiple EPG sources from different muxes/frequencies would result in multiple grabber channels.  At the channel level, the users selects the required EPG source.

image

The 'path' could look something like: `DVBT/219.5/4116/1538/1569` (Protocol/Mux-Frequency/ONID/TSID/SID).  If there was another path like `DVBT/599.5/4116/1538/1569` that had an empty extended description, the user could select the first one on that channel.

This could also be used to resolve the New Zealand EPG issue for areas outside of Auckland where the EPG DVB triplets don't match the transmission triplets.

Also, multiple `services` can supply a single `channel`.  If a user gets a channel from both DVB-T and DVB-S, they will need to decide which EPG source to prefer for that channel.  For single service channels, the setup wizard could just 1:1.  For multi-service channels, the service with the higher priority could be used.

Flole998:
It shouldn't be "protocol" but "network" (multiple satellites carrying the same channel) other than that it's a great idea.

DeltaMikeCharlie:
> It shouldn't be "protocol" but "network"

Yes, this concept would be subject to much fine-tuning such as this.

For example, behind the scenes, the `path` would actually be a chain of `UUIDs` that pointed to the actual objects in question.  The text that is displayed would be to aid the user in making their selection.

DeltaMikeCharlie:
1a) Looking at the `epggrab_channel` structure, perhaps a new `channel_signature` property of type `htsmsg_t` could be added to store the Network/Frequency/ONID/TSID/SID/etc parameters that would be required to uniquely match EPG data from a grabber to a channel.

https://github.com/tvheadend/tvheadend/blob/d2e41b553e7cc6eb06fd21b42bbed4b3a1f28bc0/src/epggrab.h#L81C1-L107C21

https://github.com/tvheadend/tvheadend/blob/d2e41b553e7cc6eb06fd21b42bbed4b3a1f28bc0/src/epggrab/channel.c#L741C1-L892C3

The reason that I am suggesting a `htsmsg_t` property as opposed to multiple fixed elementary properties (int, char, etc) is to provide flexibility in the matching process as not all channels (maybe IPTV channels) may have/require the exact parameters listed above.

Provided that the `mod` property matches the EPG source grabber, the variable contents of the `channel_signature ` property could then be used by the source module to map the EPG data to the destination channel.

1b) Instead of using a single `htsmsg_t`, perhaps 2 x fixed 128 bit signatures could be used.  One signature for the grabber side and 1 signature for the channel side.  This would be less flexible, but would allow for quicker matching by not having to encode/decode a `htsmsg_t` message for every EPG event.  The channel side signature could be the ONID/TSID/SID values bit-shifted to fit and the grabber side would be the bit-shifted network UUID plus frequency or IP address or whatever.

2a) Another possible solution that I considered and discarded was to change the EPG database processing so that ALL EPG data that is obtained from ALL sources is stored.  The `channel grabber` concept would connect the `channel` to the EPG Database and not the EPG grabber directly.  In this way, the user could browse the contents of various EPG sources for a particular channel before deciding on the best one for their requirements.

This would be a big change to the way that the EPG is stored and could potentially GREATLY increase the RAM and HDD required to store the EPG data.  It would also only be of short-term benefit during the EPG source selection process.  Once an EPG source had been selected, the unselected EPG data sources would never be used and waste RAM/HDD.

3a) Perhaps if a huge change is palatable, the EPG DB could be changed to use something like sqlite3.  ALL EPG data, regardless of source, would be written directly into the DB, with only EPG sources actually assigned to channels being cached in RAM.  This would be problematic to implement under the current architecture because the EPG is held in RAM and the EPG on disk is basically a serialised version of the RAM EPG storage structure using htsmsg.  Because data from all EPG sources are stored, the user could change between various live EPG sources.

Haxe18:
> 
> If the broadcaster truly intends to erase the description, this proposed change will defeat that intention.
> 
As mentioned in the issue, on my Dreambox and i also guess on TVs and Vodafone/Sky Provider Receivers the epg description always present and not empty !
So i bet they don't want to erase the description deliberately  ;)

>
> If a channel has multiple EPG sources, some with and some without descriptions, those grabber sources with descriptions could be prioritised instead. 
>As mentioned in the issue that you raised a few weeks ago, this is not really a bug.
>
In my opinion the bug is in TVH blindly overwriting infomations. For me these informations should be compared an merged.
There is already a logic recognizing same epg events, why not comapre and reuse data from those equal events as i partly done ?

>
> This should really be fixed at the provider level.... 
>
I agree, but I am 99% percent sure Vodafone tell me, with their Hardware there is no problem, please use our hardware

>
> What about all the other data the provider transmits on the wrong frequency?
>
I've not noticed wrong data in epg yet, only the missing description bothers me.

>
>I will personally reach out to them and ask them if they have any idea what they are doing there.....
>
I can't reach those persons as they are in 3rd Level and normally you only reach 1st Level via Hotine/Chat etc.

---

I think this "workaround" is not negative affecting the current epg handling in any way.

Think from what i've understand out of the code the new part is located in the "update existing epg event" section, so from the tvh side the epg events are the same.

TVH checks epg data, find out this already exists and is preparing to overwrite the existing data.

If the new epg event:
 - Has a description -> Handled like before
 - Has no description:
   - Old event has no description -> Handled like before
   - Old event has a description -> Reuse it in the updated event
  
I agree with you reworking the epg scan code with the frequency limit is the cleaner and proper way to avoid this and other possible interferences.
But as a first step this is an okay fix for missing descriptions. 

I am not an expert in c coding so creating this feature is unfortunately beyond my capabilities but i'm looking forward to get the pull request merged.

Flole998:
> In my opinion the bug is in TVH blindly overwriting infomations. For me these informations should be compared an merged.

Those information shouldn't exist in the first place. Tvheadend receives perfectly valid information and then uses it. That is not a bug. If the data is not empty for some reason you'd have the same issue again. This is not a proper fix, in fact this will cause weird behavior that eventually later on someone will find and revert again. So like this it can't be merged.

DeltaMikeCharlie:
@Haxe18 - Have you tried disabling the EIT on the service that has an empty description?

image

Haxe18:
@DeltaMikeCharlie 
Also with this setting the description is missing after a scan.


---

@Flole998 
> 
> Those information shouldn't exist in the first place. 
> Tvheadend receives perfectly valid information and then uses it. 
>
Welcome to real world where not everything is perfect and will never be.

>
>That is not a bug. If the data is not empty for some reason you'd have the same issue again.
>
Different problem and not my current. This fix is not addressing this case.

>
> This is not a proper fix, in fact this will cause weird behavior that eventually later on someone will find and revert again.
>
I don't see a case in which this change will cause weird behavior.
But i don't know the whole code. Only from my little point of view:
Description empty keep old, if not empty same logic as in the past.


>
>  So like this it can't be merged.
>
It seems no one else except me having this issue, therefore i accept this decision and I need to build my own releases in the future. 

---

The fix is working perfect for my case.
Full description on all channels and all events.

Thanks to both of you so far for your time effort and input in here.

I would appreciate the feature to be able to limit epg data to the current frequency.
I can't do it myself, maybe someone else is able to do it, so i need to wait for it.

Feel free to close the PR at this point.

Haxe18:
@DeltaMikeCharlie 
If I disable mux scanning some channels have no epg anymore.
Not an option.

Because all channels are showing up only once in "services" overview (another bug - There are "invisible" channels where TVH scanns epg from ?), it's not possible to disable the broken epg source in any way for me.

PR #1736, opened 2024-08-01
Add start timeout to streaming profile

lbschenkel:
This allows overriding the hardcoded grace period of 20 seconds.
It should address the problems described in [1][2].

In addition, timeout code has been slightly refactored for readability and more debug logging.

[1] https://tvheadend.org/d/8330-increase-timeout-when-tuning-iptv-mux/2
[2] https://tvheadend.org/d/8158-several-problems-questions-about-using-tvheadend-starting-with-not-waiting-long-enoough-for-stream-to-begin

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -340,7 +340,7 @@ http_stream_run(http_connection_t *hc, profile_chain_t *prch,
   int run = 1, started = 0;
   streaming_queue_t *sq = &prch->prch_sq;
   muxer_t *mux = prch->prch_muxer;
-  int ptimeout, grace = 20, r;
+  int ptimeout = 5, ptimeout_start = 0, grace = 20, r;
   struct timeval tp;
   streaming_start_t *ss_copy;
   int64_t lastpkt, mono;
@@ -356,7 +356,23 @@ http_stream_run(http_connection_t *hc, profile_chain_t *prch,
     socket_set_dscp(hc->hc_fd, config.dscp, NULL, 0);
 
   lastpkt = mclk();
-  ptimeout = prch->prch_pro ? prch->prch_pro->pro_timeout : 5;
+
+  if (prch->prch_pro) {
+    struct profile *pro = prch->prch_pro;
+
+    ptimeout = pro->pro_timeout;
+    ptimeout_start = pro->pro_timeout_start;
+    tvhdebug(LS_WEBUI, "Using timeouts from profile: %s", pro->pro_name);
+    tvhdebug(LS_WEBUI, "Packet timeout (from profile): %d secs", ptimeout);
+  } else {
+    tvhdebug(LS_WEBUI, "Packet timeout (default): %d secs", ptimeout);
+  }
+  if (ptimeout_start > 0) {
+    grace = ptimeout_start;
+    tvhdebug(LS_WEBUI, "Grace period (from profile): %d secs", grace);
+  } else {
+    tvhdebug(LS_WEBUI, "Grace period (default): %d secs", grace);
+  }
 
   if (hc->hc_no_output) {
     tvh_mutex_lock(&sq->sq_mutex);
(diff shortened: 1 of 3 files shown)


src/webui/webui.c
```diff
@@ -410,11 +428,19 @@ http_stream_run(http_connection_t *hc, profile_chain_t *prch,
       break;
 
     case SMT_GRACE:
-      grace = sm->sm_code < 5 ? 5 : grace;
+      if (sm->sm_code < 5) {
```
lbschenkel, line 431: In my testing `SMT_GRACE` is called immediately when starting the stream. The old version of the code would instantly downgrade the grace period from 20 seconds to 5.

I believe this was not intentional and it is a bug. I've changed it so it can only increase the period, but not decrease it.

src/webui/webui.c
```diff
@@ -410,11 +428,17 @@ http_stream_run(http_connection_t *hc, profile_chain_t *prch,
       break;
 
     case SMT_GRACE:
-      grace = sm->sm_code < 5 ? 5 : grace;
+      if (sm->sm_code > 5) {
```
lbschenkel, line 431: I'm not quite sure what is the intent of the old logic, so I have preserved it here.

In my testing, playing every stream results in a `SMT_GRACE` with a code of 3 on startup, which ends up being ignored.

src/webui/webui.c
```diff
@@ -410,11 +428,17 @@ http_stream_run(http_connection_t *hc, profile_chain_t *prch,
       break;
 
     case SMT_GRACE:
-      grace = sm->sm_code < 5 ? 5 : grace;
+      if (sm->sm_code > grace) {
```
lbschenkel, line 431: I don't quite understand the old logic. Grace period could be changed to 5, or left as it was.

In my system, every new stream ends up with `SM_GRACE` with a code of 3 being invoked. The value is upgraded to 5, and that downgrades the grace period of 20 to 5. So the actual grace period is always 5 seconds and not the hardcoded period of 20.

I am not sure that was intentional. I believe it was a bug. I have changed the behaviour so the grace period can only be increased, and to a minimum of 5 seconds. That is the best assumption I could make of the original intention and it was needed to make the code respect the timeouts.

This can be reworked if I misunderstood how it is supposed to work.

DeltaMikeCharlie:
> Btw: this is my first contribution to this project. I hope I followed all the guidelines.

@lbschenkel - Welcome.

I don't know this area of code at all, but I have some random thoughts:

What can this new setting do that can not be achieved with the existing 'timeout' property?
Is there any duplication of functionality or potential for conflicting behaviour or misunderstanding between the 2 settings?

Should this setting be for 'Expert' level?  The other data timeout is not.

Do you also plan to update the screen shot from the help text?

https://github.com/tvheadend/tvheadend/blob/master/src/webui/static/img/doc/stream/stream_profiles_tab.png

lbschenkel:
> What can this new setting do that can not be achieved with the existing 'timeout' property? Is there any duplication of functionality or potential for conflicting behaviour or misunderstanding between the 2 settings?

So, let me illustrate with a scenario. There is one channel with 8 IPTV sources (1 source = 1 mux = 1 service). Imagine that they are configured so they fail over after 15 seconds without data. The intent is to fail over so all sources can be tried before failing/closing the subscription.

In the current code, it is impossible to that no matter how you configure it, because:
1. the subscription will fail after the time of 20 seconds which is hardcoded — that is not enough to try over more than 1 or 2 services
2. even worse, the code as it behaves today lowers that 20 second timeout to only 5 seconds (see my comment above)

I _could_ have made the grace period equal to the profile timeout, however in the scenario above you would need a 2 minute timeout to be able to try all the sources before giving up. That has the unfortunate side effect of also having to wait for 2 minutes when there's a network stall. Separating the two allows having a sane "regular" timeout to fail the current service, let's say 15-30 seconds, and then a larger "initial" timeout (2 minutes) so you have time to cycle through all alternatives. That is why I created a new setting.

I hope that makes sense. You can also check https://tvheadend.org/d/8330-increase-timeout-when-tuning-iptv-mux/2 for another scenario.

Now that I wrote that, I'm inclined to think that "initial timeout" might be a better name that "start timeout". What do you think?

> Should this setting be for 'Expert' level? The other data timeout is not.

Not sure, to be honest. Since most settings are at expert level, and this setting might be a bit niche, I thought that making it expert was the conservative choice so we don't show it unnecessarily. But I'm OK with either option.

> Do you also plan to update the screen shot from the help text?
> 
> https://github.com/tvheadend/tvheadend/blob/master/src/webui/static/img/doc/stream/stream_profiles_tab.png

I missed it, but I will of course update it — but I'll leave it for last once we're happy with the other details, as we might still rename/change things.

DeltaMikeCharlie:
> What do you think?

I don't anything about this area of the code, but your explanation sounds superficially coherent.  Perhaps you could also add an abridged version to the online help documentation.

https://github.com/tvheadend/tvheadend/blob/master/docs/class/profile.md

src/webui/webui.c
Flole998: I will see if I can make any sense of the old code or understand why it was done this way

PR #1743, opened 2024-08-16
fix bug in AAC channel layout configuration tab

uknunknown:
There are few issues:
1. first entry in combo should be AUTO (with value 0) - in original code was set to 1 (and overwritten later)
2. l->nb_channel is not the best way to cycle though layouts available. At the end I think is accessing some region outside of the struct (because I see is lopped also after 7.1). The way I knew how to fix was to add the filter (l->nb_channels < 32). Maybe changing the while to for will be a better option.
3. av_channel_layout() is returning the length of the string ... we should use l_buf only when retuned value > 0 ... when is < 0 l_buf was not updated.


uknunknown:
I think the regression was introduced by: https://github.com/tvheadend/tvheadend/commit/078a822cf548b37bc474475fa57e48e9604090ee

PR #1744, opened 2024-08-22
HTSP: Expose service provider name with channel information.

ksooo:
For upcoming Kodi v22 support for "PVR Providers" I need the service provider for a channel exposed via HTSP. Thanks.

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -51,7 +51,7 @@
 
 static void *htsp_server, *htsp_server_2;
 
-#define HTSP_PROTO_VERSION 37
+#define HTSP_PROTO_VERSION 38
 
 #define HTSP_ASYNC_OFF  0x00
 #define HTSP_ASYNC_ON   0x01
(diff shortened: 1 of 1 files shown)


Flole998:
Sure, can you also add the change to the docs? Either directly in the markdown files in the documentation repository or using the (easier to use) GitBook editor at https://app.gitbook.com/invite/GUiJM9P5N3CkY42Hcs8P/7hmZSVRZkZqEIJarLFX5 (it will open a PR in the background when you submit the changes)

src/htsp_server.c
```diff
@@ -906,6 +906,10 @@ htsp_build_channel(channel_t *ch, const char *method, htsp_connection_t *htsp)
     if (t->s_hbbtv)
       htsmsg_add_msg(svcmsg, "hbbtv", htsmsg_copy(t->s_hbbtv));
 
+    /* Provider */
+    if (t->s_provider_name)
+      htsmsg_add_str(svcmsg, "providerName", t->s_provider_name);
```
Flole998, line 911: You can use `htsmsg_add_str2` here, which already includes the null check (great naming of that function, I know, that is also on the todo list)

ksooo: Thanks. Will do.

PR #1745, opened 2024-08-23
Fix FTBFS introduced by 76d8fc8bc5455322558c764c84755ebbba254ad5

mamarley:
Older versions of GCC don't like declaring a variable in the middle of a switch/case and will fail with "error: a label can only be part of a statement and a declaration is not a statement".


Flole998:
Just out of curiosity: How old? We're running some pretty old stuff in our CI already, how much older do we have to go?

mamarley:
> Just out of curiosity: How old? We're running some pretty old stuff in our CI already, how much older do we have to go?

My PPA builds fail on Ubuntu 20.04 (which uses GCC 9) and older.

Flole998:
I guess we aren't including this code then in our builds.

Issue #1738, opened 2024-08-04
[FEATURE]: Allow restricting UDP ports for IPTV reception (notably for RTSP)


Arno500:
### Description

I have a Freebox at home, in France, that allows to recast IPTV streams for all the channels that I have access to. This is great, but there is one issue: the streams are done via RTSP and only via UDP unicast. This means that if I want to containerize TVHeadend (which I do), I would either need to bind the whole container to the host (which I do not want to do, and it would conflict with other ports notably if I intend to use the HDHomerun emulator) or bind all UDP ports to the container (which is also a big no no) AND open all the ports in the software firewall that I have in place.

The other alternative would be to actually control the `client ports` that are sent through the initial SETUP request to only allow a specific range of ports to pass through. In the options, there would be an option for a minimum and a maximum port, with a pool available (since each connection takes a minimum of two ports) and as a user, you would be in charge to ensure there are enough ports for the number of input streams allowed in your various sources at all time. This is something that FFMPEG provides on their side and that can actually be really useful.

I already drafted a fix, this should be pretty trivial, but my time is limited for now, and I will be on holidays soon.

### Help and Guidance

I do not think I need particular help and guidance, but I simply want to ensure that I am going in the proper direction with the plan:
1. Add `Minimum UDP port` and `Maximum UDP port` in the options
2. When opening a stream, if the options above are set, take the lowest available port (+1 for the control channel) and lock them somewhere so they cannot be used until they are freed (or use the current logic of taking the current connection port as a base). I also need to make sure the two available ports are contiguous due to the protocol, but if my code is good, it should not be a problem.
3. If there are no more ports available, throw an error and clearly indicate to the user what is wrong

Would that fit the project as intended?

### Acceptance

- [X] The proposal will be developed and maintained by me

Flole998:
> and lock them somewhere so they cannot be used until they are freed

This is the difficult part, as a port may be in use by another application. I have done something similar in a different project and IIRC you basically need to try to bind to 2 consecutive ports within the range, if it fails on the first one, move up 1 port, if it fails on the second one, move up 2 ports and try again, until it works out.

Arno500:
Just pushed the first version of the feature here: #1751

alo0:
Hello @Arno500 ,

First of all, thank you for your contribution.
I am in the same situation as you: would like to avoid host network for my TVH docker container and using Free ISP. 
Could you please explain me where your port settings are available in TVH ?

[...]

PR #1751, opened 2024-08-30
iptv: allow to limit UDP ports for unicast inputs

Arno500:
Here is the associated feature request with explanations: #1738

### DIFF
--- a/src/input/mpegts/iptv/iptv_rtsp.c
+++ b/src/input/mpegts/iptv/iptv_rtsp.c
@@ -18,6 +18,7 @@
  */
 
 #include "tvheadend.h"
+#include "config.h"
 #include "iptv_private.h"
 #include "iptv_rtcp.h"
 #include "http.h"
@@ -263,6 +264,8 @@ iptv_rtsp_start
   http_client_t *hc;
   udp_connection_t *rtp, *rtcp;
   int r;
+  int rtp_port;
+  int max_rtp_port;
 
   if (!(hc = http_client_connect(im, RTSP_VERSION_1_0, u->scheme,
                                  u->host, u->port, NULL)))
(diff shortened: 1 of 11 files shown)


src/udp.c
```diff
@@ -171,6 +171,10 @@ udp_bind ( int subsystem, const char *name,
   uc->rxsize               = rxsize;
   uc->txsize               = txsize;
 
+  if (bindaddr) {
+    uc->multicast = strncmp("239.", bindaddr, 4) == 0;
```
Flole998, line 175: This is not ipv6 ready

Arno500: Indeed and the worst part is that actually the multicast bit is set in the bind function just below that via reference, so no need to even bother with this, I removed it.

src/input/mpegts/iptv/iptv_rtsp.c
```diff
@@ -272,13 +274,23 @@ iptv_rtsp_start
...
     hc->hc_rtsp_pass = strdup(u->pass);
-
-  if (udp_bind_double(&rtp, &rtcp,
-                      LS_IPTV, "rtp", "rtcp",
-                      NULL, 0, NULL,
-                      128*1024, 16384, 4*1024, 4*1024) < 0) {
-    http_client_close(hc);
-    return SM_CODE_TUNING_FAILED;
+  
+  rtp_port = config.rtsp_udp_min_port;
```
Flole998, line 278: I believe this code will not fail if someone configures max_port smaller than min_port. It will still attempt to bind while it should just error out. Also a max_port of 0 should probably mean "unlimited"/unconfigured.

Arno500: Fixed! I made a specific check that will fail in this case.
I also make sure to set the starting port to 0 (aka random) if the max_port is at 0 (unconfigured).
So:
- setting a min port only will crash (min > max)
- setting a max port only will make the port random (starting port = 0)
- setting both will limit them as expected
- setting none (both at 0) will use a random port and not try anything else if the first binding fail, like before

PR #1753, opened 2024-09-02
HTSP Updates for UUID and Rating Labels

DeltaMikeCharlie:
1) UUID
For existing message types that use a u32 ID field, for example `channelAdd`, `tagAdd` and `dvrEntryAdd`, add full UUID as a 32 character string for interoperability with objects returned via the HTSP JSON proxy APIs.

2) Rating Labels
a) Add the country and authority values to HTSP messages carrying parental rating labels.
b) Add the saved country and authority values to completed recordings.
This is so that clients (like Kodi) can use the rating properties held by TVH to interpret their own rating labels if required.

3) HTSP version increased to 41.

### DIFF
--- a/src/dvr/dvr.h
+++ b/src/dvr/dvr.h
@@ -249,8 +249,10 @@ typedef struct dvr_entry {
   //may have changed, so keep an absolute copy of the values at
   //the time of recording rather than pointing to a rating label
   //object that may no longer exist many years later.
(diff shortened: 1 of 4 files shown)


Flole998:
Please add this change to the docs

DeltaMikeCharlie:
> Please add this change to the docs

Naturally, I was only waiting for it to be merged before I updated the docs.

DeltaMikeCharlie:
> Please add this change to the docs

The documentation has been updated.

PR #1754, opened 2024-09-06
HTSP: deliver 'comment' with autorecEntry(Add|Update), timerecEntry(A…

ksooo:
…dd|Update). Allow setting 'comment' with 'updateDvrEntry'.

Bump HTSP version to 41.

I need this for extended dvr property support in upcoming Kodi / pvr.hts release.

PR #1755, opened 2024-09-06
Fixup updating comment in _dvr_entry_update. Only overwrite existing …

ksooo:
…comment if new comment is not NULL. Follows the same logic now as other updates done in this function.

This small functional regression sneaked in with https://github.com/tvheadend/tvheadend/pull/1754.

### DIFF
--- a/src/dvr/dvr_db.c
+++ b/src/dvr/dvr_db.c
@@ -2569,9 +2569,9 @@ static dvr_entry_t *_dvr_entry_update
   }
 
   /* Comment */
-  if (strcmp(de->de_comment ?: "", comment ?: "")) {
+  if (comment && strcmp(de->de_comment ?: "", comment)) {
     free(de->de_comment);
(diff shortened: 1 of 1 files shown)


ksooo:
@Flole998 the 2 new Sonarcloud issues are nothing I have introduced with this PR. They are both related to the latest PR from @DeltaMikeCharlie

Flole998:
Yup, saw it. In fact they aren't even part of that PR, they must be even older.... @DeltaMikeCharlie Could you check if there is anything that should be done with that parameter? If it's unused and can be removed, that should be done.

ksooo:
@Flole998 speaks something against merging this PR? The Sonarcloud failure is unrelated, as we already found out.

Flole998:
I was just waiting for all checks to run through, automerge does not work if Sonarcloud is failing, so I had to wait and do it manually.

DeltaMikeCharlie:
> If it's unused and can be removed, that should be done.

I will look into it.

PR #1760, opened 2024-09-20
Remove HTSP client version tests for rating labels and string UUIDs

DeltaMikeCharlie:
Currently, all HTSP client versions only receive rating label fields if rating labels are enabled within TVH.  Older clients are also 'protected' from receiving rating label fields using the client preferred protocol version number exchanged during the session handshake.

This is not in keeping with the original intent of the protocol version numbering scheme whereby only changes to existing fields needed to be controlled via versioning, not the presence of new fields.

This PR removes all version-dependent tests relating to rating labels from HTSP and provides all rating label fields regardless of the client version.  It should be noted that if rating label processing is not enabled in TVH, no rating label fields will be provided via HTSP regardless of the client version.

Following the same logic, the client version tests governing the supply of UUIDs as string fields have also been removed.

These changes have been tested with Kodi v20 and pre-release Kodi v22-Alpha1 as clients.

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -869,11 +869,8 @@ htsp_build_channel(channel_t *ch, const char *method, htsp_connection_t *htsp)
   htsmsg_t *services = htsmsg_create_list();
 
   htsmsg_add_u32(out, "channelId", channel_get_id(ch));
-
-  if (htsp->htsp_version > 40){
-    htsmsg_add_str(out, "channelIdStr", idnode_uuid_as_str(&ch->ch_id, buf));
-  }
-
+  htsmsg_add_str(out, "channelIdStr", idnode_uuid_as_str(&ch->ch_id, buf));
+  
   htsmsg_add_u32(out, "channelNumber", channel_get_major(chnum));
   if (channel_get_minor(chnum))
     htsmsg_add_u32(out, "channelNumberMinor", channel_get_minor(chnum));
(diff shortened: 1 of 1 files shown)


ksooo:
> This is not in keeping with the original intent of the protocol version numbering scheme whereby only changes to existing fields needed to be controlled via versioning, not the presence of new fields.

Exactly.

src/htsp_server.c
```diff
@@ -1019,82 +1010,66 @@ htsp_build_dvrentry(htsp_connection_t *htsp, dvr_entry_t *de, const char *method
...
-    }
+          //If the rating icon is not null.
+          if(de->de_rating_label->rl_icon){
+            str = de->de_rating_label->rl_icon;
+            if (!strempty(str)) {
+              str = imagecache_get_propstr(str, buf, sizeof(buf));
+              if (str)
+                htsmsg_add_str(out, "ratingIcon", str);
+            }//END got an imagecache location
+          }//END icon not null
```
ksooo, line 1033: As a general side note: If the respective code blocks are short, imo these comments do not add any value as you always have the related `if` /`else`on the screen. But you always risk that the comments get outdated whenever you change the content of the code block. I suggest to just remove these comments.

OTOH; if the code blocks are so long that you do not see the conditions on your screen when looking at the end of the block, then this is a clear indicator that the block is to long. You should factor out (parts of) the code in the block to separate helper functions with speaking names. So you get readability of the code without the price for the comments explaining which block end we're looking at.

ksooo:
Hmm, seems that my comments regarding readable/maintainable code just got ignored. I will refrain from commenting such things from now on.

Flole998:
@ksooo Oops sorry, I must have somehow missed it.

I'm not really sure how we should approach this, there are many different coding styles already used and in general it should be a single-ish coding style. So there has to be some consensus on how to add such comments, and that needs to be enforced. I'd propose the following: Do not add any comments to the end of blocks but add comments to the beginning of the blocks. That way most IDEs can use some plugin to show those comments at the end, for the whole codebase, or if no comment exists show the condition again. I'm sure such plugins do exist. It would effectively avoid the need to maintain or enforce such comments, and only those who actually want to see them will see them.

ksooo:
My approach would be to avoid this comments completely by shortening long code blocks, like i wrote in my review comment.

PR #1765, opened 2024-09-29
Update remote timeshift description, RTSP only

spdfrk:


spdfrk:
It makes sense to do it like that. From what I remember it was easier to do it like this to have the info for enabling remote ts at the right time. If you think it doesn't take much effort feel free to change it. I can take a look but it will take some time.

Flole998:
I'm not sure how much effort it takes, but the current approach is wrong and causes issues. If you could have a look at how to implement it on the service level that would be great. I can prepare a PR to remove it from the channel level if you want, that would also solve the bug where it causes crashes.

Flole998:
I am going to prepare a PR to remove this from the channel level again as this is not the right place and causes issues. This has been sitting here for quite some time without the real, underlying issue being addressed.

Issue #1764, opened 2024-09-28
[BUG]: Audio transcoding not working with FFMPEG 7


dkadioglu:
### Tvheadend Version

v4.3 (development)

### Build Number

4.3.r2368~g554004da

### Platform

Other..

### Architecture

ARMv8 (aarch64)

### Category

Transcoding

### Description

As described in https://github.com/tvheadend/tvheadend/pull/1721 building TVHeadend with FFMPEG 7, I have no more audio except with the profile pass. I hope it is ok to open a new issue for that.
Whenever transcoding is involved, I see this in the log:

```
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 01:H264: ==> Copy
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 02:MPEG2AUDIO:
==> Using profile webtv-aac
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 03:MPEG2AUDIO:
==> Filtered out
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 05:DVBSUB: ==>
Filtered out
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 06:AC3: ==> Filtered out
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 08:MPEG2AUDIO:
==> Filtered out
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 09:TEXTSUB: ==> Filtered out
Aug 25 23:17:58 dkrpitv tvheadend[15102]: transcode: 0001: 02:AAC-LATM: [mp2 => aac]: filters: failed to set option: 'ch_layouts'
```

System is an RPi 4 with Arch Linux and self built TVHeadend using the PKGBUILD from AUR: https://aur.archlinux.org/packages/tvheadend-git (I modified the PKGBUILD to force build with FFMPEG 7)

If you need further info, please ask.

uknunknown:
please check if branch: https://github.com/uknunknown/tvheadend/tree/dev_ffmpg7
is fixing your problem.

dkadioglu:
> please check if branch: https://github.com/uknunknown/tvheadend/tree/dev_ffmpg7 is fixing your problem.

Unfortunately not. With that tvheadend crashes:
```
[...]

PR #1772, opened 2024-10-07
update audio abuffersink from deprecated channel_layouts to ch_layouts

uknunknown:
remove audio abuffersink options starting ffmpeg7 as recommended in ffmpeg-7.0/doc/examples/filter_audio.c line 177 --> "/* This filter takes no options. */"

Fixes: #1764

### DIFF
--- a/src/transcoding/codec/internals.h
+++ b/src/transcoding/codec/internals.h
@@ -162,6 +162,69 @@ tvh_codecs_register(void);
 void
 tvh_codecs_forget(void);
 
+/* ffmpeg constants */
+#if LIBAVCODEC_VERSION_MAJOR > 59
+// **** AUDIO ****
+// aac
+#define FF_AV_PROFILE_AAC_MAIN                  AV_PROFILE_AAC_MAIN
+#define FF_AV_PROFILE_AAC_LOW                   AV_PROFILE_AAC_LOW
+#define FF_AV_PROFILE_AAC_LTP                   AV_PROFILE_AAC_LTP
+#define FF_AV_PROFILE_MPEG2_AAC_LOW             AV_PROFILE_MPEG2_AAC_LOW
+// **** VIDEO ****
+// vp9
+#define FF_AV_PROFILE_VP9_0                     AV_PROFILE_VP9_0
+#define FF_AV_PROFILE_VP9_1                     AV_PROFILE_VP9_1
+#define FF_AV_PROFILE_VP9_2                     AV_PROFILE_VP9_2
+#define FF_AV_PROFILE_VP9_3                     AV_PROFILE_VP9_3
+// h265
+#define FF_AV_PROFILE_HEVC_MAIN                 AV_PROFILE_HEVC_MAIN
+#define FF_AV_PROFILE_HEVC_MAIN_10              AV_PROFILE_HEVC_MAIN_10
+#define FF_AV_PROFILE_HEVC_REXT                 AV_PROFILE_HEVC_REXT
+// h264
+#define FF_AV_PROFILE_H264_BASELINE             AV_PROFILE_H264_BASELINE
+#define FF_AV_PROFILE_H264_CONSTRAINED_BASELINE AV_PROFILE_H264_CONSTRAINED_BASELINE
+#define FF_AV_PROFILE_H264_MAIN                 AV_PROFILE_H264_MAIN
+#define FF_AV_PROFILE_H264_HIGH                 AV_PROFILE_H264_HIGH
+#define FF_AV_PROFILE_H264_HIGH_10              AV_PROFILE_H264_HIGH_10
+#define FF_AV_PROFILE_H264_HIGH_422             AV_PROFILE_H264_HIGH_422
+#define FF_AV_PROFILE_H264_HIGH_444             AV_PROFILE_H264_HIGH_444
+// mpeg2
+#define FF_AV_PROFILE_MPEG2_MAIN                AV_PROFILE_MPEG2_MAIN
+#define FF_AV_PROFILE_MPEG2_SIMPLE              AV_PROFILE_MPEG2_SIMPLE
+// common
+#define FF_AV_PROFILE_UNKNOWN                   AV_PROFILE_UNKNOWN
+#else
+// **** AUDIO ****
+// aac
+#define FF_AV_PROFILE_AAC_MAIN                  FF_PROFILE_AAC_MAIN
+#define FF_AV_PROFILE_AAC_LOW                   FF_PROFILE_AAC_LOW
+#define FF_AV_PROFILE_AAC_LTP                   FF_PROFILE_AAC_LTP
+#define FF_AV_PROFILE_MPEG2_AAC_LOW             FF_PROFILE_MPEG2_AAC_LOW
+// **** VIDEO ****
+// vp9
+#define FF_AV_PROFILE_VP9_0                     FF_PROFILE_VP9_0
+#define FF_AV_PROFILE_VP9_1                     FF_PROFILE_VP9_1
+#define FF_AV_PROFILE_VP9_2                     FF_PROFILE_VP9_2
+#define FF_AV_PROFILE_VP9_3                     FF_PROFILE_VP9_3
+// h265
+#define FF_AV_PROFILE_HEVC_MAIN                 FF_PROFILE_HEVC_MAIN
+#define FF_AV_PROFILE_HEVC_MAIN_10              FF_PROFILE_HEVC_MAIN_10
+#define FF_AV_PROFILE_HEVC_REXT                 FF_PROFILE_HEVC_REXT
+// h264
+#define FF_AV_PROFILE_H264_BASELINE             FF_PROFILE_H264_BASELINE
+#define FF_AV_PROFILE_H264_CONSTRAINED_BASELINE FF_PROFILE_H264_CONSTRAINED_BASELINE
+#define FF_AV_PROFILE_H264_MAIN                 FF_PROFILE_H264_MAIN
+#define FF_AV_PROFILE_H264_HIGH                 FF_PROFILE_H264_HIGH
+#define FF_AV_PROFILE_H264_HIGH_10              FF_PROFILE_H264_HIGH_10
+#define FF_AV_PROFILE_H264_HIGH_422             FF_PROFILE_H264_HIGH_422
+#define FF_AV_PROFILE_H264_HIGH_444             FF_PROFILE_H264_HIGH_444
+// mpeg2
+#define FF_AV_PROFILE_MPEG2_MAIN                FF_PROFILE_MPEG2_MAIN
+#define FF_AV_PROFILE_MPEG2_SIMPLE              FF_PROFILE_MPEG2_SIMPLE
+// common
+#define FF_AV_PROFILE_UNKNOWN                   FF_PROFILE_UNKNOWN
+#endif
+
 
 /* codec_profile_class */
 
@@ -239,11 +302,8 @@ typedef struct tvh_codec_profile_audio {
     char *language3;
     int sample_fmt;
     int sample_rate;
-#if LIBAVCODEC_VERSION_MAJOR > 59
-    AVChannelLayout channel_layout;
-#else
+    // this variable will be used also as ch_layout_u_mask when LIBAVCODEC_VERSION_MAJOR > 59
     int64_t channel_layout;
-#endif
 } TVHAudioCodecProfile;
 
 
--- a/src/transcoding/codec/profile_audio_class.c
+++ b/src/transcoding/codec/profile_audio_class.c
@@ -106,7 +106,7 @@ tvh_codec_audio_get_list_channel_layouts(TVHAudioCodec *self)
 #if LIBAVCODEC_VERSION_MAJOR > 59
             l = channel_layouts;
             ADD_ENTRY(list, map, s64, 0, str, AUTO_STR);
-            while (l->nb_channels < 32) {
+            while (l->nb_channels != 0) {
                 if (!(map = htsmsg_create_map())) {
                     htsmsg_destroy(list);
                     list = NULL;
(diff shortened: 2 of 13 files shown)


uknunknown:
@Flole998 I converted to draft because we have some mixed feedback related to ffmpeg 7. I wanted to use the infrastructure to see if any coverage tests will flag anything failing.

#everyone: Ideally we need more tests to be performed with this branch ... any help would be greatly appreciated

uknunknown:
Step 1: passed.
Step 2: user testing.
please report any good or bad experience you had with this branch.
Note:
- this will compile with ffmpeg 7 ...  make sure you check transcoding.
- I personally verified VAAPI (Intel) , software transcoding video and audio (including AAC decoding / AAC encoding)

th0ma7:
If that can be of use, pushed changes to have a SynoCommunity testing package made available for Synology users.

uknunknown:
Based on feedback from @Flole998, I made my changes as addendum to changes by André Apitzsch from  https://github.com/tvheadend/tvheadend/commit/078a822cf548b37bc474475fa57e48e9604090ee 
After we confirm the changes I will squash into a single commit.

uknunknown:
@Flole998 I updated the branch in order to fix the channel layout change. Master has this functionality broken (if you try to force 'stereo' in layout is not working ... or in my tests was not working). I kept the commits separated for now, in order to clarify first what we do with Sonar. I don't know how to fix that problem :( .
Te commits have also separate functionalities fixed:
#1 - is fixing the 3 parameters
#2 - is fixing the layout force / limit
#3 - is trying to fix Sonar report
Let's clarify the functionality (I tried to extract this info from old code):
For an input = [stereo, 5.1]
Setting             /                 Output                  
mono              |            [mono, mono]             
stereo             |            [stereo, stereo]             
5.1                  |              [stereo, 5.1]               
7.1                 |               [stereo, 5.1]               
After you clarify what we do with Sonar I will squash the commits into one.
I am afraid to push the 'Update branch' ... in case something goes wrong, I don't know how to fix it.

uknunknown:
Sonar report was fixed; I am tempted to say this branch is ready for submit ... I don't know how to fix that failure (assuming is a real failure).

src/transcoding/codec/profile_audio_class.c
```diff
@@ -358,7 +360,7 @@ const codec_profile_class_t codec_profile_audio_class = {
                 .group    = 4,
                 .opts     = PO_ADVANCED | PO_PHIDDEN,
                 .get_opts = codec_profile_audio_class_channel_layout_get_opts,
-                .off      = offsetof(TVHAudioCodecProfile, channel_layout),
+                .off      = offsetof(TVHAudioCodecProfile, ch_layout_u_mask),
```
Flole998, line 363: This will only exist for new ffmpeg versions, compilation will fail on old versions

src/transcoding/codec/profile_audio_class.c
```diff
@@ -149,9 +149,11 @@ tvh_codec_profile_audio_is_copy(TVHAudioCodecProfile *self, tvh_ssc_t *ssc)
     int ssc_channels = ssc->es_channels ? ssc->es_channels : 2;
     int ssc_sr = ssc->es_sri ? sri_to_rate(ssc->es_sri) : 48000;
 #if LIBAVCODEC_VERSION_MAJOR > 59
-    if ((self->channel_layout.order != AV_CHANNEL_ORDER_UNSPEC &&
-         ssc_channels != self->channel_layout.nb_channels) ||
-        (self->sample_rate && ssc_sr != self->sample_rate)) {
+    AVChannelLayout *ch_layout = NULL;
+    av_channel_layout_default(ch_layout, 0);
+    av_channel_layout_from_mask(ch_layout, self->ch_layout_u_mask);
+    if ((self->ch_layout_u_mask && ch_layout != NULL && ssc_channels != ch_layout->nb_channels) ||
```
Flole998, line 155: This will always be false as ch_layout is never assigned. Above there will be a null pointer dereference. Please make yourself familiar with the concept of pointers and memory management in C, that's crucial when trying to develop C-code as otherwise things like this can happen.

src/transcoding/codec/profile_audio_class.c
uknunknown: fixed.

src/transcoding/codec/profile_audio_class.c
uknunknown: I changed initialization as recommended in: build.linux/ffmpeg/ffmpeg-7.0/libavutil/channel_layout.c line 288 and uninit like recommended at line 294.

Flole998:
As you can see it still fails on some ffmpeg versions:

```
141.8 src/transcoding/codec/codecs/aac.c:30:7: error: 'AV_PROFILE_MPEG2_AAC_LOW' undeclared here (not in a function); did you mean 'FF_PROFILE_MPEG2_AAC_LOW'?
141.8    30 |     { AV_PROFILE_MPEG2_AAC_LOW, "MPEG2_LC" },
141.8       |       ^~~~~~~~~~~~~~~~~~~~~~~~
141.8       |       FF_PROFILE_MPEG2_AAC_LOW
141.8 make: *** [Makefile:722: /src/build.linux/src/transcoding/codec/codecs/aac.o] Error 1
```

uknunknown:
@Flole998 please don't merge yet ... we have to update all FF_PROFILE_* --> AV_PROFILE_* (will deprecate in 62).

uknunknown:
New issue highlighted by Sonar was not changed in this PR. Do you think we have to fix that? is a real problem?

uknunknown:
@Flole998 I realized that I forgot in PR also the ffmpeg 6--> ffmpeg 7 update. Do you want me to remove it? (note: change list is not reflecting the ffmpeg 7 update). Let me know if you want me to revert back to ffmpeg 6, or you can revert that code if you prefer.

uknunknown:
I reverted back from ffmpeg 7 to ffmpeg 6.1.1.

uknunknown:
This PR is required to fix https://github.com/tvheadend/tvheadend/issues/1764

src/transcoding/codec/internals.h
```diff
@@ -240,7 +240,7 @@ typedef struct tvh_codec_profile_audio {
     int sample_fmt;
     int sample_rate;
 #if LIBAVCODEC_VERSION_MAJOR > 59
-    AVChannelLayout channel_layout;
+    int64_t ch_layout_u_mask;
```
Flole998, line 243: You now have 2 different int64_t variables with a similar name. This can be simplified, just use the same variable in both cases, even though the naming might be a little less ideal then.

src/transcoding/codec/profile_audio_class.c
Flole998: This can also be simplified with the change mentioned above.

src/transcoding/codec/profile_class.c
```diff
@@ -80,7 +85,11 @@ tvh_codec_profile_base_open(TVHCodecProfile *self, AVDictionary **opts)
 {
     AV_DICT_SET_TVH_REQUIRE_META(opts, 1);
     // profile
+#if LIBAVCODEC_VERSION_MAJOR > 59
+    if (self->profile != AV_PROFILE_UNKNOWN) {
```
Flole998, line 89: You created NV_ENC_PROFILE_UNKNOWN in another file. Either this is not NV_ENC related, then NV_ENC_PROFILE_UNKNOWN should have a different name as it applies to all profiles, or NV_ENC_PROFILE_UNKNOWN is not defined here, then it should be moved to a common .h file, probably to internal.h

src/transcoding/codec/profile_class.c
```diff
@@ -281,7 +290,11 @@ const codec_profile_class_t codec_profile_class = {
                 .get_opts = codec_profile_class_profile_get_opts,
                 .off      = offsetof(TVHCodecProfile, profile),
                 .list     = codec_profile_class_profile_list,
+#if LIBAVCODEC_VERSION_MAJOR > 59
+                .def.i    = AV_PROFILE_UNKNOWN,
```
Flole998, line 294: See above

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -154,11 +154,20 @@ tvhva_context_profile(TVHVAContext *self, AVCodecContext *avctx)
     switch (avctx->codec->id) {
         case AV_CODEC_ID_MPEG2VIDEO:
             switch (avctx->profile) {
+#if LIBAVCODEC_VERSION_MAJOR > 59
```
Flole998, line 157: I'd just create an ifdef for these somewhere, as there are a lot of cases here.

src/transcoding/codec/profile_class.c
uknunknown: At this level AV_PROFILE_UNKNOWN applies to all codecs (also to VAAPI .. for example). 
Original implementation for NVENC has defined a new NV_ENC_PROFILE_UNKNOWN that is equal with AV_PROFILE_UNKNOWN --> NV_ENC_PROFILE_UNKNOWN was only used in nvenc.c.
I see the problem that you are trying to highlight ... now we have a series of dependencies (like NV_ENC_PROFILE_UNKNOWN must be equal with AV_PROFILE_UNKNOWN otherwise NVENCoding is broken (but that dependency is only known by me :( ).
On vaapi.c for example we are using AV_PROFILE_UNKNOWN to avoid this 'dependency'.
Frankly I wanted to keep the 'mindset from before' in the nvenc.c implementation ... but maybe we should replace NV_ENC_PROFILE_UNKNOWN with AV_PROFILE_UNKNOWN also in nvenc.c.
Let me know if you want me to replace it.

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: for my clarification: you want to replace `LIBAVCODEC_VERSION_MAJOR > 59` only in this vaapi.c file? or you are referring to full project? (hopefully only this file ... because full project has a lot of forks generated by LIBAVCODEC_VERSION_MAJOR)

Flole998: No, similar to the case above, make something like FF_AV_PROFILE_MPEG2_MAIN which is either FF_.... or AV_.... depending on the ffmpeg version. Then you have a single place where you do the ifdef and define FF_AV_PROFILE_MPEG2_MAIN to represent the relevant definition, and later on no more ifdefs and only FF_AV_PROFILE_MPEG2_MAIN. The main idea behind this is to keep the code as readable as possible.

I hope you understand what I mean, if it's not clear please just ask and I'll be happy to provide an example once I'm near a computer again.

src/transcoding/codec/profile_class.c
uknunknown: Actually AV_PROFILE_UNKNOWN was changed at some point in ffbpeg  to FF_PROFILE_UNKNOWN and ffmpeg 7 is removing all AV_* defs. I don't think I can remove the:
`#if LIBAVCODEC_VERSION_MAJOR > 59`
I can replace the NV_ENC_PROFILE_UNKNOWN with FF_PROFILE_UNKNOWN (defined in ffmpeg 7) ... because was defined by me.

Flole998: Yes I understand that. What I was proposing is this:

```
#if LIBAVCODEC_VERSION_MAJOR > 59
#define FF_AV_PROFILE_UNKNOWN FF_PROFILE_UNKNOWN
#else
#define FF_AV_PROFILE_UNKNOWN AV_PROFILE_UNKNOWN
#endif
```

and only use FF_AV_PROFILE_UNKNOWN from then on, which will always be the correct one. Same for below with the formats, have a common definition depending on the version and use that everywhere.

uknunknown: I understand now :) minimize the forks ... let me work on that. I will update the PR when I have a proposal.

src/transcoding/codec/profile_audio_class.c
uknunknown: Done.

src/transcoding/codec/internals.h
uknunknown: Done. added a comment for tracking in the future.

src/transcoding/codec/profile_class.c
uknunknown: I move all the forks in a single place ... where is accessed by all other files. Let me know if this is what you had in mind.

src/transcoding/codec/profile_class.c
uknunknown: Done.

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: @Flole998 please confirm if the changes are matching your expectation ... if Yes I will squash commits ... if No I will return to previous.
Thank you for understanding.

src/transcoding/codec/profile_audio_class.c
```diff
@@ -179,17 +179,14 @@ tvh_codec_profile_audio_open(TVHAudioCodecProfile *self, AVDictionary **opts)
...
 #if LIBAVCODEC_VERSION_MAJOR > 59
-    if (self->ch_layout_u_mask) {
-        AV_DICT_SET_INT(opts, "ch_layout_u_mask", self->ch_layout_u_mask,
-                        AV_DICT_DONT_OVERWRITE);
-    }
+        AV_DICT_SET_INT(opts, "ch_layout_u_mask", self->channel_layout,
 #else
-    if (self->channel_layout) {
         AV_DICT_SET_INT(opts, "channel_layout", self->channel_layout,
+#endif
```
Flole998, line 187: Please don't split the argument list with an ifdef. Just keep the ```
AV_DICT_DONT_OVERWRITE);
    }
```
so the code within the ifdef is "complete" in both branches and you don't have fragments of it outside the ifdef.

src/transcoding/codec/profile_class.c
Flole998: Yes, and I think you can see now why, it just looks a lot cleaner and the code itself is shorter.

src/transcoding/transcode/hwaccels/vaapi.c
Flole998: Yes, looks great

src/transcoding/codec/profile_audio_class.c
uknunknown: Done.

Flole998:
Lets squash and merge this then :)

uknunknown:
upon review I found one more ifdef that is splitting a line ... let me fix also this one. 
split

uknunknown:
I only fixed two ifdef splits from:
src/transcoding/transcode/audio.c
src/transcoding/transcode/video.c
From my point of view is ready.

PR #1792, opened 2024-12-30
update vaapi transcoding as recommended by ffmpeg 6.1.1/doc/examples/…

uknunknown:
- update vaapi transcoding as recommended by ffmpeg 6.1.1/doc/examples/vaapi_*.c
- src/transcoding/transcode/hwaccels/vaapi was completely redesigned

### DIFF
--- a/src/libav.h
+++ b/src/libav.h
@@ -54,6 +54,13 @@ This list must be updated every time we use a new AV_CODEC_ID
 #define AV_CODEC_ID_DVB_TELETEXT CODEC_ID_DVB_TELETEXT
 #endif
 
+// Enable new transcoding starting ffmpeg 4
+#if LIBAVCODEC_VERSION_MAJOR < 58
+#define ENABLE_FFMPEG4_TRANSCODING 0
+#else
+#define ENABLE_FFMPEG4_TRANSCODING 1
+#endif
+
 enum AVCodecID streaming_component_type2codec_id(streaming_component_type_t type);
 streaming_component_type_t codec_id2streaming_component_type(enum AVCodecID id);
 void libav_set_loglevel(void);
--- a/src/transcoding/transcode/video.c
+++ b/src/transcoding/transcode/video.c
@@ -226,10 +226,37 @@ tvh_video_context_open_encoder(TVHContext *self, AVDictionary **opts)
 #if ENABLE_HWACCELS
     self->oavctx->coded_width = self->oavctx->width;
     self->oavctx->coded_height = self->oavctx->height;
-    if (hwaccels_encode_setup_context(self->oavctx, self->profile->low_power)) {
+    #if ENABLE_FFMPEG4_TRANSCODING
+    // hwaccel is the user input for Hardware acceleration from Codec parameteres
+    int hwaccel = -1;
+    if ((hwaccel = tvh_codec_profile_video_get_hwaccel(self->profile)) < 0) {
         return -1;
     }
+    if (_video_filters_hw_pix_fmt(self->oavctx->pix_fmt)){
+        // encoder is hw accelerated
+        if (hwaccel) {
+            // decoder is hw accelerated 
+            // --> we initialize encoder from decoder as recomanded in: 
+            // ffmpeg-6.1.1/doc/examples/vaapi_transcode.c line 169
+            if (hwaccels_initialize_encoder_from_decoder(self->iavctx, self->oavctx)) {
+                return -1;
+            }
+        }
+        else {
+            // decoder is sw
+            // --> we initialize as recommended in:
+            // ffmpeg-6.1.1/doc/examples/vaapi_encode.c line 145
+            if (hwaccels_encode_setup_context(self->oavctx)) {
+#else
+            if (hwaccels_encode_setup_context(self->oavctx, self->profile->low_power)) {
 #endif
+                return -1;
+            }
+#if ENABLE_FFMPEG4_TRANSCODING
+        }
+    }
+#endif
+#endif // from ENABLE_HWACCELS
 
     // XXX: is this a safe assumption?
     if (!self->iavctx->framerate.num) {
--- a/src/transcoding/transcode/video.c
+++ b/src/transcoding/transcode/video.c
@@ -226,7 +226,7 @@ tvh_video_context_open_encoder(TVHContext *self, AVDictionary **opts)
 #if ENABLE_HWACCELS
     self->oavctx->coded_width = self->oavctx->width;
     self->oavctx->coded_height = self->oavctx->height;
-    #if ENABLE_FFMPEG4_TRANSCODING
+#if ENABLE_FFMPEG4_TRANSCODING
     // hwaccel is the user input for Hardware acceleration from Codec parameteres
     int hwaccel = -1;
     if ((hwaccel = tvh_codec_profile_video_get_hwaccel(self->profile)) < 0) {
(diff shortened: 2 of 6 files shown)


src/libav.h
```diff
@@ -54,6 +54,13 @@ This list must be updated every time we use a new AV_CODEC_ID
 #define AV_CODEC_ID_DVB_TELETEXT CODEC_ID_DVB_TELETEXT
 #endif
 
+// Enable new transcoding starting ffmpeg 6
+#if LIBAVCODEC_VERSION_MAJOR < 60
+#define ENABLE_NEW_TRANSCODING 0
```
chewitt, line 59: Can we not use the `#if LIBAVCODEC_VERSION_MAJOR < 60` test everywhere instead of adding a new variable? I'm generally against calling things "new" in code because the new method is inevitably superseded and becomes another old method at some point requiring a future "new2" or "nextgen" naming silliness.

uknunknown: @chewitt let me clarify why I choose this solution (maybe not the best): I don't know how far back this new methodology of transcoding works ... so in order to easily test and update later on, is better to update a single line that will allow users to test functionality. Maybe the name was not the best option ... frankly I am open to any naming convention. ENABLE_TRANSCODING_V2 ? this is a little mode future proof. 
More importantly: @Flole998 do you think the failures are real failures?

src/transcoding/transcode/internals.h
```diff
@@ -164,7 +164,11 @@ struct tvh_context {
     char *hw_accel_device;
     AVBufferRef *hw_device_ref;
     void *hw_accel_ictx;
+#if ENABLE_NEW_TRANSCODING
+    AVBufferRef *hw_frame_octx;
```
Flole998, line 168: You are basically just renaming this and introduce a bunch of preprocessor-if-directives for this? That's about the worst thing you could do from a maintanability point of view.

src/transcoding/transcode/video.c
```diff
@@ -255,13 +282,21 @@ tvh_video_context_open_filters(TVHContext *self, AVDictionary **opts)
...
+#endif
+
     // source args
     memset(source_args, 0, sizeof(source_args));
     if (str_snprintf(source_args, sizeof(source_args),
             "video_size=%dx%d:pix_fmt=%s:time_base=%d/%d:pixel_aspect=%d/%d",
             self->iavctx->width,
             self->iavctx->height,
+#if ENABLE_HWACCELS && ENABLE_NEW_TRANSCODING
+            av_get_pix_fmt_name(pix_fmt),
```
Flole998, line 296: Why is the new way of getting the pix_fmt_name better than the old way?

src/transcoding/transcode/internals.h
uknunknown: ENABLE_NEW_TRANSCODING was not introduced only to 'mask' the new variable change ... also changes a lot of other things. Please check src\transcoding\transcode\hwaccels\vaapi.c
There is a difference between TVH variable names and ffmpeg (example) variable names. In order for me to bring TVH naming convention to ffmpeg naming conventions I decided to rename that variable only for new transcoding (v.2).
This can be resolved in one of the 3 options:
1. option 1: Rename that variable in transcoding v.1 in TVH to match ffmpeg.
2. option 2: Keep the TVH naming and rename the variable in ffmpeg code imported in TVH. This is not recommended because will fork TVH away from ffmpeg.
3. option 3: Leave v.1 with old name from TVH and v.2 with new name from ffmpeg --> and transition to new names only for new ffmpeg. 

I selected option 3 because my expectation was that we will deprecate at some point transcoding v.1 and that code will be removed (at some later point). After this deprecation TVH will be aligned to ffmpeg.
Before you tell me what option do you prefer ... please think about the question: how far back should we prove the transcoding v.2 work in order for you to accept the transition directly to v.2 (and remove the v.1 code completely)? if it works with ffmpeg 4 are you OK to remove v.1?

src/transcoding/transcode/video.c
uknunknown: This change is not required for vaapi implementation .... but will be required for QSV implementation: please check https://github.com/tvheadend/tvheadend/pull/1781/files#diff-8ecb3c20d3cb77afadf97cb19260496ae8a05ac2e03486107abe5d15e30e6fceR315
If you want me to remove that part until we implement QSV ... please let me know.

uknunknown: removed

src/transcoding/transcode/internals.h
uknunknown: Experiment 1: I changed makefile.ffmpeg to compile with ffmpeg 5.1.2 and enabled transcoding v.2 for ffmpeg 5 --> all my tests have worked as expected.
Experiment 2: I changed makefile.ffmpeg to compile with ffmpeg 4.4 --> unfortunately I can't compile due to some errors. But the error is not related to the new transcoding (/tvheadend_new_transcoding/build.linux/ffmpeg/build/ffmpeg/lib/libavcodec.a] Error 2)
According to the date when ffmpeg has added the example (used for transcoding implementation) https://github.com/FFmpeg/FFmpeg/commit/23db3a1ae6d1be3438aec73c4dc91185d7958300#diff-860cbca3d54a06c6224b8b88bba80bbb62f1d7d8f34d3062b1dad5cc08ec2984 ffmpeg 4.4 should work also.

uknunknown: Experiment 3: I generated a new branch from: commit https://github.com/tvheadend/tvheadend/commit/470f02fb3f00d3f88e61303cd5db7ec303d0145d + new transcoding changes -->results is available on branch (if you want to test it yourself) https://github.com/uknunknown/tvheadend/tree/dev_transc_ffmpg4 compiled with ffmpeg 4.4.3
HEVC VAAPI was not working properly even before I change anything, but I tested H264 before and after that change and was working the same.
@Flole998 taking into account that VAAPI transcoding is only made with x86 platforms, I don;t think that any OS is using older than ffmpeg 4.4, is this far enough back to accept the transcoding change without keeping the old implementation? If you want to go further back I can try the same change on some older build (how old should that be)?

src/libav.h
uknunknown: change the name to be connected with ffmpeg

src/transcoding/transcode/internals.h
uknunknown: remove the rename ... simplified as much as possible the changes.

uknunknown:
@Flole998 my proposal is to use this switch (ENABLE_FFMPEG4_TRANSCODING) for one year or so; if we have any complains we can ask users to switch back and retest. If we don't find an issue, we can remove the old implementation and we keep the new implementation (some concept like deprecation).

jahutchi:
I had been having all sorts of off issues with TVH Built-in VAAPI transcoding on my Intel NUC Coffee Lake. Ended up using TS-Spawn but that was far from perfect and could never get it working reliably.

I came across this pull request so thought I would give it a go. My VAAPI transcoding is now working perfectly - so hope this gets merged one day. Great Job 👍

chewitt:
I'm still not understanding why the abstration and extra ifdef for `ENABLE_FFMPEG4_TRANSCODING 0` is required when testing for `LIBAVCODEC_VERSION_MAJOR < 58` does the same thing, no? NB: I am not a coding developer (any time I fiddle with code and it still compiles I'm surprised) and thus might be missing something obvious. That's also the reason I should not be a formal reviewer for the PR!

uknunknown:
If you replace `ENABLE_FFMPEG4_TRANSCODING 0` with `LIBAVCODEC_VERSION_MAJOR < 58` will be very hard for user to go back and forth ... because in order to go back to old implementation is forced to change ffmpeg4 in `makefile.ffmpeg`. Practically master will not compile now with ffmpeg4 ... as it is (if you enable transcoding). You need to make some changes, beside going back to some very old version of master (I made this exercise in https://github.com/uknunknown/tvheadend/tree/dev_transc_ffmpg4).
**In conclusion**: is to make it more accessible for users to go back and forth between implementations.

src/transcoding/transcode/video.c
```diff
@@ -226,10 +226,37 @@ tvh_video_context_open_encoder(TVHContext *self, AVDictionary **opts)
...
+            // decoder is sw
+            // --> we initialize as recommended in:
+            // ffmpeg-6.1.1/doc/examples/vaapi_encode.c line 145
+            if (hwaccels_encode_setup_context(self->oavctx)) {
+#else
+            if (hwaccels_encode_setup_context(self->oavctx, self->profile->low_power)) {
+#endif
+                return -1;
+            }
+#if ENABLE_FFMPEG4_TRANSCODING
```
Flole998, line 255: This is wild... If you need an extra #ifdef just to close some } then your code is most likely hard to understand. Just copy the `return -1`, overall readability will be improved even though it's there twice then.

uknunknown: I am sorry Flole ... I am not sure what was in my head. Thank you for catching that. Fixed.
Lets squash after you finish the review.

Flole998: Much better, thanks!

Flole998:
Lets squash and merge

PR #1797, opened 2025-02-02
Fix CI builds

Flole998:


Autobuild/armv7l.sh
```diff
@@ -1 +1,2 @@
 AUTOBUILD_CONFIGURE_EXTRA="${AUTOBUILD_CONFIGURE_EXTRA:-} --arch=armhf"
+EXTRA_X265_CMAKE_OPTS="${EXTRA_X265_CMAKE_OPTS:-} -DCROSS_COMPILE_ARM=1"
```
vassilit, line 2: add newline ?

git-developer:
The build of x265 currently fails on distros containing CMake < 3.2 because of the following line in in [x265/Version.cmake](https://github.com/videolan/x265/blob/419182243fb2e2dfbe91dfc45a51778cf704f849/source/cmake/Version.cmake#L155):
```cmake
string(SUBSTRING "${git_repositorychangeset}" 0 9 X265_REVISION_ID)
```
Cause: `git_repositorychangeset` is shorter than 9 bytes. [SUBSTRING](https://cmake.org/cmake/help/latest/command/string.html#substring) tolerates this syntax in CMake 3.2 and above.

My proposal is to change the above line to
```cmake
string(REGEX MATCH "^.{1,9}" X265_REVISION_ID "${git_repositorychangeset}")
```

This change must be applied either [upstream](https://github.com/videolan/x265) or as patch in this project (I don't know how to do that).

An alternative would be to drop support of EOL distros.

WDYT?

chewitt:
@git-developer It's always desirable to upstream patches, but as the upstream source has been static for several years it might take a while to find out how/where to submit, and to get it merged. Please do that, but also submit a local patch for Tvheadend. See https://github.com/tvheadend/tvheadend/tree/master/support/patches and https://github.com/tvheadend/tvheadend/blob/master/Makefile.ffmpeg#L66 for how that's been done before.

git-developer:
A patch is prepared in https://github.com/tvheadend/tvheadend/pull/1822. Is it possible to trigger a CI build?

Flole998:
For some reason it's not possible to run a CI build on that PR, I'll just merge it as it's for my testing branch anyways and I'll need to cleanup anyway, so if it's not helping I'll simply revert it when cleaning up. Right now 10 failing, let's see if that makes it better or worse.

git-developer:
Seems to help, there are now 2 more problems concerning x265:

[armv7l](https://github.com/tvheadend/tvheadend/actions/runs/14954120952/job/42007501275?pr=1797#step:3:3755):
```
/usr/bin/ld: libx265.a(cpu.cpp.o): in function `x265::cpu_detect(bool)':
cpu.cpp:(.text+0x4): undefined reference to `x265_cpu_fast_neon_mrc_test'
```

[aarch64](https://github.com/tvheadend/tvheadend/actions/runs/14954120952/job/42007501321?pr=1797#step:3:4161):
```
cc1: error: unknown value 'armv9-a+i8mm+sve2' for '-march'
cc1: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a armv8.6-a armv8-r native
```

git-developer:
`armv9-a` was [added in gcc-12](https://gcc.gnu.org/gcc-12/changes.html#arm-targets).

Flole998:
> `armv9-a` was [added in gcc-12](https://gcc.gnu.org/gcc-12/changes.html#arm-targets).

If you have no (easy) fix just disable x265 for those architectures. I wanted to go ahead and do fhat but haven't had time to do it.

IIRC the armv7 issue did appear after updating x265 but I am not sure.

chewitt:
The armv7 issue appears to be discussed here: https://github.com/kapitainsky/handbreak-RaspberryPi#6-we-have-to-make-quick-and-dirty-hack-to-x265-source-code

Flole998:
I think this is the same as in https://bitbucket.org/multicoreware/x265_git/issues/977/undefined-symbols-when-cross-building-for ?

They mention that this fix is not correct in the comments.

chewitt:
Some of these might be relevant? https://bitbucket.org/multicoreware/x265_git/pull-requests/23/commits

git-developer:
I think disabling x265 should be no big problem on `armv7l`, because the hardware for this platform is probably not able to deal with it either.

`aarch64` is a different case, hardware might be capable of x265 here. Maybe it's possible to use `armv8` instead of `armv9` for gcc < 12?

EricV2:
More generally, using external libs would relieve the build problems. Here update are on hold because some  package used by tvheadend fails to build.

CI is fine but so far it blocks commits to tvheadend core that are not related to tvheadend. On old arch, OS, why not use dynamic build for most packages (ffmpeg,x265,...)?

Flole998:
If that's something you want to work on, feel free to do so. It would be really appreciated. I wouldn't call Ubuntu 22.04 on aarch64 old, it's not even "End of standard support". Also for raspberry pi it's the latest OS for armv7. So even if all old builds are abandoned, it wouldn't change a thing as recent builds have the exact same issue. Fixing it for one fixes it for all.

In fact as part of this PR support for Debian Jessie was dropped as that would require more effort and hacks to get it to work again (signing key expired).

Flole998:
We're now back to `error: there are no arguments to 'vld1q_s16_x4' that depend on a template parameter, so a declaration of 'vld1q_s16_x4' must be available` which also sounds familiar, I think I have seen that one before.

I might be wrong but this sounds like some cpu extension which should be available but is not for some reason.

git-developer:
The compiler [suggests](https://github.com/tvheadend/tvheadend/actions/runs/14956836403/job/42013633668#step:3:1664) using `vld1q_s64` instead of `vld1q_s16_x4`.

git-developer:
The build logs differ from each other, so `vld1q_s64` might not available everywhere and is probably no common fix.

[gcc#71233](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71233) lists `vld1q_s16_x4` as unavailable in gcc. Newer versions contain if in [arm_neon.h](https://github.com/gcc-mirror/gcc/blob/40cecd49fba1db0cb98b81016d0742398ce3c861/gcc/config/aarch64/arm_neon.h#L12627).

Flole998:
> The compiler [suggests](https://github.com/tvheadend/tvheadend/actions/runs/14956836403/job/42013633668#step:3:1664) using `vld1q_s64` instead of `vld1q_s16_x4`.

I don't think thats correct though. This function is simply not available on older compilers and I think that is fine, if someone really wants x265 they would have to upgrade to a more recent OS. We are down to 6 failing now, one is the raspberry armv7 and the other ones are aarch64 builds.

git-developer:
Nice. I didn't notice that the failing builds all occur on older distros. Then it might indeed be appropriate to exclude x265 from these.

Flole998:
Well I wanted to originally exclude all of those that were failing before you jumped in to rescue a few more, so good job and thanks a lot! It should be all green now and then I'll squash all commits and it can be merged.

Flole998:
I just changed our branch protection rules for the new namings, once I get near a computer I am going to squash and merge this

Flole998:
Squashed and automerge enabled, it will be merged once all builds have completed. Huge thanks to @git-developer, you really helped a lot to make this happen!

git-developer:
You're welcome. I hope this unblocks some of the waiting PRs. Thanks for your quick reaction!

chewitt:
@git-developer many thanks for the help on getting this over the line, finally! 👍🏻

PR #1800, opened 2025-02-11
allow NVENC, VAAPI and MMAL to coexist in the same build

uknunknown:
- allow NVENC, VAAPI and MMAL to coexist in the same build.
- give the user the capability for force hw decoder or to match the hw decoder with hw encoder

### DIFF
--- a/src/transcoding/codec/profile_video_class.c
+++ b/src/transcoding/codec/profile_video_class.c
@@ -33,6 +33,25 @@ scaling_mode_get_list( void *o, const char *lang )
     return strtab2htsmsg(tab, 1, lang);
 }
 
+static htsmsg_t *
+hwaccel_get_list( void *o, const char *lang )
+{
+    static const struct strtab tab[] = {
+        { N_("auto (recommended)"),                 HWACCEL_AUTO},
+#if ENABLE_VAAPI
+        { N_("prioritize VAAPI"),                   HWACCEL_PRIORITIZE_VAAPI},
+#endif
+#if ENABLE_NVENC
+        { N_("prioritize NVDEC"),                   HWACCEL_PRIORITIZE_NVDEC},
+#endif
+#if ENABLE_MMAL
+        { N_("prioritize MMAL"),                    HWACCEL_PRIORITIZE_MMAL},
+#endif
+    };
+    return strtab2htsmsg(tab, 1, lang);
+}
+
+
 /* TVHCodec ================================================================= */
 
 static htsmsg_t *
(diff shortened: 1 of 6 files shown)


src/transcoding/codec/profile_video_class.c
```diff
@@ -207,12 +227,13 @@ const codec_profile_class_t codec_profile_video_class = {
                 .def.i    = 0,
             },
             {
-                .type     = PT_BOOL,
+                .type     = PT_INT,
```
Flole998, line 230: This looks like it will cause problems during migrations. I suggest leaving this as-is and introducing a new field. Possibly during migration it could set the new field to "match" if this one was set, and to "disable" if this one wasn't set.

uknunknown: got it :) ... let me work on this direction.

Flole998: Feel free to double-check how it's actually stored in the config, if a bool is stored as an int then this would actually work and it could only cause issues if it's changed to something >1 and then a downgrade of Tvheadend is done.

uknunknown: I reverted back to PT_BOOL. I had to touch the code in several places (especially in javascript).

src/transcoding/transcode/stream.c
```diff
@@ -57,16 +57,26 @@ tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
...
+    if ((hwaccel_detailes = tvh_codec_profile_video_get_hwaccel_details(profile)) < 0) {
+        return -1;
+    }
 #if ENABLE_MMAL
     if (idnode_is_instance(&profile->idnode,
                            (idclass_t *)&codec_profile_video_class)) {
-      if (tvh_codec_profile_video_get_hwaccel(profile) > 0) {
+      if (hwaccel &&
+         ((hwaccel_detailes == HWACCEL_AUTO && strstr(profile->codec_name, "mmal")) ||
+         hwaccel_detailes == HWACCEL_PRIORITIZE_MMAL)) {
```
chewitt, line 79: s/detailes/details .. same for all other instances

chewitt:
I assume the support for MMAL is continuance of pre-existing RPi capabilities? MMAL and OMX need 32-bit userspace and were formally deprecated with the RPi4/5 oriented switch to 64-bit userspace. On 64-bit userspace distros RPi4/5 need to use the V4L2 m2m encoder. Perhaps something for a follow-up PR.

src/transcoding/transcode/stream.c
uknunknown: fixed.

PR #1814, opened 2025-04-14
add mpegts parameters from input stream

uknunknown:
- add service_name, service_provider, mpegts_transport_id, mpegts_service_type, mpegts_pmt_start_pid, mpegts_start_pid, mpegts_service_id, mpegts_original_service_id

### DIFF
--- a/src/muxer/muxer_libav.c
+++ b/src/muxer/muxer_libav.c
@@ -356,6 +356,7 @@ lav_muxer_init(muxer_t* m, struct streaming_start *ss, const char *name)
   AVDictionary *opts = NULL;
   lav_muxer_t *lm = (lav_muxer_t*)m;
   char app[128];
+  char mpegts_info[8];
 #if LIBAVCODEC_VERSION_MAJOR > 58
   const AVOutputFormat *fmt;
 #else
@@ -396,15 +397,24 @@ lav_muxer_init(muxer_t* m, struct streaming_start *ss, const char *name)
     return -1;
   }
   av_dict_set(&oc->metadata, "title", name, 0);
-  av_dict_set(&oc->metadata, "service_name", name, 0);
-  av_dict_set(&oc->metadata, "service_provider", app, 0);
+  av_dict_set(&lm->lm_oc->metadata, "title", name, 0);
+  if (ss->ss_si.si_service){
+    av_dict_set(&lm->lm_oc->metadata, "service_name", ss->ss_si.si_service, 0);
+    tvhdebug(LS_LIBAV,  "service_name = %s", ss->ss_si.si_service);
+  } else
+    av_dict_set(&oc->metadata, "service_name", name, 0);
+  if (ss->ss_si.si_provider){
+    av_dict_set(&lm->lm_oc->metadata, "service_provider", ss->ss_si.si_provider, 0);
+    tvhdebug(LS_LIBAV,  "service_provider = %s", ss->ss_si.si_provider);
+  } else
+    av_dict_set(&oc->metadata, "service_provider", app, 0);
 
   lm->bsf_h264_filter = av_bsf_get_by_name("h264_mp4toannexb");
   if (lm->bsf_h264_filter == NULL) {
     tvhwarn(LS_LIBAV,  "Failed to get BSF: h264_mp4toannexb");
   }
   lm->bsf_hevc_filter = av_bsf_get_by_name("hevc_mp4toannexb");
-  if (lm->bsf_h264_filter == NULL) {
+  if (lm->bsf_hevc_filter == NULL) {
     tvhwarn(LS_LIBAV,  "Failed to get BSF: hevc_mp4toannexb");
   }
   lm->bsf_vp9_filter = av_bsf_get_by_name("vp9_superframe");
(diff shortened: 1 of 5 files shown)


src/muxer/muxer_libav.c
```diff
@@ -441,6 +451,42 @@ lav_muxer_init(muxer_t* m, struct streaming_start *ss, const char *name)
     av_dict_set(&opts, "ism_lookahead", "0", 0);
   }
 
+  if(lm->m_config.m_type == MC_MPEGTS) {
+    // implement as many parameters from: https://ffmpeg.org/ffmpeg-formats.html#mpegts-1
```
chewitt, line 455: the comment should describe the functional change and/or why they are included, not the act of implementation

chewitt:
Please describe why these are being included. What problem does the change solve?

src/muxer/muxer_libav.c
uknunknown: should I replace with:
`// implemented using documentation: https://ffmpeg.org/ffmpeg-formats.html#mpegts-1`

chewitt: better to describe why they are added or the purpose they serve

uknunknown:
This submission is adding some parameters for mpegts container. Some tools are more strict than other when it comes to standards ... like discussed on https://tvheadend.org/d/8186-tvheadend-buildinstall-on-lubuntu-2004-64bit-with-igpu-8-transcoding/101.
Let me clarify what this PR does:
1. is transferring as many parameters as possible from input stream to mpeg-ts/av-lib container.
2. Some parameters cannot be transferred with the same value due to limited amount of commands available: https://ffmpeg.org/ffmpeg-formats.html#mpegts-1
3. A good example is: mpegts_pmt_start_pid and mpegts_start_pid that allows me to define the start value; and the rest of the streams will have id incremented by 1 --> is not a perfect situation but I did what I could with commands available.
4. some streams from Private data category (like Teletext) are dropped in transcoding process due to deprecation / non-compliance with DVB.
5. I didn't managed to transfer EIT ... we can target that for future improvement.

Overall the end results are showing in this side-by-side comparison:
comparison

src/muxer/muxer_libav.c
uknunknown: done.

uknunknown:
@Flole998 we need your help to asses what and if we should add a series of additions to this PR. There some improvements discussed in https://tvheadend.org/d/8186-tvheadend-buildinstall-on-lubuntu-2004-64bit-with-igpu-8-transcoding/131  
I can try to address #1 (because some users might use those default MPEG-TS parameters) in this PR. We can work on #2 later on (on a different PR).
Let me know if you agree to add #1 "Reset MPEG-TS SI parameters" to this PR (so I can start working on it).

Flole998:
> Reset MPEG-TS SI parameters

The idea is to basically have an option to disable the current behaviour (which does what?) and "start from 0"? Could you explain on a high level what that change should do (like "PID is currently set to the input PID, the change would set it to xyz")? It feels like there should be individual options to override the changes (like "Output PID, -1 to copy the input PID/leave it unchanged").

uknunknown:
My initial idea was to reset to 'some default value'; but your idea to copy the functionality from profile pass is much better/flexible. ffmpeg will not allow me to change all the parameters from pass :( ... because some don't have a ffmpeg implementation/command ... but I can add as many I can :) .
We can keep the same override implementation from pass profile - for consistency.
https://github.com/tvheadend/tvheadend/blob/master/src/profile.c#L1364
 `      .desc     = N_("Rewrite service identifier (SID) using the specified "
                     "value (usually 1). Zero means no rewrite."),`

uknunknown:
I found the following correspondence: 
tvh-pass  --> ffmpeg
"Rewrite Service ID"  --> "mpegts_service_id"
"Rewrite PMT" --> "mpegts_pmt_start_pid"
"Rewrite PAT" 
"Rewrite SDT"
"Rewrite NIT" --> "nit"
"Rewrite EIT"

If you are OK dropping "Rewrite PAT" , "Rewrite SDT" and "Rewrite EIT" --> I can start implementing the other ones.

Flole998:
I'm not entirely sure about the functionality of those features. I believe they rewrite the tables to contain the new service ID?

This wouldn't affect the pass profile in any way, right? So all you would do is add additional options to different profile types, right?

uknunknown:
>  I believe they rewrite the tables to contain the new service ID?

Yes

> This wouldn't affect the pass profile in any way, right? So all you would do is add additional options to different profile types, right?

That is correct ... actually will be added only to profile_transcode_class profile: https://github.com/tvheadend/tvheadend/blob/master/src/profile.c#L2475 (under Container).
Also I will try to disable by default (from javascript) those fields and enable those only when MPEG-TS/av-lib (as Container) is selected.

uknunknown:
@Flole998 during my implementation I found that: https://github.com/tvheadend/tvheadend/blob/master/src/profile.c#L1760
left is `int` and right is `uint16_t` and we don't have an explicit cast.
Do you think is a problem? Compiler does the job of casting automatically?

uknunknown:
Or we should change the union to replace (int) with (uint16_t):
int_uint16_t

uknunknown:
@Flole998 please take a look at the issues highlighted by sonar ... let me know if you think requires more work.

src/profile.c
```diff
@@ -2484,6 +2537,51 @@ const idclass_t profile_transcode_class =
...
+      .group    = 3
+    },
+    {
+      .type     = PT_BOOL,
+      .id       = "rewrite_pmt",
+      .name     = N_("Rewrite PMT"),
+      .desc     = N_("Rewrite PMT (Program Map Table) packets to only "
+                     "include information about the currently-streamed "
+                     "service. "
+                     "Rewrite can be unset only if 'Rewrite Service ID' "
```
Flole998, line 2561: This sentence required quite some thinking about what it means. How about "Must  be set if ..... is non-zero"?

Flole998: Or instead of "non-zero" maybe call it "active" or "used", whatever fits best. The current double negation ist just more difficult to understand IMO.

src/webui/static/app/profile.js
```diff
@@ -2,6 +2,38 @@
...
+            // enable settings for MPEG-TS/av-lib
+            // /src/muxer.h:32  MC_MPEGTS      = 2,
+            if (form.findField('container').getValue() == 2) {
+                form.findField('sid').setDisabled(false);
+                form.findField('rewrite_pmt').setDisabled(false);
+                form.findField('rewrite_nit').setDisabled(false);
+            }
+            else {
+                form.findField('sid').setDisabled(true);
+                form.findField('rewrite_pmt').setDisabled(true);
```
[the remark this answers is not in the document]
uknunknown: The reason of "disable" is to show the user that those settings are not used if container is different than MPEG-TS/av-lib. doesn't matter if are enabled or disabled ... are not used. When MPEG-TS/av-lib is selected those components will become active (to be changed by the user).
Let me know if you disagree with this mind set.
image

image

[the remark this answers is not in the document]
Flole998: Got it, I thought this was to force-enable them when "Rewrite service ID" is used.

src/profile.c
uknunknown: frankly this is a copy/paste from pass profile:
image

If you want to change this one ... we should change both of them (to be in sync). Let me know your decision.

Issue #1827, opened 2025-05-17
[BUG]: Failed to transcode audio track


xswqaz:
### Tvheadend Version

v4.3 (development)

### Build Number

4.3-2392~g03272650d

### Platform

Other..

### Architecture

Intel x86_64

### Category

Transcoding

### Description

Video transcode with vaapi is no problem
But, It failed to transcode audio track ac3 to aac
Configuration is not changed from previous version so I think it is bug of latest version

```
May 17 13:24:10 tvheadend[242418]: transcode: 0001: 01:MPEG2VIDEO: ==> Using profile vaapi
May 17 13:24:10 tvheadend[242418]: transcode: 0001: 02:AC3: ==> Filtered out
```

My aac codec  profile
```
{
        "bit_rate": 64,
        "qscale": 0,
        "coder": "twoloop",
        "tracks": 1,
        "sample_fmt": -1,
        "sample_rate": 0,
        "channel_layout": 0,
        "name": "webtv-aac",
        "description": "WEBTV codec AAC",
        "codec_name": "aac",
        "profile": 0
}
```

Flole998:
This bug report is missing debug logs.

xswqaz:
Here is debug log related with AC3
```
May 17 14:24:12 tvheadend[243463]: <none>: esfilter: "ATSC/303MHz/(HD) YTN" 002 05732 AC3
May 17 14:24:12 tvheadend[243463]: transcode: 0007: 02:AC3: ==> Filtered out
May 17 14:24:12 tvheadend[243463]: tbl-base: pmt:    type AC3 position 0
May 17 14:24:12 tvheadend[243463]: pcr: ATSC/303MHz/(HD) YTN: audio DTS: -9223372036854775808 dts 6638703772 [AC3/5732]
May 17 14:24:12 tvheadend[243463]: parser: deliver (pkt stream 2 AC3 pcr <unset> dts <unset> pts <unset> dur 0 len 0 err 1)
May 17 14:24:12 tvheadend[243463]: pcr: ATSC/303MHz/(HD) YTN: audio DTS: 6638711257 dts 6638721052 [AC3/5732]
May 17 14:24:12 tvheadend[243463]: parser: deliver (pkt stream 2 AC3 pcr 6638711257 dts 6638703772 pts 6638703772 dur 2880 len 1024 err 0)
[...]

PR #1828, opened 2025-05-19
fix bug 1827: video hw accel should only be applied for video streams

uknunknown:
- video hw accel should only be applied for video streams

Fixes: #1827

### DIFF
--- a/src/transcoding/transcode/stream.c
+++ b/src/transcoding/transcode/stream.c
@@ -57,70 +57,66 @@ tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
 {
     enum AVCodecID icodec_id = streaming_component_type2codec_id(ssc->es_type);
     const AVCodec *icodec = NULL, *ocodec = NULL;
-    int hwaccel = -1;
-    int hwaccel_details = -1;
 
     if (icodec_id == AV_CODEC_ID_NONE) {
         tvh_stream_log(self, LOG_ERR, "unknown decoder id for '%s'",
                        streaming_component_type2txt(ssc->es_type));
         return -1;
     }
-    if ((hwaccel = tvh_codec_profile_video_get_hwaccel(profile)) < 0) {
-        return -1;
-    }
-    if ((hwaccel_details = tvh_codec_profile_video_get_hwaccel_details(profile)) < 0) {
-        return -1;
-    }
-#if ENABLE_MMAL
-    if (idnode_is_instance(&profile->idnode,
-                           (idclass_t *)&codec_profile_video_class)) {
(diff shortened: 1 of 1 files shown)


src/transcoding/transcode/stream.c
```diff
@@ -57,70 +57,74 @@ tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
...
-    if (idnode_is_instance(&profile->idnode,
-                           (idclass_t *)&codec_profile_video_class)) {
-      if (hwaccel &&
-         ((hwaccel_details == HWACCEL_AUTO && strstr(profile->codec_name, "mmal")) ||
-         hwaccel_details == HWACCEL_PRIORITIZE_MMAL)) {
-        if (icodec_id == AV_CODEC_ID_H264) {
-            icodec = avcodec_find_decoder_by_name("h264_mmal");
-        } else if (icodec_id == AV_CODEC_ID_MPEG2VIDEO) {
-            icodec = avcodec_find_decoder_by_name("mpeg2_mmal");
+        if (idnode_is_instance(&profile->idnode,
```
Flole998, line 77: You are now nesting 2 if's here, which sonarqube is flagging. The conditions can be combined. Use a new line when you think its appropriate to make the meaning more clear, like
```
(condition 1)
&&
(condition 2)
```

uknunknown: at the expense of longer lines I decided not to mix && and || --> let me know if you disagree.

src/transcoding/transcode/stream.c
```diff
@@ -57,70 +57,67 @@ tvh_stream_setup(TVHStream *self, TVHCodecProfile *profile, tvh_ssc_t *ssc)
...
+            } else if (icodec_id == AV_CODEC_ID_VP8) {
+                icodec = avcodec_find_decoder_by_name("vp8_vaapi");
+            }else if (icodec_id == AV_CODEC_ID_MJPEG) {
+                icodec = avcodec_find_decoder_by_name("mjpeg_vaapi");
+            }
         }
-      }
-    }
 #endif
+    }
```
Flole998, line 119: I think you messed up these now, it should be before the` #endif`?

uknunknown: I had an extra bracket in ENABLE_MMAL.

uknunknown: brackets

uknunknown:
I don' t think this is a real problem ... second term doesn't have to be executed when return -1:
problem

Flole998:
Sonarqube just got confused with the additional bracket, and yes the side effects don't matter in this case.

Issue #1749, opened 2024-08-26
[BUG]: 1 new defect(s) introduced to tvheadend/tvheadend found with Coverity Scan


Flole998:
### Tvheadend Version

v4.3 (development)

### Build Number

4.3

### Platform

Other..

### Architecture

Other..

### Category

Other..

### Description

New defect(s) Reported-by: Coverity Scan
Showing 1 of 1 defect(s)

** CID 508927:    (RESOURCE_LEAK)
/src/transcoding/codec/profile_audio_class.c: 110 in tvh_codec_audio_get_list_channel_layouts()
/src/transcoding/codec/profile_audio_class.c: 137 in tvh_codec_audio_get_list_channel_layouts()

________________________________________________________________________________________________________
*** CID 508927:    (RESOURCE_LEAK)
/src/transcoding/codec/profile_audio_class.c: 110 in tvh_codec_audio_get_list_channel_layouts()
104             }
105             else {
106     #if LIBAVCODEC_VERSION_MAJOR > 59
107                 l = channel_layouts;
108                 ADD_ENTRY(list, map, s64, 0, str, AUTO_STR);
109                 while (l->nb_channels < 32) {
>>>     CID 508927:    (RESOURCE_LEAK)
>>>     Overwriting "map" in "map = htsmsg_create_map()" leaks the storage that "map" points to.
110                     if (!(map = htsmsg_create_map())) {
111                         htsmsg_destroy(list);
112                         list = NULL;
113                         break;
114                     }
115                     l_buf[0] = '\0';
/src/transcoding/codec/profile_audio_class.c: 137 in tvh_codec_audio_get_list_channel_layouts()
131                         ADD_ENTRY(list, map, s64, l, str, l_buf);
132                     }
133                 }
134     #endif
135             }
136         }
>>>     CID 508927:    (RESOURCE_LEAK)
>>>     Variable "map" going out of scope leaks the storage it points to.
137         return list;
138     }
139     
140     
141     /* TVHCodecProfile ========================================================== */
142

uknunknown:
@Flole998 if by calling:
` if (!(map = htsmsg_create_map()))`
is leaking memory ... we have many instances where we call this in a 'while' loop (several times). 
Can you confirm this is the problem highlighted?

Flole998:
The issue here is, that if map is non-null the old map is overwritten without being freed on the second iteration.

Flole998:
[...]

PR #1830, opened 2025-05-20
fix memory leak

uknunknown:
My understanding is that htsmsg_create_map() is allocating memory and ADD_ENTRY() is consuming that.
We should always have pairs of htsmsg_create_map() <--> ADD_ENTRY()
If we check the code (before update) - with GREEN is when LIBAVCODEC <= 59 and with RED when LIBAVCODEC > 59:

before

GREEN has clear pairs: 1 <--> 2 and 3 <--> 4
RED has 1 <--> 2 , but 3 <--> 4 is only when 5 == TRUE 
When 5 == FALSE memory is allocated but is not consumed.

The fix proposal is:

after

This will always guarantee (no matter what 5 will be) we have pairs:  1 <--> 2 and 3 <--> 4

Fixes: #1749

### DIFF
--- a/src/transcoding/codec/profile_audio_class.c
+++ b/src/transcoding/codec/profile_audio_class.c
@@ -107,14 +107,14 @@ tvh_codec_audio_get_list_channel_layouts(TVHAudioCodec *self)
             l = channel_layouts;
             ADD_ENTRY(list, map, s64, 0, str, AUTO_STR);
             while (l->nb_channels != 0) {
-                if (!(map = htsmsg_create_map())) {
-                    htsmsg_destroy(list);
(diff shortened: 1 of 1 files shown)


src/htsmsg.c
```diff
@@ -303,6 +303,7 @@ htsmsg_destroy(htsmsg_t *msg)
   memoryinfo_free(&htsmsg_memoryinfo, sizeof(htsmsg_t));
 #endif
   free(msg);
+  msg = NULL;
```
Flole998, line 306: This is the last instruction in the function, setting it to NULL simply does nothing.

PR #1831, opened 2025-05-20
MKV Tags - Change rating label.  Add Sub-title and Comment.

DeltaMikeCharlie:
Change the ‘LAW_RATING’ tag to contain the ‘rl_display_label’ field augmented with either the ‘rl_country’ field or the ‘rl_authority’ field rather than the age.  Eg: ‘PG (AUS)’ or ‘PG (ACMA)’

Add the ‘SUBTITLE’ tag when the appropriate EPG field is available.

Add the ‘COMMENT’ tag when the appropriate DVR entry field is available.

Add some comments to existing code regarding MKV tag values in use.

### DIFF
--- a/src/muxer/muxer_mkv.c
+++ b/src/muxer/muxer_mkv.c
@@ -43,6 +43,7 @@
 #include "parsers/parser_hevc.h"
 #include "muxer_mkv.h"
 
+#include "epggrab.h"  //Needed to see if global processing of parental rating labels is enabled.
 
 extern int dvr_iov_max;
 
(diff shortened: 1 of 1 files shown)


src/muxer/muxer_mkv.c
```diff
@@ -780,14 +784,32 @@ _mk_build_metadata(const dvr_entry_t *de, const epg_broadcast_t *ebc,
...
-  if(ebc) {
-    tmp_age = ebc->age_rating;
-    addtag(q, build_tag_int("LAW_RATING", tmp_age, 0, NULL));
-  }
+  //Only add the 'LAW_RATING' tag if 1) rating label processing is enabled globally
+  //AND 2) this EPG entry has a rating label worth adding.
+  if(epggrab_conf.epgdb_processparentallabels && ebc && ebc->rating_label)
+  {
+    //If the rating label for this event is enabled AND it has a display_label.
+    if(ebc->rating_label->rl_enabled && ebc->rating_label->rl_display_label)
```
Flole998, line 792: Sonarqube is flagging that this `if` could be merged with the enclosing one, what to you think? You could add the condition to a new line.

DeltaMikeCharlie: > what to you think?

It is just a personal preference, I prefer to break up my `if` statements into more manageable, readable, logical chunks.

Outer `if`: Are we processing parental rating labels and is some RL data present?
Inner `if`: Do we meet the exact conditions for being able to write out the tag?

If you would prefer these two steps to be merged into a single `mega-if`, I can do that.  Let me know how you would like to proceed.

PR #1834, opened 2025-05-23
Remove "see Help" text from online help pages.

dave-p:
Some help items contain "see Help for further details". There is no longer a separate help area within the project website so these references should be removed.


dave-p:
The PR is changing the in-app help, which is now the only source of information.

Flole998:
You are removing references to the in-app help despite further information being there at least in one case, as for example in https://github.com/tvheadend/tvheadend/blob/7e1f9caa95bf8d3d3e5c3a5eb687e7acf5b9a916/docs/property/change_parameters.md?plain=1#L4

dave-p:
The text in change_parameters.md continues to be pulled into the Users in-app help page. All I have removed is the text "see Help for further details" because there is no further help available.

Flole998:
That **is** the further help though. The tooltip refers to the help for more explanations in that case. Why do you want to remove that?

PR #1845, opened 2025-06-01
New API Call to determine suitability for entering low-power mode and calculating a reawakening time.

DeltaMikeCharlie:
When TVH is used as part of a DVR appliance configuration, it may be desirable to implement a low-power mode during periods of prolonged inactivity.  Although this information is currently available, it requires multiple API calls and the interpretation of cron strings to obtain.  This new API call simplifies and consolidates the existing process.

This new single API call provides information required for assessing the suitability for entering low-power mode and a suitable time for reawakening.  The following fields are provided:

- Current time
- Earliest scheduled activity (see list below)
- Current number of active subscriptions
- Current number of active inbound TCP connections

For finer analysis and control, individual scheduling times are also provided for the following activities:

- DVR Recordings
- OTA EPG Grabber
- Internal EPG Grabber
- Mux Scheduler

Times are presented as standard Unix epoch values.  A value of zero indicates that nothing for that activity type is currently scheduled.

It should be noted that the times provided are strictly from the point of view of TVH internal activities and that the appliance software must take into account operating system and hardware boot times when determining the precise time to reawaken the appliance.

Times are only reported for the EPG Grabbers if there are enabled grabbers of that type.  Likewise, only enabled Mux Schedulers are reported.

DVR recordings in progress are also included, therefore, the reported DVR time may be in the past.

The connection and subscription counts are duplications of the summaries provided by the ‘status/connections’ and ‘status/subscriptions’ API calls and are included here in order to provide a single consolidated API call for assessing the suitability for entering low-power mode.

**Example:**

http://[TVH_IP]:[TVH_PORT]/api/status/activity

```
{
  "current_time": 1748659324,
  "next_activity": 1748659500,
  "activities": {
    "dvr": 1748764623,
    "ota_grabber": 1748659500,
    "int_grabber": 0,
    "mux_scheduler": 1748725200
  },
  "subscription_count": 1,
  "connection_count": 2
}
```


EricV:
There  has been a discussion long time ago about similar functionality in https://github.com/tvheadend/tvheadend/pull/388.

Basing decision to go to STR on tvheadend info only somehow makes the assumption, tvheadend is the only function that runs on the machine you want to suspend.  Providing the info and exposing it is fine, but the decision to really suspend should be system wide.

DeltaMikeCharlie:
> but the decision to really suspend should be system wide.

Agreed.  This new call just provides more information for making that decision.

src/dvr/dvr_db.c
```diff
@@ -3072,6 +3072,38 @@ dvr_entry_find_by_id(int id)
...
+    if(dvr_entry_is_upcoming(de) && de->de_enabled)
+    {
+      start = dvr_entry_get_start_time(de, 1);
+      if(earliest == 0)
+      {
+        earliest = start;
+      }
+      else
+      {
+        if(start < earliest)
```
Flole998: This should be an `else if`

DeltaMikeCharlie: Thanks for that.  I also spotted another one in the mux scheduler and fixed that too.

PR #1847, opened 2025-06-02
fix memory leak 3 - transcoding

uknunknown:
- fix memory leak 3 - transcoding

### DIFF
--- a/src/transcoding/transcode/hwaccels/vaapi.c
+++ b/src/transcoding/transcode/hwaccels/vaapi.c
@@ -32,7 +32,6 @@ typedef struct tvh_vaapi_context_t {
     int width;
     int height;
     AVBufferRef *hw_device_ref;
-    AVBufferRef *hw_frame_ref;
 } TVHVAContext;
 
 
@@ -52,10 +51,6 @@ tvhva_context_destroy(TVHVAContext *self)
         if (self->hw_device_ref) {
             av_buffer_unref(&self->hw_device_ref);
             self->hw_device_ref = NULL;
-        }
-        if (self->hw_frame_ref) {
-            av_buffer_unref(&self->hw_frame_ref);
-            self->hw_frame_ref = NULL;
         }
         free(self);
         self = NULL;
@@ -631,18 +626,17 @@ vaapi_decode_setup_context(AVCodecContext *avctx)
         tvherror(LS_VAAPI, "Decode: Failed to allocate VAAPI context (TVHVAContext)");
         return AVERROR(ENOMEM);
     }
-
+    // lifted from ffmpeg-6.1.1/doc/examples/vaapi_transcode.c line 237
     /* Open VAAPI device and create an AVHWDeviceContext for it*/
     if ((ret = av_hwdevice_ctx_create(&self->hw_device_ref, AV_HWDEVICE_TYPE_VAAPI, ctx->hw_accel_device, NULL, 0)) < 0) {
         tvherror(LS_VAAPI, "Decode: Failed to Open VAAPI device and create an AVHWDeviceContext for device: "
                             "%s with error code: %s", 
                             ctx->hw_accel_device, av_err2str(ret));
-        // unref self
         free(self);
         self = NULL;
         return ret;
     }
-
+    // lifted from ffmpeg-6.1.1/doc/examples/vaapi_transcode.c line 95
     /* set hw_frames_ctx for decoder's AVCodecContext */
     avctx->hw_device_ctx = av_buffer_ref(self->hw_device_ref);
     if (!avctx->hw_device_ctx) {
@@ -651,7 +645,6 @@ vaapi_decode_setup_context(AVCodecContext *avctx)
         // unref hw_device_ref
         av_buffer_unref(&self->hw_device_ref);
         self->hw_device_ref = NULL;
-        // unref self
         free(self);
         self = NULL;
         return AVERROR(ENOMEM);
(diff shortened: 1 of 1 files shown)


uknunknown:
fixes the Coverity scan: 552898

Flole998:
That can't be right, you are doing the ref right before, assigning it to something different and then do the unref. The unref needs to be done when the pointer is no longer needed

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -651,7 +650,6 @@ vaapi_decode_setup_context(AVCodecContext *avctx)
         // unref hw_device_ref
         av_buffer_unref(&self->hw_device_ref);
```
Flole998, line 646: Looking at this, if the ref failed theres no point in unreffin I think

uknunknown:
before unref:
image

after:
image

I don't think hw_frame_ref pointer is required.

Flole998:
That is intended for automatic memory management, it is required to automatically free that later on but I am not sure if it even works in this case as you are allocating a lot more than just that thing....

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: I think you are right ... but: This is an intriguing case: how was optimized so early?
image

uknunknown: I can remove that ... but there is a risk that will be flagged by Coverity (I am not sure is Coverity will apply the compiler optimization before verifying code integrity).

uknunknown:
After few attempts I decided to propose this version. I remove self from encoding because I don't think is needed. Having access to AVBuffer doesn't bring any benefit because we don't want to tamper with that.  
The self from decoding probably is required for later ... I have to investigate two warnings reported:
2025-06-03 21:08:45.410 [WARNING] libav: AVCodecContext: Driver does not advertise encoder features, using guessed defaults.
2025-06-03 21:08:45.410 [WARNING] libav: AVCodecContext: Driver does not advertise encoder block size, using guessed defaults.
This might require to transfer some information from decoder to encoder .... so self might be used for that.
It would be useful if you could run Coverity before submit :) .

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -764,37 +757,33 @@ int
...
-        free(self);
-        self = NULL;
-        return ret;
+        goto vaapi_encode_setup_context_unref;
     }
-    ctx->hw_device_octx = av_buffer_ref(self->hw_frame_ref);
-    free(self);
-    self = NULL;
+
+    ctx->hw_device_octx = av_buffer_ref(hw_frame_ref);
```
Flole998: You can't call ref/unref on memory that wasn't allocated as av_buffer.

uknunknown: I think is clean now ... let me know your opinion.

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -764,37 +757,29 @@ int
...
-
     /* Open VAAPI device and create an AVHWDeviceContext for it*/
-    if ((ret = av_hwdevice_ctx_create(&self->hw_frame_ref, AV_HWDEVICE_TYPE_VAAPI, NULL, NULL, 0)) < 0) {
+    if ((ret = av_hwdevice_ctx_create(&ctx->hw_device_octx, AV_HWDEVICE_TYPE_VAAPI, NULL, NULL, 0)) < 0) {
         tvherror(LS_VAAPI, "Encode: Failed to open VAAPI device and create an AVHWDeviceContext for it."
                 "Error code: %s",av_err2str(ret));
-        free(self);
-        self = NULL;
-        return ret;
+        goto vaapi_encode_setup_context_exit;
```
Flole998: Goto is just "return ret"? Don't use goto then. Unless the code gets very very very hard to understand without goto, you should never use it.

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -764,37 +757,29 @@ int
...
+    if ((ret = set_hwframe_ctx(avctx, ctx->hw_device_octx)) < 0) {
         tvherror(LS_VAAPI, "Encode: Failed to set hwframe context."
                 "Error code: %s",av_err2str(ret));
-        // unref hw_frame_ref
-        av_buffer_unref(&self->hw_frame_ref);
-        self->hw_frame_ref = NULL;
-        free(self);
-        self = NULL;
-        return ret;
+        av_buffer_unref(&ctx->hw_device_octx);
```
Flole998, line 774: Why unreffing here? This isn't a av_buffer so you can and should not ref/unref it

Flole998: Actually this is an av_buffer, I just need to figure out where its coming from

Flole998: Okay I think the right thing to do here is just set it to NULL after unreffing, and then returning.

src/transcoding/transcode/hwaccels/vaapi.c
```diff
@@ -764,37 +757,29 @@ int
...
+        goto vaapi_encode_setup_context_exit;
     }
-    ctx->hw_device_octx = av_buffer_ref(self->hw_frame_ref);
-    free(self);
-    self = NULL;
+
+    // exit without error
+    ret = 0;
```
Flole998: It's already 0 here, no point in doing this. The whole variable seems useless if its only ever going to be 0.

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: done

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: so you want me to remove all this highlighted regions?
image

src/transcoding/transcode/hwaccels/vaapi.c
uknunknown: ok .. done.

src/transcoding/transcode/hwaccels/vaapi.c
Flole998: I think we need to go over C memory management at some point, otherwise this is just going to be trial and error.

No need to remove those.

uknunknown: Sure ... but I am not sure what I did wrong:
hw_device_octx started with NULL , was 'ref', 'unref' and ended with NULL. In my head this is the perfect cycle (nothing is left 'undone').

Flole998: It's not a pointer though, you are creating a pointer just to unref it. That does nothing in the best case and tries to free something that was never allocated in the worst case.

I'm not sure why there's so much code here doing what seems to be wrong.

Flole998: Maybe I'm wrong, I need to have a look what's actually allocated where and how

PR #1848, opened 2025-06-03
Coverity CID 552897

DeltaMikeCharlie:
Resolution for Coverity CID 552897.

- Encase mux scheduler read operation in a mutex lock/unlock transaction.

### DIFF
--- a/src/epggrab.c
+++ b/src/epggrab.c
@@ -516,17 +516,21 @@ time_t epggrab_get_next_int(void)
 {
   time_t ret_time;
   struct timespec current_time;
-  
+
   clock_gettime(CLOCK_REALTIME, &current_time);
-  if(!cron_multi_next(epggrab_cron_multi, current_time.tv_sec, &ret_time))
+
+  tvh_mutex_lock(&epggrab_mutex);
+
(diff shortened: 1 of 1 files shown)


src/epggrab.c
```diff
@@ -514,19 +514,32 @@ const idclass_t epggrab_class = {
...
 {
-  time_t ret_time;
+  time_t ret_time = 0;
   struct timespec current_time;
+  int mutex_ret = 0;
   
   clock_gettime(CLOCK_REALTIME, &current_time);
-  if(!cron_multi_next(epggrab_cron_multi, current_time.tv_sec, &ret_time))
+
+  mutex_ret = tvh_mutex_lock(&epggrab_mutex);  //A return code of zero means success.
```
Flole998: We are never checking the return code, so we shouldn't do it here either. If you really want to, it's probably a good idea to add the check to the tvh_mutex_lock macro, but it shouldn't ever fail.

DeltaMikeCharlie: The return code test has been removed.

Flole998:
I removed the initialization and the spaces in the new line, this should be ready for merging now (didn't want to bother you for those 2 small changes).

PR #1852, opened 2025-06-06
ci: disable coverity on forks

chewitt:
Add an `if` clause so coverity only runs on our main repo and not personal forks. This should stop the weekly email telling me the action on my fork failed.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   coverity:
+    if: ${{ secrets.COVERITY_SCAN_TOKEN != '' }}
     runs-on: ubuntu-latest
     environment: Coverity
     steps:


Flole998:
I suggest checking if ${{ secrets.COVERITY_SCAN_TOKEN }} is set instead, then people can run it if they wish and it will only run if configured properly

chewitt:
@Flole998 I've switched to checking the secret, please check the logic/cleanup done.

.github/workflows/coverity.yml
```diff
@@ -8,8 +8,13 @@ on:
     - cron: "0 7 * * 1"
   workflow_dispatch:
 
+env:
```
Flole998: Is there any good reason to do this? The current way ensures the credentials are only available to the underlying code when they are actually needed. Especially they are not available when compiling code that users frequently touch (our main codebase)

PR #1853, opened 2025-06-06
remove coded_width and coded_height from encoding

uknunknown:
according to AVCodecContext documentation this is only used for decoding, oavctx is used for encoding

image

### DIFF
--- a/src/transcoding/transcode/video.c
+++ b/src/transcoding/transcode/video.c
@@ -224,8 +224,6 @@ tvh_video_context_open_encoder(TVHContext *self, AVDictionary **opts)
     }
 
 #if ENABLE_HWACCELS
-    self->oavctx->coded_width = self->oavctx->width;
-    self->oavctx->coded_height = self->oavctx->height;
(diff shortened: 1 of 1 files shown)


woooow1:
In my configuration this PR breaks output resolution. For example: without PR- downscale 1920x1080 to 1280x720, with this PR tvheadend downscale 1920x1080 to 900x720 with strange aspect ratio. More investigation is needed!

uknunknown:
@woooow1 can you provide a sample (1 min) of that mux and tell me what channel is showing that behavior? I tried to replicate this problem but channels in my area are not showing this problem. https://docs.tvheadend.org/documentation/creating-a-mux-dump.
Also please clarify what codec you use for encoder (I don't know if it matters).

woooow1:
@uknunknown  Today everything works as expected. I don`t know... maybe I patch video.c incorrectly... Sorry guys, next time I will double check everything before write. Here is my settings:

uknunknown:
no problem @woooow1 ... better safe than sorry.

Issue #1855, opened 2025-06-09
[BUG]: User config doesn't force DVR profile


DeltaMikeCharlie:
### Tvheadend Version

v4.3 (development)

### Build Number

0.0.0-unknown

### Platform

Ubuntu 22.04 (Jammy)

### Architecture

Intel x86_64

### Category

API

### Description

Transfer of issue first raised on the old forum on 23 July 2020 by Luis Alves.

> I have a user configuration entry that has a specific "DVR configuration profile".
> 
> When that user creates recordings via Kodi, the DVR configuration is set to
> "(default config)" instead of the one ticked in the user config.

https://old.tvheadend.org/issues/5930

This issue is referred to in PR #1443.

PR #1856, opened 2025-06-09
User's DVR Configuration profile not used when scheduling recordings via HTSP

DeltaMikeCharlie:
When scheduling a recording via HTSP (Kodi client), the '(Default profile)' is used instead of the user's nominated DVR Configuration profile.  However, when the same operation with the same user is performed via the WebUI, the correct profile was used.

Align the HTSP DVR Configuration profile selection process with the process used in the equivalent JSON API function [dvr_config_find_by_list()].

https://github.com/tvheadend/tvheadend/blob/42ed6affc42ef06e427fb8b1465532a4489e182c/src/api/api_dvr.c#L145C1-L147C64

Tested by scheduling a recording from Kodi with the same user, with and without a nominated DVR profile.  Please note, the HTSP connection retains its previous settings until reauthenticated.

Fixes: #1855
Supersedes: #1443

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -2042,13 +2042,12 @@ htsp_method_addDvrEntry(htsp_connection_t *htsp, htsmsg_t *in)
   conf = htsmsg_create_map();
   htsmsg_copy_field(conf, "enabled", in, NULL);
   s = htsmsg_get_str(in, "configName");
-  if (s) {
-    dvr_conf = dvr_config_find_by_uuid(s);
-    if (dvr_conf == NULL)
-      dvr_conf = dvr_config_find_by_name(s);
+  dvr_conf = dvr_config_find_by_list(htsp->htsp_granted_access->aa_dvrcfgs, s);
     if (dvr_conf)
+  {
       htsmsg_add_uuid(conf, "config_name", &dvr_conf->dvr_id.in_uuid);
   }
+  
   htsmsg_copy_field(conf, "start_extra", in, "startExtra");
   htsmsg_copy_field(conf, "stop_extra", in, "stopExtra");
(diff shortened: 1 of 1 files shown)


Flole998:
Thank you!!!!

I believe this is the reason why it worked on 4.2 but not on 4.3?

This is done at the creation time, right? So if I force a profile for a user, the user creates a recording and I change my mind and force a different profile, the recording will still use the old profile?

DeltaMikeCharlie:
> Thank you!!!!

You are welcome.

> This is done at the creation time, right?

Yes, the profile is attached when the DVR entry is created.

> So if I force a profile for a user, the user creates a recording and I change my mind and force a different profile, the recording will still use the old profile?

I did not test that exact scenario, I suspect that you are correct.  Maybe there is some last minute check before the recording starts or something like that, I did not look.

Also note that the permissions remain valid for the duration of the htsp session regardless of the changes made.  To force the changes, the htsp session needs to be disconnected and re-authenticated.  This seems to be a global feature and nothing specifically to do with this particular issue.

Flole998:
Awesome!

Btw: If you just write "Fixes: <issuenumber>" it will auto close the issue when merging. I changed that

DeltaMikeCharlie:
> If you just write "Fixes: " it will auto close the issue when merging. I changed that

Thanks.  I thought that 'Closes' worked too, perhaps it failed because I had 'Closes Issue' instead of just 'Closes'.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

Flole998:
Yes that was likely the reason. I thought the colon is also important but I guess according to the docs it isn't.

PR #1858, opened 2025-06-12
remove deprecated ticks_per_frame

uknunknown:
- remove deprecated ticks_per_frame

image


Flole998:
You can't just remove it, you need to use the suggested alternative. Removing something without properly replacing it is almost certainly never going to work.

uknunknown:
Recommended is to use AVCodecContext.framerate ... this is what we use in the line highlighted.
image
That line was already in the code ... 
I tested several vaapi encoders and software encoders.
Let me know if you still have doubts on the proposal.

jahutchi:
Note: I also have #1859: which adjusts ticks_per_frame when field-rate de-interlacing is configured, I left this in my PR for the time being to preserve compatibility, but I’m aware it’s deprecated. If the deprecation/removal in your PR goes ahead, I’m happy to drop this line entirely from my PR. I've therefore added myself as a follower on this PR.

uknunknown:
my recommendation is for us to wait until the PR-s will pass or will be rejected - to avoid conflicts.

heitbaum:
- ffmpeg-8.0 has dropped ticks_per_frame #1902

Flole998:
Lets improve the comments here and rebase to resolve the conflict, then it can be merged :)

uknunknown:
I will try my best to fix the conflicts (I have some instructions from DMC).

uknunknown:
@jahutchi With the changes from this PR (after rebase) I am able to repro the problem highlighted in the thread: https://tvheadend.org/d/9202-hardware-transcoding-stopped-working/22
Can you clone this branch and try to understand why is no longer working?

uknunknown:
We need to discuss if this is a good practice, but I want to highlight that in general we don't have \n in our logs :( . It generates a strange result overall.
request

jahutchi:
> @jahutchi With the changes from this PR (after rebase) I am able to repro the problem highlighted in the thread: https://tvheadend.org/d/9202-hardware-transcoding-stopped-working/22 Can you clone this branch and try to understand why is no longer working?

No problems - happy to dig into this when time allows but might need to give me a couple of days. Until then please don't merge into master as it looks like the changes in the 3rd commit undo some of the changes that were put in to support field (double frame-rate) Deinterlacing 😕https://github.com/tvheadend/tvheadend/pull/1858/commits/a87135b0ce82caa4b71ad2f5805ea5e3a8ad454b

In the meantime could someone clarify what versions of ffmpeg we currently support? As I do wonder if we need to retain setting ticks-per-frame on oavctx (maybe using ifdef) for older ffmpeg versions. I'm happy to draft an alternative approach along these lines when I dig further.

jahutchi:
> We need to discuss if this is a good practice, but I want to highlight that in general we don't have \n in our logs :( . It generates a strange result overall. request

I used ChatGPT to do a 'final' review and IIRC one of it's suggestions was to adjust the logging in this way to make it easier for anyone reviewing a debug log to pick out the encoder configuration. I originally had it all on one long log line, which I must say did look cluttered. But I'm 100% happy to submit a PR to put it back that way if I've broken some coding/logging standards (couldn't find any at the time of putting together the PR).

uknunknown:
> In the meantime could someone clarify what versions of ffmpeg we currently support? As I do wonder if we need to retain setting ticks-per-frame on oavctx (maybe using ifdef) for older ffmpeg versions.

Thank you for pointing that. According to Google: AVCodecContext.framerate was introduce in ffmpeg v.3 --> should we bother to do an ifdef for "before ffmpeg 3"? I am tempted to say is 'not needed' ... anyone has a different opinion?

jahutchi:
> > In the meantime could someone clarify what versions of ffmpeg we currently support? As I do wonder if we need to retain setting ticks-per-frame on oavctx (maybe using ifdef) for older ffmpeg versions.
> 
> Thank you for pointing that. According to Google: AVCodecContext.framerate was introduce in ffmpeg v.3 --> should we bother to do an ifdef for "before ffmpeg 3"? I am tempted to say is 'not needed' ... anyone has a different opinion?

Maybe I was worrying unduly on whether not setting ticks_per_frame on the AVCodecContext would lead to any unwanted consequences for those still building with FFMpeg version <6.1 (the point at which it actually became deprecated). 

So my concern was... if a user is still on FFmpeg 6.0 (for example) and we are no longer setting AVCodecContext->ticks_per_frame then would this user experience any different behaviour; since it looks like a [fair amount of code was adjusted](https://ffmpeg.org/pipermail/ffmpeg-cvslog/2023-May/137274.html?utm_source=chatgpt.com) when it became deprecated.

I don't readily have any FFmpeg <6.1 environments to test on; and with those FFMpeg commits being too large to trawl over I asked AI to do a deeper dive into this subject which maybe now puts my mind at rest:

-----------------------

ChatGPT-5 
Short answer: if you set AVCodecContext.framerate (added in FFmpeg 3.0), you don’t need to touch ticks_per_frame on any FFmpeg ≤ 6.0. From 6.1 it’s deprecated anyway.

Here’s what the code + docs say:

**Deprecation:** AVCodecContext.ticks_per_frame was deprecated in FFmpeg 6.1 as “redundant for encoding”; the replacement is to use AVCodecContext.framerate.

**How encoders behave if you don’t set it (<6.1):**
Encoders typically take FPS from framerate if present, and only fall back to time_base×ticks_per_frame if framerate is not set. For example the x264 wrapper does:

if avctx->framerate is set → use it directly;

else → fps_den = time_base.num * ticks_per_frame.
So when framerate is set, ticks_per_frame is ignored.

**Other places it appears:** Some internal code references the field (e.g., ratecontrol), but the modern, supported path is to specify framerate for encoders and a sensible time_base.

**Practical recommendation for your codebase**

FFmpeg 3.0–6.0: Set avctx->framerate and avctx->time_base = av_inv_q(avctx->framerate). Do not set ticks_per_frame; behavior will be identical for mainstream encoders (e.g., libx264/libx265).

FFmpeg < 3.0 (only if you still support it): There is no framerate field. Either:

set time_base = 1/fps and leave ticks_per_frame = 1, or

if you keep a half-frame time_base, then set ticks_per_frame = 2.

FFmpeg ≥ 6.1: ticks_per_frame is deprecated—don’t set it; rely on framerate.

Bottom line for your debate

If your supported matrix is FFmpeg 3.0+, it’s safe to delete any oavctx->ticks_per_frame writes (and skip an #ifdef). Just ensure you always fill framerate and use a frame-duration time_base. The only time you’d still need an #ifdef is if you truly support pre-3.0 builds.


👍

Flole998:
> So my concern was... if a user is still on FFmpeg 6.0 (for example) and we are no longer setting AVCodecContext->ticks_per_frame then would this user experience any different behaviour; since it looks like a [fair amount of code was adjusted](https://ffmpeg.org/pipermail/ffmpeg-cvslog/2023-May/137274.html?utm_source=chatgpt.com) when it became deprecated.

Thats also what I was thinking

jahutchi:
> > @jahutchi With the changes from this PR (after rebase) I am able to repro the problem highlighted in the thread: https://tvheadend.org/d/9202-hardware-transcoding-stopped-working/22 Can you clone this branch and try to understand why is no longer working?
> 

### Summary of Commit a87135b0

That commit did more than just remove `ticks_per_frame`:

- It removed the field assignment **and also changed how** `framerate`, `time_base`, and `gop_size` are calculated.  
- It dropped the **90 kHz timebase alignment** and the **field_rate multiplier** (for deinterlaced streams).  
- GOP size calculation semantics were altered.

### Potential Issues

- Without the 90 kHz base, MPEG-TS muxed output **freezes or drifts** in some players (confirmed in testing).  
- Dropping `field_rate` breaks correct pacing when deinterlacing doubles the framerate.  
- GOP size logic is no longer equivalent to the original.  

So the commit is over-broad: it doesn’t just remove a deprecated field, it changes encoder timing behaviour.

### Safer Alternative

We only need to **remove the write to the deprecated `AVCodecContext.ticks_per_frame`** while keeping all the existing maths intact (90 kHz, field_rate, GOP length).  
That way there’s no behavioural change.

### Minimal Patch

Note: this is completely untested

```diff
--- a/src/transcoding/transcode/video.c
+++ b/src/transcoding/transcode/video.c
@@ -269,11 +269,14 @@ tvh_video_context_open_encoder(TVHContext *self, AVDictionary **opts)
     if (!self->iavctx->framerate.num) {
         self->iavctx->framerate = av_make_q(30, 1);
     }

-    self->oavctx->framerate = av_mul_q(self->iavctx->framerate, (AVRational){ field_rate, 1 });
-    self->oavctx->ticks_per_frame = (90000 * self->oavctx->framerate.den) / self->oavctx->framerate.num;
-    ticks_per_frame = av_make_q(self->oavctx->ticks_per_frame, 1);
-    self->oavctx->time_base = av_inv_q(av_mul_q(self->oavctx->framerate, ticks_per_frame));
-    self->oavctx->gop_size = ceil(av_q2d(av_inv_q(av_mul_q(self->oavctx->time_base, ticks_per_frame))));
+    self->oavctx->framerate = av_mul_q(self->iavctx->framerate, (AVRational){ field_rate, 1 });
+
+    /* compute ticks_per_frame locally instead of writing to deprecated field */
+    int ticks_tmp = (90000 * self->oavctx->framerate.den) / self->oavctx->framerate.num;
+    AVRational ticks_per_frame = av_make_q(ticks_tmp, 1);
+
+    self->oavctx->time_base = av_inv_q(av_mul_q(self->oavctx->framerate, ticks_per_frame));
+    self->oavctx->gop_size  = ceil(av_q2d(av_inv_q(av_mul_q(self->oavctx->time_base, ticks_per_frame))));

uknunknown:
@jahutchi Thank you for your review. I decided to select the minimal changes (two lines: replace `self->oavctx->ticks_per_frame` with `ticks_per_frame_tmp`). I think you proposal is very good.

jahutchi:
> We need to discuss if this is a good practice, but I want to highlight that in general we don't have \n in our logs :( . It generates a strange result overall. request

https://github.com/tvheadend/tvheadend/pull/1912 created to address this.

Flole998:
This is ready for merging, right?

uknunknown:
Yes ... is ready. Minimal changes that will fix the problem.

uknunknown:
I don't know how to resolve this conflict ... I will close it in one month or so.

PR #1865, opened 2025-07-04
linuxdvb_adapter: increase MAX_DEV_OPEN_ATTEMPTS to 50

smp79:
This fixes 2 issues with my TBS 5520SE and TBS 5930 USB devices. 
Issue 1: no adapters present in tvh after a system reboot. Tvh required a delayed start (wait for frontend init).
Issue 2: no adapters present in tvh after physically disconnecting/reconnecting the adapter. Tvh required a restart.

### DIFF
--- a/src/input/mpegts/linuxdvb/linuxdvb_adapter.c
+++ b/src/input/mpegts/linuxdvb/linuxdvb_adapter.c
@@ -43,7 +43,7 @@
 #define CI_PATH  "%s/ci%d"
 #define SEC_PATH "%s/sec%d"
 
-#define MAX_DEV_OPEN_ATTEMPTS 20
+#define MAX_DEV_OPEN_ATTEMPTS 50
 
 /* ***************************************************************************
  * DVB Adapter


DeltaMikeCharlie:
> Issue 1: no adapters present in tvh after a system reboot. Tvh required a delayed start.
> Issue 2: no adapters present in tvh after physically disconnecting/reconnecting the adapter. Tvh required a restart.

I'm wondering if these issues are common to all USB devices or just the particular make/model that you have.

If the issues are device-specific, rather than making a global change, perhaps a new command line argument specifying the number of retries would be a better solution.  It would not impact existing users and users with this issue could tweak the number of retries to their liking.  Using a command line argument would also allow for specifying a number higher than 50 if needed.

smp79:
> I'm wondering if these issues are common to all USB devices

It's driver-specific. Tvh seem to have trouble with some "slower" devices. TBS 5520SE is a dual tuner device. If I hack the driver and disable the terrestrial tuner - there is no issue. With 2 enabled tuners it runs into trouble. TBS 5930 is single tuner but also have this issue. I have a bunch of other USB devices and they work fine with the default 20.

Not really sure about the command line option. Issues with those TBS devices are an edge case, a higher global option should be ok I think. It does not affect the "fast" devices in any way so there is no regression for existing users.

PR #1871, opened 2025-07-10
Implement age ratings on XMLTV export

### LINKED ISSUE #1794: [BUG]: XMLTV Output does not display age ratings
### Tvheadend Version

v4.3 (development)

### Build Number

4.3-2155~gfe4df311d

### Platform

Debian 12 (Bookworm)

### Architecture

ARMv8 (aarch64)

### Category

EPG

### Description

Hello. I noticed that XMLTV output (http://tvh:port/xmltv/channels) does not include age ratings like the main WebUI screen does. 

I think this is the function that puts the data into the XML
https://github.com/tvheadend/tvheadend/blob/653bd0400b4413db96b80c807f0f7524f9248adb/src/webui/xmltv.c#L242-L244

And epg_broadcast_t has age rating
https://github.com/tvheadend/tvheadend/blob/653bd0400b4413db96b80c807f0f7524f9248adb/src/epg.h#L280

So in theory it just needs
```
<rating>
     <value>{ebc->age_rating}</value>
</rating>
```
*Not to be confused with star-rating, this is age rating*

Thank you

### Logfile

_No response_

### Reproduction

_No response_

edisionnano:
Properly closes #1794 
I took the ideas from the first comment by @DeltaMikeCharlie for the fallbacks.
By default it should output a block that looks like this
```
    <rating system="VCHIP">
      <value>TV-G</value>
    </rating>
```
If the rl_authority is not present it will fallback to rl_country which is the ISO country code (gbr for example).
If the rl_display_label is not present it will fallback rl_display_age and then to rl_age. 
If process is disabled on epg grabber settings it will fallback to age_rating and use a hardcoded "AGE" as the system.

### DIFF
--- a/src/webui/xmltv.c
+++ b/src/webui/xmltv.c
@@ -23,6 +23,7 @@
 #include "http.h"
 #include "string_list.h"
 #include "imagecache.h"
+#include "epggrab.h"
 
 #define XMLTV_FLAG_LCN		(1<<0)
 
@@ -184,6 +185,8 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
   lang_str_ele_t *lse;
   epg_genre_t *genre;
   char buf[64];
+  char prop_buf[512];
+  const char *str;
 
   if (ebc->subtitle)
     RB_FOREACH(lse, ebc->subtitle, link) {
@@ -222,6 +225,54 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
     }
     htsbuf_append_str(hq, "  </credits>\n");
   }
+
+  if (epggrab_conf.epgdb_processparentallabels && ebc->rating_label) {
+      const char *system = ebc->rating_label->rl_authority;
+      if (!system) system = ebc->rating_label->rl_country;
+      if (system && (ebc->rating_label->rl_display_label || ebc->rating_label->rl_display_age || ebc->rating_label->rl_age)) {
+          htsbuf_append_str(hq, "  <rating system=\"");
+          htsbuf_append_and_escape_xml(hq, system);
+          htsbuf_append_str(hq, "\">\n");
+          if (ebc->rating_label->rl_display_label) {
+              htsbuf_append_str(hq, "    <value>");
+              htsbuf_append_and_escape_xml(hq, ebc->rating_label->rl_display_label);
+              htsbuf_append_str(hq, "</value>\n");
+          } else if (ebc->rating_label->rl_display_age) {
+              htsbuf_append_str(hq, "    <value>");
+              htsbuf_qprintf(hq, "%d", ebc->rating_label->rl_display_age);
+              htsbuf_append_str(hq, "</value>\n");
+          } else if (ebc->rating_label->rl_age) {
+              htsbuf_append_str(hq, "    <value>");
+              htsbuf_qprintf(hq, "%d", ebc->rating_label->rl_age);
+              htsbuf_append_str(hq, "</value>\n");
+          }
+          if (ebc->rating_label->rl_icon) {
+              str = ebc->rating_label->rl_icon;
+              if (!strempty(str)) {
+                  str = imagecache_get_propstr(str, prop_buf, sizeof(prop_buf));
+                  if (str) {
+                      htsbuf_append_str(hq, "    <icon src=\"");
+                      htsbuf_append_and_escape_xml(hq, str);
+                      htsbuf_append_str(hq, "\"/>\n");
+                  }
+              }
+          }
+          htsbuf_append_str(hq, "  </rating>\n");
+      } else if (ebc->age_rating) {
+        htsbuf_append_str(hq, "  <rating system=\"AGE\">\n");
+        htsbuf_append_str(hq, "    <value>");
+        htsbuf_qprintf(hq, "%d", ebc->age_rating);
+        htsbuf_append_str(hq, "</value>\n");
+        htsbuf_append_str(hq, "  </rating>\n");
+      }
+  } else if (ebc->age_rating) {
+      htsbuf_append_str(hq, "  <rating system=\"AGE\">\n");
+      htsbuf_append_str(hq, "    <value>");
+      htsbuf_qprintf(hq, "%d", ebc->age_rating);
+      htsbuf_append_str(hq, "</value>\n");
+      htsbuf_append_str(hq, "  </rating>\n");
+  }
+
   _http_xmltv_programme_write_string_list(hq, ebc->category, "category");
   LIST_FOREACH(genre, &ebc->genre, link) {
     if (genre && genre->code) {


DeltaMikeCharlie:
@edisionnano - Thanks for doing this, it has been on my 'to do' list for some time.

In addition to the AI findings, I'd like to recommend a slight tweak in your logic:  Evaluate the global Parental Rating flag and the EPG entry's RL pointer `epggrab_conf.epgdb_processparentallabels && ebc->rating_label` instead of just `ebc->rating_label`.  If that is false, fallback to `ebc->age_rating`.

Only if the global flag is true export the values from the rating labels.  If rating label processing was previously enabled, but recently disabled, then there may be some residual rating label pointers in the EPG data.  Evaluating `ebc->rating_label` alone may give a false positive.

Also, export `ebc->rating_label->rl_display_age` rather than just `ebc->rating_label->rl_age `.  The `rl_age` is the raw value received from the EIT and the `rl_display_age` is the processed age because 3 needs to be added to the raw EIT value to get the true age.

https://github.com/tvheadend/tvheadend/blob/d431956cc62f80115fa17103374f7c921081d72c/src/ratinglabels.h#L26C1-L43C17

https://github.com/tvheadend/tvheadend/blob/d431956cc62f80115fa17103374f7c921081d72c/src/epggrab/module/eit.c#L439C1-L442C48

edisionnano:
> @edisionnano - Thanks for doing this, it has been on my 'to do' list for some time.
> 
> In addition to the AI findings, I'd like to recommend a slight tweak in your logic: Evaluate the global Parental Rating flag and the EPG entry's RL pointer `epggrab_conf.epgdb_processparentallabels && ebc->rating_label` instead of just `ebc->rating_label`. If that is false, fallback to `ebc->age_rating`.
> 
> Only if the global flag is true export the values from the rating labels. If rating label processing was previously enabled, but recently disabled, then there may be some residual rating label pointers in the EPG data. Evaluating `ebc->rating_label` alone may give a false positive.
> 
> Also, export `ebc->rating_label->rl_display_age` rather than just `ebc->rating_label->rl_age `. The `rl_age` is the raw value received from the EIT and the `rl_display_age` is the processed age because 3 needs to be added to the raw EIT value to get the true age.
> 
> https://github.com/tvheadend/tvheadend/blob/d431956cc62f80115fa17103374f7c921081d72c/src/ratinglabels.h#L26C1-L43C17
> 
> https://github.com/tvheadend/tvheadend/blob/d431956cc62f80115fa17103374f7c921081d72c/src/epggrab/module/eit.c#L439C1-L442C48

Done

src/webui/xmltv.c
```diff
@@ -222,6 +223,48 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
...
+          }
+          htsbuf_append_str(hq, "  </rating>\n");
+      } else if (ebc->age_rating) {
+        htsbuf_append_str(hq, "  <rating system=\"AGE\">\n");
+        htsbuf_append_str(hq, "    <value>");
+        htsbuf_qprintf(hq, "%d", ebc->age_rating);
+        htsbuf_append_str(hq, "</value>\n");
+        htsbuf_append_str(hq, "  </rating>\n");
+      }
+  } else if (ebc->age_rating) {
```
Flole998, line 268: Looks like this is the same as the code block above this? If yes, please just add it outside the if-block. Or maybe this is not intended?

edisionnano: Yeah it's the same code. The idea is that if process parental control labels is enabled and ebc->rating_label is not null but rl_authority and rl_country is, it should fall back to rl_age but if process parental control labels is disabled or ebc->rating_label is null it should also fall back to rl_age thus the repetition

src/webui/xmltv.c
```diff
@@ -222,6 +223,48 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
     }
     htsbuf_append_str(hq, "  </credits>\n");
   }
+
+  if (epggrab_conf.epgdb_processparentallabels && ebc->rating_label) {
+      const char *system = ebc->rating_label->rl_authority;
+      if (!system) system = ebc->rating_label->rl_country;
+      if (system) {
```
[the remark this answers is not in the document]
edisionnano: Can rating_label even exist without rl_age?

src/webui/xmltv.c
Flole998: Yeah makes sense

src/webui/xmltv.c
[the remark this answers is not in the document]
Flole998: @DeltaMikeCharlie can it?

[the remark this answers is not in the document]
DeltaMikeCharlie: > @DeltaMikeCharlie can it?

Yes, it can be zero.

1) The `rating_label` would have to be auto-generated from an XMLTV EPG source and the user would have to not enter a value.
2) If the source is EPG EIT, then only if the broadcaster sets the age to zero which the standard states is 'undefined'.  This can be used in some regions to indicate 'Not yet classified'.

Note: `rl_age` is used for mapping to EIT ratings, `rl_display_age` is what should be output and that can be zero.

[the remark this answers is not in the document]
edisionnano: > > @DeltaMikeCharlie can it?
> 
> Yes, it can be zero.
> 
>     1. The `rating_label` would have to be auto-generated from an XMLTV EPG source and the user would have to not enter a value.
> 
>     2. If the source is EPG EIT, then only if the broadcaster sets the age to zero which the standard states is 'undefined'.  This can be used in some regions to indicate 'Not yet classified'.
> 
> 
> Note: `rl_age` is used for mapping to EIT ratings, `rl_display_age` is what should be output and that can be zero.

I see, I accepted the change

Flole998 (approved):
Code is fine for me, this just needs to be squashed

src/webui/xmltv.c
```diff
@@ -222,6 +223,48 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
...
+              htsbuf_qprintf(hq, "%d", ebc->rating_label->rl_display_age);
+              htsbuf_append_str(hq, "</value>\n");
+          } else if (ebc->rating_label->rl_age) {
+              htsbuf_append_str(hq, "    <value>");
+              htsbuf_qprintf(hq, "%d", ebc->rating_label->rl_age);
+              htsbuf_append_str(hq, "</value>\n");
+          }
+          if (ebc->rating_label->rl_icon) {
+              htsbuf_append_str(hq, "    <icon src=\"");
+              htsbuf_append_and_escape_xml(hq, ebc->rating_label->rl_icon);
```
DeltaMikeCharlie: Sorry for not seeing this earlier.  If you are exporting an image that TVH is expected to serve, then it needs to come from the image cache.

Normally, the icon path should be something like `file:///local/path/to/icon/images/icon.png` but to be accessed externally, it needs to be something like `imagecache/n` where `n` is the index within the cache.  The client will then need to prepend the TVH server details, eg. `http://tvh_ip:tvh_port/imagecache/1`

Have a look at the htsp server and EPG API for an examples of this.

https://github.com/tvheadend/tvheadend/blob/d1fb6da0a3a4583d9bb125276bb680ab7b2b87fe/src/htsp_server.c#L1025C1-L1033C31

https://github.com/tvheadend/tvheadend/blob/d1fb6da0a3a4583d9bb125276bb680ab7b2b87fe/src/api/api_epg.c#L198C1-L205C45

Also have a look at the sample output from `api/epg/events/grid` https://docs.tvheadend.org/documentation/development/json-api/api-description/epg#epg-events-grid

DeltaMikeCharlie:
@edisionnano - Also, can you please post a copy of the rating labels that your system has along with a note about your location and EPG sources.  I'd like to review the values that you have and add them to the global database that I am compiling.

edisionnano:
> @edisionnano - Also, can you please post a copy of the rating labels that your system has along with a note about your location and EPG sources. I'd like to review the values that you have and add them to the global database that I am compiling.

Yeah so about that, when process is enabled it shows like that `DVB:gbr:13` although I'm in Greece not Great Britain. Without process it works fine. My source only is DVB-T for Greece and Cyprus.
The labels are K (for everyone) and then 8, 12, 16, 18 which is the age you need to be to watch this specific program. When it's K there is no rating_label or age_rating however they do include `[K]` in the program title.
In Cyprus CYBC (the state owned channels) do no have a rating in their EPG, Velister which has the privately owned channels does have rating in their EPG.
The rating labels there are:
K, 12, 15, 18, A
Same as Greece except for `A` which is only for pay per view porn channels and you won't find it on DVB-T. The only difference is that when there's `K` the age_rating is 4
Velister won't be around for long as Hellas Sat will take over to complete the transition to DVB-T2 in Cyprus beginning 2026.

DeltaMikeCharlie:
> Yeah so about that, when process is enabled it shows like that `DVB:gbr:13` although I'm in Greece not Great Britain.

Perhaps your broadcasters include ratings from multiple countries and `gbr` is the first one encountered.  The DVB specification allows for that.  I was concerned that this may happen when I was writing the module, but I could not find an multi-country example to test with.  If this is the case, I could change the module so that it does not stop at the first rating in the EPG, it will process all that it receives but only use the first active RL.  This way, the system would still learn `gbr`, `grc` ad `cyp`, but `gbr` can be set as disabled and ignored.

For `DVB:gbr:13` you can change the display_label to `16` and you could also add an icon.
New_Greek_TV_rating_system_-_K16 svg

Are you able to provide a screenshot of your Rating Labels page from the TVH WebUI?

Would you be willing to create a mux dump so that I can analyse what your broadcasters are sending?

https://docs.tvheadend.org/documentation/creating-a-mux-dump

The files can be sent via https://filetransfer.io/

If you are interested, rather than using GitHub comments, you could join the forum (https://tvheadend.org/) and we can continue the conversation there.

edisionnano:
> > Yeah so about that, when process is enabled it shows like that `DVB:gbr:13` although I'm in Greece not Great Britain.
> 
> Perhaps your broadcasters include ratings from multiple countries and `gbr` is the first one encountered. The DVB specification allows for that. I was concerned that this may happen when I was writing the module, but I could not find an multi-country example to test with. If this is the case, I could change the module so that it does not stop at the first rating in the EPG, it will process all that it receives but only use the first active RL. This way, the system would still learn `gbr`, `grc` ad `cyp`, but `gbr` can be set as disabled and ignored.
> 
> For `DVB:gbr:13` you can change the display_label to `16` and you could also add an icon. New_Greek_TV_rating_system_-_K16 svg
> 
> Are you able to provide a screenshot of your Rating Labels page from the TVH WebUI?
> 
> Would you be willing to create a mux dump so that I can analyse what your broadcasters are sending?
> 
> https://docs.tvheadend.org/documentation/creating-a-mux-dump
> 
> The files can be sent via https://filetransfer.io/
> 
> If you are interested, rather than using GitHub comments, you could join the forum (https://tvheadend.org/) and we can continue the conversation there.

How can I check if it includes data for multiple rating systems?
image
https://files.fm/u/vppwvrbtfa
Here's a transponder dump from Greece, I took this one using dvbjet. Digea is the bouquet with the privately owned channels and ERT the state ones. I can get one from Cyprus if needed too. For the icons I have vectors of the Greek ones and good quality raster for the Cypriot ones.
Yes I will join the forum, which thread should I post on?

DeltaMikeCharlie:
> Here's a transponder dump from Greece

Many thanks.

> How can I check if it includes data for multiple rating systems?

You can use a tool like '[DVB Inspector](https://github.com/EricBerendsen/dvbinspector)'.  However, I have had a quick look using a hex dump and you appear to have 2 ratings.

image

All of the ones that I looked at have the same age value.

> I can get one from Cyprus if needed too.

Yes please, that would be very helpful.

> Yes I will join the forum, which thread should I post on?

'Development' would be the best place.  I can always move it later if needed.

DeltaMikeCharlie:
@edisionnano - Try disabling the `gbr` rating labels and then run the EPG scan again, I suspect that the `gre` ones may turn up.

edisionnano:
> > Here's a transponder dump from Greece
> 
> Many thanks.
> 
> > How can I check if it includes data for multiple rating systems?
> 
> You can use a tool like '[DVB Inspector](https://github.com/EricBerendsen/dvbinspector)'. However, I have had a quick look using a hex dump and you appear to have 2 ratings.
> image
> 
> All of the ones that I looked at have the same age value.
> 
> > I can get one from Cyprus if needed too.
> 
> Yes please, that would be very helpful.
> 
> > Yes I will join the forum, which thread should I post on?
> 
> 'Development' would be the best place. I can always move it later if needed.

I use hexedit
image
I disabled them like that, stopped tvheadend and deleted epgdb.v3 but it just fell back to rl_age

https://files.fm/u/m3pj249ejy
Here are the 3 multiplexes that exist in Cyprus

src/webui/xmltv.c
edisionnano: I totally forgot about the image cache. I think I did it correctly, please check.

DeltaMikeCharlie:
> I disabled them like that, stopped tvheadend and deleted epgdb.v3 but it just fell back to rl_age

OK thanks, it was worth a try.  I'll have to look further into this.

> Here are the 3 multiplexes that exist in Cyprus

Thanks, I will analyse these today.

DeltaMikeCharlie (approved):
Looks good.

DeltaMikeCharlie:
@edisionnano - Here is a cheap and nasty hack to get Greek ratings working.  Apply the attached diff and recompile.  Scan the EPG once, disable the 'gbr' labels and then scan the EPG again.
[dmc.zip](https://github.com/user-attachments/files/21203305/dmc.zip)

Flole998:
This still needs to be squashed into a single commit.

edisionnano:
> @edisionnano - Here is a cheap and nasty hack to get Greek ratings working. Apply the attached diff and recompile. Scan the EPG once, disable the 'gbr' labels and then scan the EPG again. [dmc.zip](https://github.com/user-attachments/files/21203305/dmc.zip)

That seems to have worked
image
and on the xmltv/channels endpoint
image
I don't know if the 0 rating also showing is intended
The numbers also didn't change

edisionnano:
> This still needs to be squashed into a single commit.

Done

edisionnano:
image
ERT channels show as 902
Also, display label doesn't match display age

DeltaMikeCharlie:
> ERT channels show as 902

This is the DVB country group for 'All Countries'.

> Also, display label doesn't match display age

It's not supposed to.  These are placeholder values to prompt the user to add the required values manually.  See my forum posting on the `gre` and `CYP` data analysis.

Flole998:
This can be merged once the failing builds have been fixed (which will be done with #1872).

PR #1872, opened 2025-07-14
Fix builds on debian buster

Flole998:


Flole998:
If this runs through the same changes must be applied to the repo builds.

edisionnano:
I think http://archive.raspberrypi.org/debian/dists/buster/ should be used instead of legacy

Flole998:
Apparently armv6 is not EOL/on the legacy server yet, only armv7 and aarch64, so only those 2 need to get the change.

PR #1877, opened 2025-08-05
fixes coverity 530653

uknunknown:
- fixes coverity 530653
- handle gracefully inability to allocate memory


Flole998:
This doesn't fix the issue coverity flagged (it complains about the memory from strdup never beeing freed there) and we never check for memory allocation errors in tvheadend. There was an idea in #1411 but it was never finished. Only doing it here is not a good idea, this should be done globally in the entire codebase (or just left as-is).

Issue #1878, opened 2025-08-05
[BUG]: Regression Commit 0af87f1? failed to add filters: 'deinterlace_vaapi=mode=21867...


dkonerma:
### Tvheadend Version

v4.3 (development)

### Build Number

4.3-2433~gf1c460feb

### Platform

Debian 11 (Bullseye)

### Architecture

Intel x86_64

### Category

Transcoding

### Description

Maybe a cornercase, but I have a transcoding profile with codec profile libx264 with vaapi hw-deinterlacing on decode enabled.

Using the profile results in

2025-08-05 15:10:20.157 [  DEBUG]:transcode: 0009: 01:H264: [h264 => libx264]: filters: source args: 'video_size=1280x720:pix_fmt=vaapi:time_base=1/90000:pixel_aspect=1/1'
2025-08-05 15:10:20.157 [  DEBUG]:transcode: 0009: 01:H264: [h264 => libx264]: filters: filters: 'deinterlace_vaapi=mode=21867:rate=2:auto=0,scale_vaapi=w=682:h=384,hwdownload,format=pix_fmts=yuv420p'
2025-08-05 15:10:20.158 [  ERROR]:libav: AVFilter: Value 21867.000000 for parameter 'mode' out of range [0 - 4]
2025-08-05 15:10:20.158 [  ERROR]:libav: AVFilter: Error setting option mode to value 21867.
2025-08-05 15:10:20.158 [  ERROR]:libav: AVFilterGraph: Error processing filtergraph: Numerical result out of range
2025-08-05 15:10:20.158 [  ERROR]:transcode: 0009: 01:H264: [h264 => libx264]: filters: failed to add filters: 'deinterlace_vaapi=mode=21867:rate=2:auto=0,scale_vaapi=w=682:h=384,hwdownload,format=pix_fmts=yuv420p'
2025-08-05 15:10:20.158 [WARNING]:transcode: 0009: 01:H264: [h264 => libx264]: failed to transcode packet at pts: 87523 error -34: Numerical result out of range

In the UI there are no options to configure the deinterlace_vaapi=mode (I see these options only for h264_vaapi codec profiles). Maybe something uninitialized results in deinterlace_vaapi=mode=21867?

### Reproduction

1. Create a sw-transcoding profile using libx264
2. Enable the "Hardware accelleration" option for that profile, leave other defaults untouched
3. Try to playback using this profile

Flole998:
[...]

PR #1880, opened 2025-08-06
transcode: fix VAAPI deinterlace mode handling for software profiles

jahutchi:
Move 'deinterlace_vaapi_mode' from the VAAPI codec profile to the generic Main Video Codec Profile, allowing this setting to be applied when using VAAPI hardware deinterlacing with software-based transcode profiles such as libx264.

This fixes a bug where 'deinterlace_vaapi_mode' was left uninitialized for non-VAAPI codec profiles, resulting in invalid filter strings like 'deinterlace_vaapi=mode=21867:rate=2:auto=0' and filter graph setup failures.

The patch also:
- Adds validation for the mode value (range 0–4)
- Dynamically enables/disables the VAAPI mode field in the WebUI based on encoder and decoder settings
- Consolidates enum and mode list generation logic under 'profile_video_class.c'

This ensures that VAAPI deinterlacing can be correctly configured and used across hybrid transcode profiles, improving compatibility and user control.

Fixes: #1878

### DIFF
--- a/src/transcoding/codec/codecs/libs/vaapi.c
+++ b/src/transcoding/codec/codecs/libs/vaapi.c
@@ -84,12 +84,6 @@
 #define VAAPI_ENC_B_REFERENCE_I_P       1
 #define VAAPI_ENC_B_REFERENCE_I_P_B     2
 
-#define VAAPI_DEINT_MODE_DEFAULT        0
-#define VAAPI_DEINT_MODE_BOB            1
-#define VAAPI_DEINT_MODE_WEAVE          2
-#define VAAPI_DEINT_MODE_MADI           3
-#define VAAPI_DEINT_MODE_MCDI           4
-
 #define UI_CODEC_AVAILABLE_OFFSET       0
 #define UI_MAX_B_FRAMES_OFFSET          1
 #define UI_MAX_QUALITY_OFFSET           4
@@ -136,19 +130,6 @@ rc_mode_get_list( void *o, const char *lang )
         { N_("ICQ"),    VAAPI_ENC_PARAMS_RC_ICQ },
         { N_("QVBR"),   VAAPI_ENC_PARAMS_RC_QVBR },
         { N_("AVBR"),   VAAPI_ENC_PARAMS_RC_AVBR },
-    };
-    return strtab2htsmsg(tab, 1, lang);
-}
-
-static htsmsg_t *
-deinterlace_vaapi_mode_get_list( void *o, const char *lang )
-{
-    static const struct strtab tab[] = {
-        { N_("Default"),                                 VAAPI_DEINT_MODE_DEFAULT },
-        { N_("Bob Deinterlacing"),                       VAAPI_DEINT_MODE_BOB },
-        { N_("Weave Deinterlacing"),                     VAAPI_DEINT_MODE_WEAVE },
-        { N_("Motion Adaptive Deinterlacing (MADI)"),    VAAPI_DEINT_MODE_MADI },
-        { N_("Motion Compensated Deinterlacing (MCDI)"), VAAPI_DEINT_MODE_MCDI },
     };
     return strtab2htsmsg(tab, 1, lang);
 }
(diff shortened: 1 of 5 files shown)


uknunknown:
From what I understand you are transferring mode and deint also in software decoding.
deint
But from documentation the parameters have a slightly different meaning .... even if the names are identical (mode and deint). Also software have an extra parameter.  
https://ffmpeg.org/ffmpeg-filters.html#yadif-1
The information inside combos have to be different for software deinterlace and vaapi hardware deinterlace.
How to implement that? 
Maybe have two different sets of combos:
- one set for software ... that become visible only when you disable the 'hardware decode check box'
- one set for vaapi hardware ... that become visible when you enable the hardware decoding AND hardware decoder is auto or vaapi

Please note that this way we can scale that in the future with other hardware deinterlace filters for NVDEC , QSV .... and so on.

jahutchi:
> From what I understand you are transferring mode and deint also in software decoding. deint But from documentation the parameters have a slightly different meaning .... even if the names are identical (mode and deint). Also software have an extra parameter. https://ffmpeg.org/ffmpeg-filters.html#yadif-1 The information inside combos have to be different for software deinterlace and vaapi hardware deinterlace. How to implement that? Maybe have two different sets of combos:
> 
> * one set for software ... that become visible only when you disable the 'hardware decode check box'
> * one set for vaapi hardware ... that become visible when you enable the hardware decoding AND hardware decoder is auto or vaapi
> 
> Please note that this way we can scale that in the future with other hardware deinterlace filters for NVDEC , QSV .... and so on.

IMO: the meanings of those two parameters mapped quite nicely between vaapi and yadif

**Tvheadend Setting: Deinterlace Rate Type**
Help Text: Frame rate combines the two interlaced fields to create a single frame. Field rate processes each field independently, outputting as individual frames, which enables higher temporal resolution by producing one frame per field. 

Maps to:
VAAPI: Rate setting see https://ffmpeg.org/doxygen/trunk/vf__deinterlace__vaapi_8c.html. (see deint_vaapi_options) 
Yadif: Mode setting https://ffmpeg.org/ffmpeg-filters.html#yadif-1. (link you quoted)

**Tvheadend Setting: Deinterlace Fields Only**
Help Text: Enable this option to only deinterlace fields, passing progressive frames unchanged. This is useful for mixed content, allowing progressive frames to bypass deinterlacing for improved efficiency and quality.

Maps to
VAAPI: Auto setting
Yadif: Deint setting
(see same links as above)

There is also the **Deinterlace VAAPI mode** setting in Tvheadend which (with this PR) will only be enabled when hw accel is enabled along with Prefer VAAPI or Auto selected in h/w accel details

There is an additional ffmpeg parameter for yadif named parity. I thought of adding this (for s/w deinterlace only) but reading the docs wasn't sure why anyone would choose anything other than auto here. Happy to follow up with  PR to add this option if anyone has a use-case.

uknunknown:
I don't think that implemented only some "cherry picked" options (that are common now) is the right approach. For example yadif, mode has 4 options and you implemented only 2. Don't understanding me wrong .. is much better then not having any option ... but I am saying it could be better.
mode
If you separate completely the software (yadif) implementation from hardware implementation (vaapi ... and in the future other hw implementations with very different settings), you build an infrastructure that is scalable. 
We can keep it like that, for now ... and improve it in the future.

jahutchi:
> I don't think that implemented only some "cherry picked" options (that are common now) is the right approach. For example yadif, mode has 4 options and you implemented only 2. Don't understanding me wrong .. is much better then not having any option ... but I am saying it could be better. mode If you separate completely the software (yadif) implementation from hardware implementation (vaapi ... and in the future other hw implementations with very different settings), you build an infrastructure that is scalable. We can keep it like that, for now ... and improve it in the future.

I fully agree with your observations. This specific PR addresses the original implementation in response to the blocking user issue #1878.

In the longer term, your proposed approach would be a sensible way to introduce additional configuration options for fine-tuning the YADIF parameters. We could also consider exposing alternative software filters, as some may argue that YADIF is not always the optimal choice—it often depends on the hardware capabilities. For instance, BWDIF has demonstrated up to a 23× performance improvement on systems with AVX-512 support ([source](https://www.phoronix.com/news/FFmpeg-Bwdif-AVX-512)). It may therefore be worthwhile to offer a selectable deinterlacing method in future too.

Flole998:
I need to fix the debian docker thing, then this can be merged

PR #1885, opened 2025-08-09
Fix memory leak in iptv_mux.c

Flole998:


Flole998:
Okay this one is wrong.  It needs to be moved outside the if else-path.

PR #1887, opened 2025-08-14
version: stop using tilde between build and githash

chewitt:
The tilde that separates githash in e.g. `tvheadend-4.3-2432~g0af87f13f` causes formatting issues in the forum and other locations that use markdown, so this is a naeive/blind attempt at changes to use a hyphen instead. I have no means of test building from current vacation location so let's see if CI builds? and if not, what failure messages show up.


Flole998:
I wonder if that regex/sed command is now still doing something or if it's now obsolete.

chewitt:
Closing as superseded by #1922

PR #1479, opened 2022-11-28
dvb_psi_pmt: Recognize AC-4 audio descriptor

kasper93:
This allows pass-through of AC-4 audio stream.

Will need adjustment when AC-4 gets support in FFmpeg to enable transcode.

Link: https://tvheadend.org/issues/6222
Signed-off-by: Kacper Michajłow <kasper93@gmail.com>

PR #1889, opened 2025-08-16
dvb_psi_pmt: Don't recognize all extension descriptors as AC-4 audio

mkuron:
A few of my channels are incorrectly showing up with AC-4 audio streams on Tvheadend's "Service details" screen. These tracks really are AC-3.

I believe that is a bug introduced in #1479, which originally added AC-4 support. This code looks for a DVB descriptor with _tag value_ `0x7f`, but according to [ETSI EN 300 468](https://www.etsi.org/deliver/etsi_en/300400_300499/300468/01.18.01_60/en_300468v011801p.pdf), that is not the _AC-4 descriptor_ but the _extension descriptor_. Tvheadend's `DVB_DESC_AC4` should thus be renamed to `DVB_DESC_EXTENSION`. Only if the extension descriptor's _extension tag value_ is `0x15` do we have an AC-4 descriptor.

Due to lack of an actual AC-4 broadcast, I cannot test my patch for correctness, but at least it stops marking streams with extension descriptors as AC-4.

My AC-3 streams do contain an extension descriptor, but its _extension tag value_ is `0x06`, corresponding to a _supplementary audio descriptor_. It should be easy enough to extend the code here to parse that descriptor too and extract the language and clean/hearing-impaired/visually-impaired classification.

### DIFF
--- a/src/input/mpegts/dvb_psi_pmt.c
+++ b/src/input/mpegts/dvb_psi_pmt.c
@@ -249,6 +249,7 @@ dvb_psi_parse_pmt
   int tt_position;
   int video_stream;
   int rds_uecp;
+  int ac4;
   int pcr_shared = 0;
   const char *lang;
   uint8_t audio_type, audio_version;
@@ -311,6 +312,7 @@ dvb_psi_parse_pmt
     composition_id = -1;
     ancillary_id = -1;
     rds_uecp = 0;
+    ac4 = 0;
     position = 0;
     tt_position = 1000;
     lang = NULL;
@@ -439,8 +441,14 @@ dvb_psi_parse_pmt
           hts_stream_type = SCT_EAC3;
         break;
 
-      case DVB_DESC_AC4:
-        if(estype == 0x06 || estype == 0x81)
+      case DVB_DESC_EXTENSION:
+        if(dlen < 1)
+          break;
+
+        if(ptr[0] == 0x15) /* descriptor_tag_extension : AC-4 */
+          ac4 = 1;
+
+        if((estype == 0x06 || estype == 0x81) && ac4)
           hts_stream_type = SCT_AC4;
         break;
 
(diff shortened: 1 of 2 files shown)


mkuron:
Fixing the @sonarqubecloud bot's maintainability warning would require refactoring the entire function, so I am not going to do that. My code addition mirrors the code that identifies RDS streams, which is just below.

src/input/mpegts/dvb_psi_pmt.c
```diff
@@ -439,8 +441,14 @@ dvb_psi_parse_pmt
           hts_stream_type = SCT_EAC3;
         break;
 
-      case DVB_DESC_AC4:
-        if(estype == 0x06 || estype == 0x81)
+      case DVB_DESC_EXTENSION:
+        if(dlen < 1)
```
Flole998, line 445: What are you trying to check with this code?

mkuron: If this is a zero-length descriptor, the `ptr[0]` on line 448 would be outside of the descriptor. I don't know enough about MPEG transport streams to judge whether zero-length descriptors are even possible, so I am simply mirroring what was done on line 456 for RDS streams.

src/input/mpegts/dvb_psi_pmt.c
```diff
@@ -439,8 +441,14 @@ dvb_psi_parse_pmt
...
+          break;
+
+        if(ptr[0] == 0x15) /* descriptor_tag_extension : AC-4 */
```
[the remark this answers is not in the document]
mkuron: We don‘t have named constants for elementary stream types (e.g. 0x06 and 0x81, used throughout this function) or for ancillary data IDs (line 459) either, so I think keeping this magic number is fine. Readability/maintainability should be sufficiently addressed by the comment, which mirrors how line 459 dealt with its magic number.

mkuron:
Is there anything else you would like me to change on this pull request, @Flole998? There are 4 failed builds, but those seem unrelated to my changes.

PR #1905, opened 2025-08-25
issues: misc. template updates

chewitt:
The current GitHub issue template claims v4.3 is `development` but we have declared it to be `stable` and v4.4 does not exist, so reflect that in the releases drop-list. Also update the distro drop-list with some newer items.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -10,8 +10,7 @@ body:
       description: "What Tvheadend version are you using?"
       options:
         - v4.2 (unsupported)
-        - v4.3 (development)
-        - v4.4 (stable)
+        - v4.3 (stable)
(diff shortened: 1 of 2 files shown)


Flole998:
Since #1893  we also have support for enterprise Linux 9 and 10, so while we are at it those should also be added to the dropdown

chewitt:
Updated to include EL8/EL9..

Flole998:
Sorry, totally forgot about this: It's also necessary to change 
https://github.com/tvheadend/tvheadend/blob/4bf4f7d87ccf343c3bbe9cb65eb5a6133d168044/.github/workflows/enforce-issue-template.yml#L23C8 as otherwise all new issues would be closed automatically as its not matching the expected values

Flole998:
You probably only have to remove the ` \\(development\\)|4` part from it, then it should fit :)

PR #1923, opened 2025-09-18
Fix translation error

euca01:
Fix: Translation Error in French
> "Sat" for Samedi and not "Satellite"

### DIFF
--- a/intl/tvheadend.fr.po
+++ b/intl/tvheadend.fr.po
@@ -5531,7 +5531,7 @@ msgstr ""
 
 #: src/dvr/dvr_autorec.c:777
 msgid "Sat"
-msgstr "Satellite"
+msgstr "Sam"
 
 #: src/input/mpegts/linuxdvb/linuxdvb_satconf.c:1422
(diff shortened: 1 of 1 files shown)


chewitt:
@euca01 I'm closing the PR because we only accept changes to the original en_GB text here. All translations are handled through Transifex, see: https://www.transifex.com/projects/p/tvheadend/ .. which then pushes a PR to the project, to keep everything in-sync.

euca01:
@chewitt There is no translation error in transifex.
The error is only present in Github.

i don't understand how to fix it.

When we choose "Digital Video Recorder" > "Autorecs" > "Days of week" :
Sat should be translated to Sam
(and not Satellite)

image

chewitt:
I've now corrected the translation error in Transifex (which did exist). The next problem will be that French `intl/tvheadend` where that mistake was located, is only 45% translated, and with our current sync config Transifex won't start pushing PR's with translation changes back to GitHub until the language source reaches 80%.

PR #1964, opened 2025-10-18
Update configure

j0nnymoe:
Enabling Flags: nvenc/hdhomerun_client/pngquant


Flole998:
Its unhappy about libnpp missing. Also if this is supposed to be actually merged some day have a look at the files in the autobuild directory, this is probably only relevant for x86/x64 systems and not ARM (certainly not raspberry pi os), its quite brilliant as you can choose to add a build option to all i386 architectures by editing a single file. The installation of missing libs can be done in the CI github action file (dont forget to also do it for the repo build github action file).

PR #1966, opened 2025-10-22
remove libnpp and cuda sdk dep, enable nvenc on x86_64

aptalca:
libnpp is an older cuvid option that requires the Nvidia SDK during build time as well as runtime per [ffmpeg docs](https://trac.ffmpeg.org/wiki/HWAccelIntro#cuda-nvccandlibnpp). I doubt it is necessary or valuable for tvheadend users.

This PR attempts to build tvheadend with nvenc/nvdec and cuvid support on x86_64.

If a replacement for libnpp is necessary, we can enable cuda-llvm, which doesn't require Nvidia SDK, but just clang and llvm.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -239,7 +239,7 @@ jobs:
       - name: Workaround safe directory
         run: git config --global --add safe.directory "$(pwd)"
       - name: build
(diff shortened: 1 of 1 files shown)


aptalca:
It looks like most if not all of the failures are due to `failed: Network is unreachable.` for external repos (ffmpeg dependency libs).

I also noticed that not all x86_64 builds enabled nvenc. I guess some of those use hardcoded options and not the ones from autobuilds.

Flole998:
They should all use the autobuild options, which ones don't have it included?

aptalca:
All the fedora ones it seems: https://github.com/tvheadend/tvheadend/actions/runs/18717978773/job/53414315021?pr=1966#step:7:114

Flole998:
Yeah Fedora/RPMs work in a different way

aptalca:
I see, they're defined in the workflow file

PR #2041, opened 2026-01-06
make reset counters a global option; fix linuxdvb mux lock after failed tuning; other improvements

Trujulu:
- linuxdvb: Fixes mux becoming unusable after editing its configuration due to EINVAL caused by incorrect tuning options.
- make the auto-reset counters option global, allowing the user to enable it on demand.
- fix several missing locks and memory leaks detected by Coverity (CID-352140, CID-470910, CID-551184)

### DIFF
--- a/src/channels.c
+++ b/src/channels.c
@@ -131,13 +131,15 @@ channel_class_autoname_set ( void *obj, const void *p )
 {
   channel_t *ch = (channel_t *)obj;
   const char *s;
+  char *chan_name;
   int b = *(int *)p;
   if (ch->ch_autoname != b) {
     if (b == 0 && tvh_str_default(ch->ch_name, NULL) == NULL) {
       s = channel_get_name(ch, NULL);
       if (s) {
+        chan_name = strdup(s);
         free(ch->ch_name);
-        ch->ch_name = strdup(s);
+        ch->ch_name = chan_name;
       } else {
         return 0;
       }
--- a/src/input/mpegts/iptv/iptv.c
+++ b/src/input/mpegts/iptv/iptv.c
@@ -576,6 +576,7 @@ iptv_input_thread ( void *aux )
       if ((n = im->im_handler->read(mi, im)) < 0) {
         tvherror(LS_IPTV, "read() error %s", strerror(errno));
         im->im_handler->stop(mi, im);
+        tvh_mutex_unlock(&iptv_lock);
         break;
       }
       r = iptv_input_recv_packets(im, n);
--- a/src/input/mpegts/iptv/iptv.c
+++ b/src/input/mpegts/iptv/iptv.c
@@ -336,18 +336,10 @@ iptv_input_start_mux ( mpegts_input_t *mi, mpegts_mux_instance_t *mmi, int weigh
   const char *scheme;
   url_t url;
 
-  /* Already active */
-  if (im->mm_active)
+  /* Active? */
+  if (im->mm_active) {
     return 0;
-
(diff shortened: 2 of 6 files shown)


src/input/mpegts/iptv/iptv_http.c
```diff
@@ -473,10 +473,10 @@ iptv_http_complete
         free(hp->hls_url_after_key);
         hp->hls_url_after_key = url;
         url = strdup(s);
-        free(absolute_key_url);
         hc->hc_data_complete = iptv_http_complete_key;
         sbuf_reset(&hp->key_sbuf, 32);
       }
+      free(absolute_key_url);
```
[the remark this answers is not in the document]
Trujulu: This is a bit tricky; that "loop" is a mess. I submitted a change to free at the end.

src/input/mpegts/iptv/iptv.c
```diff
@@ -336,18 +336,13 @@ iptv_input_start_mux ( mpegts_input_t *mi, mpegts_mux_instance_t *mmi, int weigh
...
-  /* Reset Error Counters */
-  atomic_set(&mmi->tii_stats.unc, 0);
-  atomic_set(&mmi->tii_stats.cc, 0);
-  tvh_mutex_lock(&mmi->tii_stats_mutex);
-  mmi->tii_stats.te = 0;
-  mmi->tii_stats.ec_block = 0;
-  mmi->tii_stats.tc_block = 0;
-  tvh_mutex_unlock(&mmi->tii_stats_mutex);
+  }
+  tvh_mutex_unlock(&iptv_lock);
```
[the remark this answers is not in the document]
Trujulu: This could all be wrapped in a single lock with the im->mm_active check at the beginning (with mutex), but I don’t like blocking the IPTV thread just to check URLs while other threads could be doing work 🤔

src/channels.c
```diff
@@ -137,7 +137,7 @@ channel_class_autoname_set ( void *obj, const void *p )
       s = channel_get_name(ch, NULL);
       if (s) {
         free(ch->ch_name);
-        ch->ch_name = strdup(s);
+        ch->ch_name = s ? strdup(s) : NULL;
```
Flole998: 2 lines above this there is already a null-check for s, so s will never be null at this point

Flole998 (changes requested):

src/channels.c
Trujulu: @Flole998 If you look at channels.c:826, inside the channel_get_name() function, you’ll notice that `s = ch->ch_name`, even though it has already been freed at that point. I’ve submitted a patch that addresses this issue more sensibly.

Flole998: I'm not sure I understand what you mean.... In your patch you are checking if s is null. That check was done 2 lines before though, so it will never be null here as it's not being written to in the meantime?

Trujulu: It’s not that `s` becomes NULL. The problem is that, in one case, `s` points to `ch->ch_name`, and then `ch->ch_name` is freed using `free(ch->ch_name)`. In that situation, calling `strdup` on the freed memory causes a crash/undefined behavior.

You can check https://scan8.scan.coverity.com/#/project-view/33873/10553?selectedIssue=352140

P.S.: I submitted a change to call `strdup(s)` and saving into a `char *` before it is freed.

Flole998: This change looks much better, if you could rename "tmp" to something more meaningful (like chan_name) that would be great!

Flole998 (approved):

PR #2053, opened 2026-01-18
dvb_psi: prevent reading outside tsb bounds

Trujulu:
- Fix potential infinite loop when handling truncated TS input (r == 0)
- Prevent out-of-bounds reads on tsb buffer with lossy or incomplete MPEG-TS streams
- Rework PUSI pointer_field reassembly to improve robustness and avoid negative section lengths (CID-498138)

Tested for a week without major issues.

### DIFF
--- a/src/input/mpegts/dvb_psi_lib.c
+++ b/src/input/mpegts/dvb_psi_lib.c
@@ -71,11 +71,12 @@ static int
 mpegts_psi_section_reassemble0
   ( mpegts_psi_table_t *mt, const char *logpref,
     const uint8_t *data,
-    int len, int start, int crc,
+    int len, uint8_t start, uint8_t crc,
     mpegts_psi_section_callback_t cb, void *opaque)
 {
   uint8_t *p = mt->mt_sect.ps_data;
-  int excess, tsize;
+  uint16_t tsize;
+  int excess;
 
   if(len <= 0)
     return -1;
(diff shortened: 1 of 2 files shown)


src/input/mpegts/dvb_psi_lib.c
```diff
@@ -166,25 +166,28 @@ mpegts_psi_section_reassemble
     return;
   }
 
-  if(pusi) {
-    uint8_t len2 = tsb[off++];
-    uint8_t off2 = off;
-    off += len2;
-    if (off > 188)
+  if (pusi) {
+    int pointer_field = tsb[off++];
```
Flole998: Is switching to an int really the right thing to do here? The pointer_field is an 8 bit unsigned int read from the table, `off` (as seen above) should also be an uint8_t (as it can never be higher than 255, let alone be negative) and so should `pre` then, only pre_end could be larger (but never negative), so uint16_t would be more appropriate for that one.

Trujulu: You are right. Rebased, thanks Flole!

Flole998: Actually I think that's wrong/incomplete: `off` could be larger than 255, so `off` should be uint16_t (and so should be `pre` then).

Btw: pusi can be uint8_t, might be worth to change that one aswell while we are touching this code.

Trujulu: Noticed that `pusi` and `crc` are booleans, they could be uint8 I guess? It seems that project does not uses stdbool

Flole998: I hadn't seen CRC but yes, if it's only 1/0 then uint8_t is fine. I don't think we use actual booleans anywhere, but the compiler will (usually) turn it into an 8 bit value anyways, so there isn't much difference between uint8_t and bool other than readability.

PR #2054, opened 2026-01-18
config.c: disable HDHomeRun configuration when not built in

Trujulu:
Hide HDHomeRun configuration when support is not enabled at build time

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2560,6 +2560,7 @@ const idclass_t config_class = {
       .opts   = PO_EXPERT,
       .group  = 5
     },
+#if ENABLE_HDHOMERUN_CLIENT
     {
       .type   = PT_STR,
       .id     = "hdhomerun_ip",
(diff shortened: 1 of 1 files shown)


src/config.c
```diff
@@ -2700,6 +2692,7 @@ const idclass_t config_class = {
     },
     {
       .type   = PT_INT,
+      .intextra = INTEXTRA_RANGE(1, 128, 1),
```
Flole998, line 2695: Could you give a (very short) explanation in the PR (or even here as a comment) why this was done? Just so if someone wonders why this was changed, they can find the answer here.

Trujulu: Yes, this range is specified in `iptv.c:55` for `tpool_count`, using values outside of it does not make sense.

```
static inline int iptv_tpool_safe_count(void)
{
  return MINMAX(config.iptv_tpool_count, 1, 128);
}

Flole998: Thanks!

PR #2083, opened 2026-03-09
 idnode.c: unlock idnode on failed uuid_set

enricolorenzoni59:
I've observed a startup failure when I had a stale .tmp file in my muxes folder. This patch seems to be an improvement.

### DIFF
--- a/src/idnode.c
+++ b/src/idnode.c
@@ -113,6 +113,7 @@ idnode_insert(idnode_t *in, const char *uuid, const idclass_t *class, int flags)
 
     if (uuid_set(&u, uuid)) {
       in->in_class = NULL;
+      idnode_unlock();
       return -1;
     }
     uuid_duplicate(&in->in_uuid, &u);


Flole998:
I think its better to not do the refactoring. I know sonar is unhappy with that but this is even harder to read than before. What do you think? Is the refactoring helping with readability?

enricolorenzoni59:
> I think its better to not do the refactoring. I know sonar is unhappy with that but this is even harder to read than before. What do you think? Is the refactoring helping with readability?

Not really, I’d pick the first two commits only, if this repo was mine. I’ve added the refactor only for sonarqube.

enricolorenzoni59:
@Flole998 do you want me to drop the 3rd commit, or will you do it when squashing/merging on your side?

Flole998:
Yes, please drop the last one. I think it might also make sense to drop the second one and only do the "real fix", so unless you insist on keeping the second one it would be good to drop that one as well and only pick the first one.

enricolorenzoni59:
> Yes, please drop the last one. I think it might also make sense to drop the second one and only do the "real fix", so unless you insist on keeping the second one it would be good to drop that one as well and only pick the first one.

No problem at all. Kept the single liner commit only. Thanks for reviewing!

PR #2084, opened 2026-03-18
satip: expose DVB-S frequency override and fix matching priority

maedula:
## Summary

- Expose the existing but hidden `mm_dvb_satip_dvbs_freq` field in the DVB-S mux editor UI by adding a property entry to `dvb_mux_dvbs_class` in `mpegts_mux_dvb.c`. This allows operators to assign virtual SAT>IP frequencies to individual DVB-S muxes.
- Move the SAT>IP frequency override matching in `rtsp_start()` to run before the normal `dvb_network_find_mux()` lookup, ensuring that explicitly configured virtual frequencies always take precedence.

## Motivation

DVB-S2X MIS (Multiple Input Streams) muxes share the same physical transponder frequency but have different ISI (Input Stream Identifier) values. The SAT>IP protocol has no native ISI parameter, and `dvb_network_find_mux()` checks `dmc_fe_stream_id` which SAT>IP clients never set for DVB-S. This means MIS muxes are unreachable via SAT>IP without this change.

By assigning a unique virtual frequency to each MIS mux (via the new UI field), SAT>IP clients can address individual streams. TVHeadend matches the virtual frequency and tunes to the real transponder with the correct ISI.

**Note:** The DVB-S field uses kHz (not Hz like DVB-C/T) because DVB-S frequencies internally use kHz after RTSP parsing (`freq_mhz * 1000`), and satellite frequencies in Hz would overflow `uint32_t`.

## Changes

- `src/input/mpegts/mpegts_mux_dvb.c`: Add `dvb_satip_dvbs_freq` property to `dvb_mux_dvbs_class` (PT_U32, PO_ADVANCED)
- `src/satip/rtsp.c`: Move SAT>IP frequency override loop before the normal mux matching loop in `rtsp_start()`, guarded by `if (mux == NULL)`

## Test plan

Tested on aarch64 with a Panasonic TV as SAT>IP client:
- Configured a DVB-S2X MIS mux (12466 MHz V, ISI 13) with virtual frequency 10700000 kHz (10700 MHz)
- Panasonic TV manual scan on 10700 MHz successfully received ISI 13 channels
- Existing non-MIS muxes continue to work via normal frequency matching
- Switching back to the official image (without this change) works -- the field is silently ignored in mux configs

### DIFF
--- a/src/input/mpegts/mpegts_mux_dvb.c
+++ b/src/input/mpegts/mpegts_mux_dvb.c
@@ -554,6 +554,19 @@ const idclass_t dvb_mux_dvbs_class =
       .desc     = N_("For example: 312000000. This frequency is 312Mhz."),
       .opts     = PO_ADVANCED
     },
+    {
+      .type     = PT_U32,
+      .id       = "dvb_satip_dvbs_freq",
+      .name     = N_("SAT>IP DVB-S frequency (kHz)"),
+      .off      = offsetof(dvb_mux_t, mm_dvb_satip_dvbs_freq),
+      .desc     = N_("Virtual DVB-S frequency in kHz for SAT>IP clients. "
+                     "When set, SAT>IP clients tune to this mux using "
+                     "this frequency instead of the real transponder frequency. "
+                     "Required for DVB-S2X MIS muxes sharing the same "
+                     "physical frequency with different ISI values. "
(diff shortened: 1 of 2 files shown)


maedula:
Squashed the branch to a single commit as requested. GitHub would not reopen this PR after the force-push, so continued in #2128.

PR #2093, opened 2026-04-04
timeshift: fix subscriptionLive flooding clients with stale packets

### LINKED ISSUE #2087: [BUG]: _timeshift_flush_to_live() delivers all buffered packets unpaced, overwhelming clients
### Tvheadend Version

v4.3 (stable)

### Build Number

4.3-2634

### Platform

Debian 12 (Bookworm)

### Architecture

Intel x86_64

### Category

Timeshifting

### Description

When a client sends `subscriptionSkip` (or `subscriptionLive`) to return to the live edge during timeshift, the server calls `_timeshift_flush_to_live()` which delivers **all remaining buffered packets in a tight `while` loop** with no pacing or flow control. For long timeshift buffers (minutes to hours of content), this can:

1. **Overflow the TCP send buffer** → kernel sends RST → connection drops
2. **Overwhelm the client's demuxer/decoder pipeline** with a burst of stale data that will never be displayed
3. **Delay actual live playback** because the client must receive/process/discard the entire burst before real-time packets resume

**Affected code** — `src/timeshift/timeshift_reader.c`, `_timeshift_flush_to_live()`:

```c
static int _timeshift_flush_to_live
  ( timeshift_t *ts, timeshift_seek_t *seek, int *wait )
{
  streaming_message_t *sm;

  while (seek->file) {
    if (_timeshift_read(ts, seek, &sm, wait) == -1)
      return -1;
    if (!sm) break;
    timeshift_packet_log("ouf", ts, sm);
    streaming_target_deliver2(ts->output, sm);  // delivers every packet, no pacing
  }
  return 0;
}
```

This `while` loop reads every remaining packet from the timeshift buffer and pushes it downstream immediately. There is no rate limiting, no check for client TCP backpressure, and no option to skip/discard packets that won't be needed.

**Proposed fix:** Since the client is skipping to live, the buffered packets between the current position and the live edge are unwanted. Instead of delivering them, they should be freed:

```c
static int _timeshift_flush_to_live
  ( timeshift_t *ts, timeshift_seek_t *seek, int *wait )
{
  streaming_message_t *sm;

  while (seek->file) {
    if (_timeshift_read(ts, seek, &sm, wait) == -1)
      return -1;
    if (!sm) break;
    timeshift_packet_log("ouf", ts, sm);
    streaming_msg_free(sm);  // discard instead of delivering
  }
  return 0;
}
```

One-line change: replace `streaming_target_deliver2(ts->output, sm)` with `streaming_msg_free(sm)`. After the loop completes, the reader thread naturally resumes reading from the live ringbuffer position, so the client receives fresh live packets without any burst.

**Alternative:** If some clients expect to receive the flushed packets for re-sync, a compromise would be to deliver only the last few seconds of buffered data and discard the rest.

**Client-side workaround:** HTSP clients can monitor `timeshiftStatus` messages and drop all incoming `muxpkt` packets after sending the skip request until `timeshiftStatus.shift` confirms the stream is back at the live edge (shift ≤ 1). This works but wastes bandwidth.

**Related:** Bug [#2744](https://tvheadend.org/issues/2744) — crash when skipping to end of timeshift buffer (fixed in 4.1-1333, but `_timeshift_flush_to_live` design unchanged).

### Logfile

_No response_

### Reproduction

1. Start watching a live channel via an HTSP client
2. Pause for 2+ minutes (builds timeshift buffer)
3. Send `subscriptionSkip` with a large forward offset to return to live
4. Observe: server dumps the entire remaining buffer at wire speed before resuming live delivery

With longer timeshift buffers, the problem is more severe. On a local network, the burst can be hundreds of MB delivered in seconds.

oliversluke:
## Summary

- Replace the `SMT_SKIP_LIVE` fast-forward hack (marked `// TODO: change this sometime!`) that delivered all buffered packets unpaced
- Instead, close the read position, re-send the latest stream start, and transition directly to `TS_LIVE`
- The timeshift buffer is preserved so users can still seek backwards after going live
- Added NULL guard for `ctrl` in the error path to prevent a crash

## Problem

When a client sends `subscriptionLive`, the reader thread set `skip_time = INT64_MAX` and fast-forwarded through the entire timeshift buffer, delivering every remaining packet in a tight loop. This flooded clients with potentially minutes of stale data, causing stuttering, high latency, and long transition times back to live. Clients had to work around this by using `subscriptionSkip` to a near-live position instead.

## Testing

Tested with a custom HTSP client (satTV) against a live DVB-S2 setup:
- Paused for 30+ seconds to build timeshift buffer
- Called `subscriptionLive` — transition to live was near-instant
- Verified backward seeking still works after going live (buffer preserved)
- Verified `subscriptionSkip` still works as before (backwards compatible)

Fixes #2087

### DIFF
--- a/src/timeshift/timeshift_reader.c
+++ b/src/timeshift/timeshift_reader.c
@@ -668,13 +668,60 @@ void *timeshift_reader ( void *p )
                   ts->full = 0;
                 }
 
-                /* Release */
-                if (sm)
+                /* Release current read packet */
+                if (sm) {
                   streaming_msg_free(sm);
+                  sm = NULL;
+                }
 
-                /* Find end */
-                skip_time = 0x7fffffffffffffffLL;
-                // TODO: change this sometime!
+                /* Close read file and reset seek position */
+                _read_close(seek);
+
+                /* Re-send the latest stream start so client has codec info */
+                {
+                  timeshift_file_t *newest = timeshift_filemgr_newest(ts);
+                  timeshift_file_t *f = newest;
+                  streaming_message_t *sstart_sm = NULL;
+                  int tmpend = 0;
+                  while (f && !sstart_sm) {
+                    timeshift_index_data_t *ti = TAILQ_LAST(&f->sstart, timeshift_index_data_list);
+                    if (ti)
+                      sstart_sm = ti->data;
+                    else
+                      f = timeshift_filemgr_prev(f, &tmpend, 0);
+                  }
+                  if (sstart_sm)
+                    streaming_target_deliver2(ts->output, streaming_msg_clone(sstart_sm));
+                  if (f != newest)
+                    timeshift_file_put(f);
+                  timeshift_file_put(newest);
+                }
+
+                /* Transition to live */
+                tvhdebug(LS_TIMESHIFT, "ts %d skip to live", ts->id);
+                cur_speed = 100;
+                ts->state = TS_LIVE;
+                tvhtrace(LS_TIMESHIFT, "reader - set TS_LIVE (skip live)");
+
+                /* Send speed change to output */
+                {
+                  streaming_message_t *spd = streaming_msg_create_code(SMT_SPEED, cur_speed);
+                  streaming_target_deliver2(ts->output, spd);
+                }
+
+                /* Build skip response for the client */
+                skip->type = SMT_SKIP_ABS_TIME;
+                skip->time = 0;
+                timeshift_fill_status(ts, &skip->timeshift, ts->buf_time);
+                mono_last_status = mono_now;
+
+                /* Deliver the skip response and clear ctrl */
+                streaming_target_deliver2(ts->output, ctrl);
+                ctrl = NULL;
+                skip = NULL;
+
+                /* Send updated timeshift status (shift=0) */
+                timeshift_status(ts, ts->buf_time);
               }
               break;
 
@@ -751,7 +798,7 @@ void *timeshift_reader ( void *p )
           }
 
           /* Error */
-          if (!skip) {
+          if (!skip && ctrl) {
             ((streaming_skip_t*)ctrl->sm_data)->type = SMT_SKIP_ERROR;
             streaming_target_deliver2(ts->output, ctrl);
             ctrl = NULL;
--- a/src/timeshift/timeshift_reader.c
+++ b/src/timeshift/timeshift_reader.c
@@ -679,22 +679,21 @@ void *timeshift_reader ( void *p )
 
                 /* Re-send the latest stream start so client has codec info */
                 {
-                  timeshift_file_t *newest = timeshift_filemgr_newest(ts);
-                  timeshift_file_t *f = newest;
+                  timeshift_file_t *f = timeshift_filemgr_newest(ts);
                   streaming_message_t *sstart_sm = NULL;
                   int tmpend = 0;
-                  while (f && !sstart_sm) {
+                  while (f) {
                     timeshift_index_data_t *ti = TAILQ_LAST(&f->sstart, timeshift_index_data_list);
-                    if (ti)
+                    if (ti) {
                       sstart_sm = ti->data;
-                    else
-                      f = timeshift_filemgr_prev(f, &tmpend, 0);
+                      break;
+                    }
+                    f = timeshift_filemgr_prev(f, &tmpend, 0);
                   }
                   if (sstart_sm)
                     streaming_target_deliver2(ts->output, streaming_msg_clone(sstart_sm));
-                  if (f != newest)
+                  if (f)
                     timeshift_file_put(f);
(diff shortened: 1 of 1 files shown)


Flole998:
I have talked to some people from Team Kodi about this and the general standpoint seems to be that this requires quite some testing on multiple systems (especially those using hardware decoding) to be sure that the decoders are fine with skipping a bunch of packets.

oliversluke:
Thanks for raising this concern and for checking with the Kodi team.

A few points worth considering:

1. **Kodi's usage of `subscriptionLive`**: From what I can tell looking at the `pvr.hts` source, Kodi doesn't appear to call `subscriptionLive` at all. It seems to rely on natural catchup — resuming playback at 1x speed and letting the server's end-of-buffer detection trigger `_timeshift_flush_to_live()`. If I've missed something there, happy to be corrected.

2. **Decoder reinitialization**: The fix delivers the most recent `sstart` (stream start / codec config, including SPS/PPS for H.264/HEVC) before transitioning to `TS_LIVE`. This gives decoders a clean configuration reset — functionally the same as what happens during a channel switch or after `subscriptionSeek`.

3. **Equivalence to existing seek paths**: A normal `subscriptionSkip` to a time near the end of the buffer already results in the server seeking past most buffered content and delivering fresh packets from the new position. Our change to `SMT_SKIP_LIVE` does essentially the same thing, just without the intermediate seek step.

4. **Testing**: I've tested this with both hardware decoding (VideoToolbox / VTDecompressionSession) and software decoding (FFmpeg) on macOS, across H.264 720p and 1080i streams. Both handle the transition cleanly.

5. **Alternative: opt-in parameter for backward compatibility**: If there's concern about changing the existing behavior, one option would be to add an optional parameter to `subscriptionLive` — something like `skipBuffered` (u32, optional, default 0). Clients that don't send it get the current flush-all-packets behavior, clients that send `skipBuffered: 1` get the instant skip-to-live transition. Since HTSP messages are map-based, this would be fully backward compatible without requiring a protocol version bump. Happy to rework the PR this way if that's preferred.

DeltaMikeCharlie:
> 5\. **Alternative: opt-in parameter for backward compatibility**: If there's concern about changing the existing behavior, one option would be to add an optional parameter to `subscriptionLive` — something like `skipBuffered` (u32, optional, default 0). Clients that don't send it get the current flush-all-packets behavior, clients that send `skipBuffered: 1` get the instant skip-to-live transition. Since HTSP messages are map-based, this would be fully backward compatible without requiring a protocol version bump. Happy to rework the PR this way if that's preferred.

@oliversluke - Just a thought:  If a client application is experiencing issues with rapid buffer dumping, then opting in on the client side means that users would have to wait for client-side developers to address the issue.

Have you considered opting out on the TVH side?  Make this new behaviour the default behaviour but allow users to opt-out if unforeseen issues arise.

In `Configuration -> Stream -> Stream Profiles`, you could add a new 'HTSP Buffer Behaviour' parameter.  Users could have multiple profiles set with this feature enabled/disabled as they see fit.  One profile for client type A and another profile for client type B, etc.

There would be no need for client-side developers to change anything.

If this new feature is demonstrated to be troublesome, then the TVH default can be changed to 'Disabled'.

You could still have client-side opt-in, but then you would need to determine which setting wins.  Perhaps the new parameter options could be, 'Enabled', 'Disabled' and 'Client Selection' rather than just a boolean.

oliversluke:
@DeltaMikeCharlie Thank you for the idea. Let me look into this and get back to it shortly.

chewitt:
I'm against adding ever-more complexity and configuration options; we have way too many already. Figure out the correct future direction and implement that. If it results in some client apps needing updates, the creators of those apps will update them (or not, in which case users will find a different app).

Flole998:
I agree with @chewitt, if this is the proper way (and it certainly sounds like that) then we should move forward, even if we might break stuff on some unpopular clients. Let's just make sure that Kodi as the most popular client doesn't fully break. As Kodi relies on ffmpeg then most other clients should also be fine.

oliversluke:
Update: After extensive client-side testing, I found that the subscription​Live transition (both before and after this PR) triggers a subscription​Stop/subscription​Start cycle that can cause audio and video decoder failures on the client. This is independent from the PR. 

I've since adopted a simpler client-side approach: move back to live TV after being in time shift is just a subscriptionSkip seek to near the live edge (in my case 3s margin). The reader stays in TSS_PASS, no mode transition occurs, and audio/video resume cleanly — identical to a normal seek. This works with and without this PR.

That said, this PR is still a net improvement — it replaces the speed=100x fast-forward hack with a proper instant transition and is architecturally cleaner. I'm happy to leave it open for the community. Just wanted to flag that clients using subscription​Live should be prepared to handle the transient subscription​Stop/subscription​Start gracefully — flush and reinitialize both video and audio decoders, as the stream restart can cause issues with codec state on either side.

PR #2096, opened 2026-04-23
channels: fix reentrancy guard in channel_class_changed()

oliversluke:
## Summary

`channel_class_changed()` uses `ch_changed_ref` as a reentrancy guard to coalesce multiple property changes into a single notification. However, `atomic_dec(&ch->ch_changed_ref, 0)` at the end of the function subtracts 0 instead of 1, so the counter never resets to 0.

After the first call to `channel_class_changed()` for a given channel, all subsequent calls see `atomic_add(&ch->ch_changed_ref, 1) > 0` and return early — permanently suppressing HTSP `channelUpdate` / `channelDelete` notifications and bouquet/tag propagation for that channel.

## Root cause

In `src/channels.c` line 95:

```c
// Bug: subtracts 0, counter stays at 1 after first call
atomic_dec(&ch->ch_changed_ref, 0);

// Fix: properly decrements counter back to 0
atomic_dec(&ch->ch_changed_ref, 1);
```

## Impact

- HTSP clients with `enableAsyncMetadata` never receive `channelUpdate` after the first property change per channel (number, name, icon, etc.)
- Channel reordering via the web UI or API is invisible to connected HTSP clients
- The initial sync is unaffected (it sends channels in a direct loop, bypassing `channel_class_changed()`)

## How to reproduce

1. Connect an HTSP client with `enableAsyncMetadata`
2. Change a channel's number or name via the web UI
3. Observe that the HTSP client receives no `channelUpdate` message
4. Verified with tcpdump: zero bytes sent from TVHeadend to the HTSP client after the change

### DIFF
--- a/src/channels.c
+++ b/src/channels.c
@@ -92,7 +92,7 @@ channel_class_changed ( idnode_t *self )
   htsp_channel_update(ch);
 
 end:
-  atomic_dec(&ch->ch_changed_ref, 0);
+  atomic_dec(&ch->ch_changed_ref, 1);
 }
 
 static htsmsg_t *


DeltaMikeCharlie:
I applied this PR on a DVB-T test system and I was able to reproduce the error and verify the proposed solution.

I used Kodi and a custom Python script to monitor HTSP activity.  When changing a channel name, without the PR, no `channelUpdate` notification message were seen by the Python script and the channel name did not change in the Kodi UI.  A restart of Kodi was required to see the new channel name.  Once the PR was applied, the Python script received `channelUpdate` notifications and the Kodi UI updated immediately without restart.

@oliversluke - Nice catch.

oliversluke:
@Flole998 Hi, it looks like the "Build on ubuntu22.04 aarch64" job failed — appears to be a CI runner issue rather than a code problem (the same job passed on other recent PRs). Would it be possible to re-run it so auto-merge can go through? Thanks!

PR #2097, opened 2026-04-24
htsp: fix signalStatus for DVBv5 hardware, add scale and raw value fields

oliversluke:
## Problem

The HTSP `signalStatus` message has been silently dropping `feSNR` and `feSignal` for **all modern DVBv5 hardware** since the DVBv5 migration in 2014. This affects every tuner that reports signal values in decibel scale (which is the preferred/accurate reporting mode in DVBv5).

**Root cause** (`src/htsp_server.c`, line ~4499):

```c
if((sig->snr != -2) && (sig->snr_scale == SIGNAL_STATUS_SCALE_RELATIVE))
    htsmsg_add_u32(m, "feSNR", sig->snr);
```

The condition requires `SIGNAL_STATUS_SCALE_RELATIVE` (enum 1), but modern drivers report `SIGNAL_STATUS_SCALE_DECIBEL` (enum 2). The values are available internally (the web UI and HTTP API display them correctly via `tvh_strdupa` formatting) — they are just never serialized into the HTSP message.

**Affected hardware**: Any tuner using DVBv5 `DTV_STAT_CNR` / `DTV_STAT_SIGNAL_STRENGTH` with decibel scale — including Digital Devices, TBS, recent Silicon Labs, and most modern demodulators.

### User reports and prior discussion

This has been a known issue for years. The Kodi forum thread [**"Kodi signal strength/snr indicator in db/dBm"**](https://forum.kodi.tv/showthread.php?tid=359189) (opened Dec 2020, 51+ replies across 4 pages) documents the problem in detail:

- Multiple users report that TVHeadend's web UI shows correct signal values, but Kodi always displays **0% / 0%** for both SNR and signal strength.
- User `bluzee` ([post #19](https://forum.kodi.tv/showthread.php?tid=359189&pid=3096699#pid3096699), May 2022) traced the issue to the exact same `SIGNAL_STATUS_SCALE_RELATIVE` check in `htsp_server.c` that this PR fixes.
- @ksooo, the Kodi pvr.hts maintainer, confirmed the root cause and stated ([post #18](https://forum.kodi.tv/showthread.php?tid=359189&pid=3096507#pid3096507), May 2022):

  > *"I know pretty well how to fix this. Again, in order to fix this in Kodi, tvheadend must be changed to offer a programming interface for the dbm values. Currently it is lacking such an API. Somebody must open a feature request for tvheadend and once this got implemented we can fix this on Kodi side."*

## Fix

For `SCALE_DECIBEL` drivers, two new fields are added carrying the raw values directly from the kernel — no conversion, no loss of precision:

| Field              | Type  | Value                                    | When sent                  |
|:-------------------|:-----:|:-----------------------------------------|:---------------------------|
| `feSNR`            | u32   | 0–65535 relative (unchanged)             | `SCALE_RELATIVE` drivers   |
| `feSignal`         | u32   | 0–65535 relative (unchanged)             | `SCALE_RELATIVE` drivers   |
| `feAbsoluteSNR`    | s64   | dB × 1000 (e.g. 15800 = 15.8 dB)        | `SCALE_DECIBEL` drivers    |
| `feAbsoluteSignal` | s64   | dBm × 1000 (e.g. -36500 = −36.5 dBm)    | `SCALE_DECIBEL` drivers    |

Clients determine the type of data by checking which field is present:
- `feSNR` / `feSignal` present → relative percentage values (legacy behavior, completely unchanged)
- `feAbsoluteSNR` / `feAbsoluteSignal` present → absolute dB/dBm values from DVBv5 hardware

The two sets of fields are mutually exclusive per signal type — a given `signalStatus` message will contain either the relative or the absolute field for each metric, never both.

HTSP protocol version bumped from 43 to 44.

### GitBook HTSP specification

The following entries should be added to the `signalStatus` message in the [HTSP specification](https://docs.tvheadend.org/documentation/development/htsp/server-to-client-methods):

```
feAbsoluteSNR    s64   optional   Signal to noise ratio in dB × 1000.
                                  Present when the tuner reports in
                                  SIGNAL_STATUS_SCALE_DECIBEL.
                                  Mutually exclusive with feSNR.
                                  Added in version 44.

feAbsoluteSignal s64   optional   Signal strength in dBm × 1000.
                                  Present when the tuner reports in
                                  SIGNAL_STATUS_SCALE_DECIBEL.
                                  Mutually exclusive with feSignal.
                                  Added in version 44.
```

Happy to submit the GitBook update directly if given contributor access.

## Impact on existing clients

**Zero regression.** The existing `feSNR` / `feSignal` fields are completely unchanged for `SCALE_RELATIVE` hardware. For `SCALE_DECIBEL` hardware, these fields were never sent before (that's the bug), so no existing client depends on them. The new `feAbsoluteSNR` / `feAbsoluteSignal` fields are additive — clients that don't read them simply ignore them.

## Testing

Tested with Digital Devices MaxSX8 (DVB-S2, MxL5xx demodulator) — a pure DVBv5 driver that only reports `SIGNAL_STATUS_SCALE_DECIBEL`:

- Before: `signalStatus` contained only `feStatus`, `feBER`, `feUNC` — no signal/SNR data
- After: `feAbsoluteSNR` and `feAbsoluteSignal` populated with correct dB×1000 / dBm×1000 values matching the TVHeadend web UI
- `feSNR` / `feSignal` correctly not sent (driver does not report `SCALE_RELATIVE`)
- Build: clean compile, no warnings

### DIFF
--- a/src/htsp_server.c
+++ b/src/htsp_server.c
@@ -51,7 +51,7 @@
 
 static void *htsp_server, *htsp_server_2;
 
-#define HTSP_PROTO_VERSION 43
+#define HTSP_PROTO_VERSION 44
 
 #define HTSP_ASYNC_OFF  0x00
 #define HTSP_ASYNC_ON   0x01
@@ -4497,9 +4497,13 @@ htsp_subscription_signal_status(htsp_subscription_t *hs, signal_status_t *sig)
   if (!htsp_anonymize(hs->hs_htsp)) {
     htsmsg_add_str(m, "feStatus",   sig->status_text);
     if((sig->snr != -2) && (sig->snr_scale == SIGNAL_STATUS_SCALE_RELATIVE))
-      htsmsg_add_u32(m, "feSNR",    sig->snr);
+      htsmsg_add_u32(m, "feSNR", sig->snr);
+    else if((sig->snr != -2) && (sig->snr_scale == SIGNAL_STATUS_SCALE_DECIBEL))
+      htsmsg_add_s64(m, "feAbsoluteSNR", sig->snr);
     if((sig->signal != -2) && (sig->signal_scale == SIGNAL_STATUS_SCALE_RELATIVE))
       htsmsg_add_u32(m, "feSignal", sig->signal);
+    else if((sig->signal != -2) && (sig->signal_scale == SIGNAL_STATUS_SCALE_DECIBEL))
+      htsmsg_add_s64(m, "feAbsoluteSignal", sig->signal);
     if(sig->ber != -2)
       htsmsg_add_u32(m, "feBER",    sig->ber);
     if(sig->unc != -2)


oliversluke:
Thanks for the review. Regarding both refactoring suggestions:

The `feSNRScale`/`feSNRNum` (and `feSignalScale`/`feSignalNum`) lines are intentionally inside the switch cases. The `default` branch deliberately sends nothing for `SIGNAL_STATUS_SCALE_UNKNOWN` — if we moved the scale/num fields outside the switch, we'd need a flag to track whether the switch actually emitted `feSNR`, adding complexity. Keeping them inline makes it structurally impossible to accidentally emit scale/num metadata without the corresponding base field.

The suggestion to document the magic numbers is fair though — adding comments for the 35 dB ceiling and −90 dBm floor in a follow-up commit.

Flole998:
IMO you shouldn't do any calculations but instead provide the values to the client so that the client can do all the calculations or display the raw dB values. @ksooo as the maintainer of the Kodi addon, what do you think?

oliversluke:
Thanks for the feedback. The reason we keep the converted relative value in `feSNR`/`feSignal` is backward compatibility — these fields have always carried 0–65535 relative values, and existing clients depend on that contract. Kodi's pvr.hts, for example, passes `feSNR`/`feSignal` straight through to Kodi's signal display without any transformation. If we put raw dB×1000 values in these fields, every existing Kodi installation would show incorrect signal readings.

This server-side dB→relative conversion also isn't a new pattern — TVHeadend's own SAT>IP server does the exact same thing in `src/satip/rtp.c` (lines 709–731), converting both `SCALE_DECIBEL` and `SCALE_RELATIVE` into the SAT>IP protocol's level/quality ranges. Our HTSP change follows the same approach.

The raw dB values are available to clients that want them — that's what the new `feSNRNum`/`feSignalNum` fields are for. So clients get the best of both worlds: backward-compatible relative values in the existing fields, and exact dB/dBm readings in the new ones.

That said, happy to adjust the approach based on your and @ksooo's guidance — you know the ecosystem and client landscape best.

Flole998:
For SAT>IP that's a whole different story. The parameters are configurable and in that protocol there simply isn't a field for the raw values and we have no control over that protocol. If there is a field in the HTSP protocol we should only use that field, if there is no field we should add it. If Kodi is lacking support for using the new field that's a small fix that should happen on the Kodi side, especially since it provides more information to the user than a percentage value.

oliversluke:
Updated the PR description with additional context — there's a long-standing Kodi forum thread ([signal strength/snr indicator in db/dBm](https://forum.kodi.tv/showthread.php?tid=359189), 51+ replies since Dec 2020) where multiple users report this exact issue and @ksooo identified the root cause as the missing server-side API for dBm values. This PR addresses that directly.

ksooo:
@oliversluke I think your approach is the correct way to address the problem. Once we have teh APi extended in tvheadend, pvr.hts could be extended, maybe we also need some changes in Kodi core code (new GUI labels and changes to support both dBm and %) and finally to skins (at least Estuary).

Flole998:
I'd do it like it was proposed in the Kodi forum: Either percentage or dB values, depending on what is available, no conversion. The problem with converting to percentage values is that you have to define 100% somehow, and that is different for the transport used, and even for the modulation used (while 30 dB SNR is already a poor signal for DVB-C with QAM-256 modulation, it is alright for DVB-C with QAM-64 and it is great for DVB-S). 35dB is certainly not the ceiling for DVB-C, those values can be above 40 in production networks (and typically are) and in a lab environment you can even go above 45. For the power level a cable tuner wouldn't even lock anymore for -90 dBm, they are generally specified for -15dBm to +15dBm, but the SNR will be quite poor at -15dBm (and exceeding +15dBm can sometimes even lead to a better signal). And this mess is only for DVB-C (and slightly DVB-S), I haven't even started talking about DVB-T, DVB-T2 or ATSC yet (mostly because I don't know the usual values for those).

The current percentage conversion is certainly not correct for all transports, and I am not sure if adding a whole bunch of conversion logic is worth the effort just so that older clients will suddenly see a percentage value. I'd do the absolute values and leave it up to the client to convert them if they deem it necessary (I doubt anybody will, for the Kodi side I am quite sure that the percentage indicators will simply be replaced by text that shows the dB values when they are provided).

WIth that said, the change could be as simple as:
`htsmsg_add_u32(m, sig->signal_scale == SIGNAL_STATUS_SCALE_RELATIVE ? "feSignal" : "feAbsoluteSignal", sig->signal);`

ksooo:
> I'd do it like it was proposed in the Kodi forum: Either percentage or dB values, depending on what is available, no conversion. 

 This. If either is available. And prefer dBm over %

oliversluke:
Thanks @Flole998 and @ksooo for the quick alignment — really appreciate the guidance.

Updated the branch to follow your approach. The change is now much simpler:

- **`feSNR` / `feSignal`** (u32) — sent only for `SCALE_RELATIVE` drivers, unchanged from current behavior
- **`feAbsoluteSNR` / `feAbsoluteSignal`** (s64) — sent only for `SCALE_DECIBEL` drivers, carrying the raw dB×1000 / dBm×1000 values directly from the kernel with no conversion

No conversion logic, no scale indicator fields — clients check which field is present and display accordingly. The diff from master is now just 4 added lines.

Flole998:
Please squash you commits, it's a single change so it should be in a single commit. Other than that, looks good to me

oliversluke:
Done — squashed into a single commit. Thanks again to both of you for the fast and thoughtful feedback, it made the end result much cleaner.

ksooo:
HTSP version should be bumped in this PR and GitBook HTSP specification should be extended to document the new fields.

oliversluke:
@ksooo Done — HTSP protocol version bumped to 44 in the same commit.

For the GitBook HTSP specification, I've included the exact text for the two new `signalStatus` fields in the PR description (under "GitBook HTSP specification"), ready to be added to the [server-to-client methods](https://docs.tvheadend.org/documentation/development/htsp/server-to-client-methods) page. Happy to submit the GitBook update directly if you can add me as a contributor, or feel free to copy it in — whichever is easier for you.

Thanks again for the clear guidance on this, really appreciate how responsive you and @Flole998 have been.

Flole998:
You should be able to sign up using the link on https://docs.tvheadend.org/documentation and then it should allow you to edit it and submit a PR directly from Gitbook. If that doesn't work let me know and I'll do it.

oliversluke:
GitBook spec update submitted as change request #164. Should be all set now.

PR #2107, opened 2026-05-05
prop.c: fix const qualifier warning in prop_write_values

Flole998:


Flole998:
@copilot The logic for parsing split integers in the PT_U32 case is identical to the prop_intsplit_from_str helper function defined in src/prop.h. Instead of just fixing the const qualifier warning, consider refactoring this entire block to use the helper function, which would improve maintainability and reduce code duplication. This would also allow you to remove the local variable s entirely.

PR #2109, opened 2026-05-09
mpegts: align DVB delivery-system defaults with the enum

oliversluke:
## Summary

Several mux classes in `src/input/mpegts/mpegts_mux_dvb.c` declare a
default value that the corresponding `*_enum` callback never emits.
Each enum function iterates a small list of `DVB_*` constants and emits
the canonical string returned by the matching `dvb_*2str` helper, which
scans the relevant strtab in `dvb_support.c` and returns the first
matching row.

Two distinct mismatches:

### 1. Delivery-system defaults

Three classes used the unhyphenated spelling and one carried the wrong
constant entirely (copy-pasted from the dvbt class above — DTMB is not
DVBT):

| Class | Default before | After |
|---|---|---|
| `dvb_mux_dvbt_class` | `"DVBT"` | `"DVB-T"` |
| `dvb_mux_dvbs_class` | `"DVBS"` | `"DVB-S"` |
| `dvb_mux_isdb_s_class` | `"ISDBS"` | `"ISDB-S"` |
| `dvb_mux_dtmb_class` | `"DVBT"` | `"DTMB"` |

The other delsys defaults (DVB-C, ATSC-T, ATSC-C, ISDB-T, DAB) already
match what the enum exposes; this PR touches only the four misaligned
ones.

### 2. QAM constellation / modulation defaults

`qamtab` carries two entries for "auto":

```
{ "AUTO",     DVB_MOD_AUTO },
{ "QAM/AUTO", DVB_MOD_QAM_AUTO },
```

The enum arrays for the QAM-based modulation classes pass
`DVB_MOD_QAM_AUTO`, so the dropdown publishes `"QAM/AUTO"` — but each
class declared the default as `"AUTO"`, which parses to a different
constant (`DVB_MOD_AUTO`) and isn't a member of the published enum.

| Class | Property | Default before | After |
|---|---|---|---|
| `dvb_mux_dvbt_class` | `constellation` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_dvbc_class` | `constellation` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_atsc_t_class` | `modulation` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_atsc_c_class` | `constellation` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_isdb_t_class` | `layera_mod` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_isdb_t_class` | `layerb_mod` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_isdb_t_class` | `layerc_mod` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_isdb_c_class` | `constellation` | `"AUTO"` | `"QAM/AUTO"` |
| `dvb_mux_dtmb_class` | `constellation` | `"AUTO"` | `"QAM/AUTO"` |

The other DVB tables (`modetab`, `guardtab`, `hiertab`) only carry a
single AUTO entry, and bandwidth / FEC fall through to
`dvb_common2str` which canonicalises value 1 to `"AUTO"`. So the
remaining 22 `MUX_PROP_STR(..., N_("AUTO"))` declarations are already
in alignment; this PR touches only the nine misaligned ones.

## Effect

Any client that validates enum membership on submit (rejects values
not present in the published enum) refuses to send the default — the
user has to manually re-pick the canonical value before the form will
save. The ExtJS combobox doesn't enforce enum membership on submit,
so the bug went unnoticed there.

## Migration

The delsys side has no on-disk impact: `dvb_str2delsys` still accepts
the unhyphenated spellings AND maps them to the same `DVB_SYS_*`
constant, so existing config files continue to load unchanged.

For the QAM cases the canonical and legacy strings parse to *different*
constants (`DVB_MOD_QAM_AUTO` vs `DVB_MOD_AUTO`). Existing configs
persisted with `"AUTO"` continue to load and tune correctly — both
signal "let the driver decide" — but the loaded value isn't a member
of the published enum. UIs that synthesise a current-value option
(any sane combobox) keep the form editable until the user re-saves
with the canonical choice; UIs that don't may need a one-off re-pick
on the affected fields.

## Test plan

- [ ] `./configure && make -j$(nproc)` builds cleanly with `-Werror`.
- [ ] Delivery system: create a DVB-T network → Add mux → Delivery System dropdown shows `"DVB-T"` selected (was `"DVBT"`, not in the dropdown). Same check on DVB-S, ISDB-S, DTMB.
- [ ] Modulation: create a DVB-T / DVB-C / ATSC-T / ATSC-C / ISDB-T / ISDB-C / DTMB network → Add mux → constellation / modulation / layer A/B/C dropdown shows `"QAM/AUTO"` selected (was `"AUTO"`, not in the dropdown).
- [ ] Existing config-file load: a config persisted before this patch still loads as the same `DVB_SYS_*` / `DVB_MOD_*` constant and tunes successfully.

### DIFF
--- a/src/input/mpegts/mpegts_mux_dvb.c
+++ b/src/input/mpegts/mpegts_mux_dvb.c
@@ -208,7 +208,7 @@ const idclass_t dvb_mux_dvbt_class =
   .ic_caption    = N_("DVB-T multiplex"),
   .ic_properties = (const property_t[]){
     {
-      MUX_PROP_STR("delsys", N_("Delivery system"), dvbt, delsys, "DVBT"),
+      MUX_PROP_STR("delsys", N_("Delivery system"), dvbt, delsys, "DVB-T"),
       .desc     = N_("The delivery system the mux uses. "
                      "Make sure that your tuner supports the delivery "
                      "system selected here."),
@@ -229,7 +229,7 @@ const idclass_t dvb_mux_dvbt_class =
                      "do not like the AUTO setting."),
     },
     {
-      MUX_PROP_STR("constellation", N_("Constellation"), dvbt, qam, N_("AUTO")),
+      MUX_PROP_STR("constellation", N_("Constellation"), dvbt, qam, N_("QAM/AUTO")),
       .desc     = N_("The COFDM modulation used by the mux. "
                      "If you're not sure of the value leave as AUTO."),
     },
@@ -324,7 +324,7 @@ const idclass_t dvb_mux_dvbc_class =
       .off      = offsetof(dvb_mux_t, lm_tuning.u.dmc_fe_qam.symbol_rate),
     },
     {
-      MUX_PROP_STR("constellation", N_("Constellation"), dvbc, qam, N_("AUTO")),
+      MUX_PROP_STR("constellation", N_("Constellation"), dvbc, qam, N_("QAM/AUTO")),
       .desc     = N_("The quadrature amplitude modulation (QAM) used by the mux. "
                      "If you're not sure of the value leave as AUTO."),
     },
@@ -462,7 +462,7 @@ const idclass_t dvb_mux_dvbs_class =
   .ic_caption    = N_("DVB-S multiplex"),
   .ic_properties = (const property_t[]){
     {
-      MUX_PROP_STR("delsys", N_("Delivery system"), dvbs, delsys, "DVBS"),
+      MUX_PROP_STR("delsys", N_("Delivery system"), dvbs, delsys, "DVB-S"),
       .desc     = N_("The delivery system used by your provider."),
     },
     {
@@ -596,7 +596,7 @@ const idclass_t dvb_mux_atsc_t_class =
       .set      = dvb_mux_dvbt_class_frequency_set,
     },
     {
-      MUX_PROP_STR("modulation", N_("Modulation"), atsc_t, qam, N_("AUTO")),
+      MUX_PROP_STR("modulation", N_("Modulation"), atsc_t, qam, N_("QAM/AUTO")),
       .desc     = N_("The modulation used on the mux."),
     },
     {}
@@ -644,7 +644,7 @@ const idclass_t dvb_mux_atsc_c_class =
       .off      = offsetof(dvb_mux_t, lm_tuning.u.dmc_fe_qam.symbol_rate),
     },
     {
-      MUX_PROP_STR("constellation", N_("Constellation"), dvbc, qam, N_("AUTO")),
+      MUX_PROP_STR("constellation", N_("Constellation"), dvbc, qam, N_("QAM/AUTO")),
       .desc     = N_("The quadrature amplitude modulation (QAM) used by the mux. "
                      "If you're not sure of the value leave as AUTO."),
     },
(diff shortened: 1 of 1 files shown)


DeltaMikeCharlie:
Tested and confirmed as working on a DVB-T system.

**Before**

image


`/api/mpegts/network/mux_class` for selected network.

```json
{
	"22": {
		"id": "delsys",
		"type": "str",
		"caption": "Delivery system",
		"description": "The delivery system the mux uses. Make sure that your tuner supports the delivery system selected here.",
		"default": "DVBT",  <===== NOTE: DEFAULT NOT IN ENUM.
		"enum": [
			"DVB-T",
			"DVB-T2",
			"TURBO"
		]
	}
}
```

**After**

image

```json
{
	"22": {
		"id": "delsys",
		"type": "str",
		"caption": "Delivery system",
		"description": "The delivery system the mux uses. Make sure that your tuner supports the delivery system selected here.",
		"default": "DVB-T",  <===== DEFAULT NOW IN ENUM.
		"enum": [
			"DVB-T",
			"DVB-T2",
			"TURBO"
		]
	}
}
```
Due to the nature of the change, it is highly probable that the default values for the other identified delivery systems will also be rectified.

DeltaMikeCharlie:
I have identified the following class /property combinations where the default value (excluding blanks, zeros and none) is not present in the enum.

dvb_mux_atsc_c / constellation
dvb_mux_atsc_t / modulation
dvb_mux_dvbc / constellation
dvb_mux_dvbt / constellation
dvb_mux_isdbc / constellation
dvb_mux_isdbt / layera_mod
dvb_mux_isdbt / layerb_mod
dvb_mux_isdbt / layerc_mod

For example:

`dvb_mux_atsc_c`

```json
{
  "id": "constellation",
  "type": "str",
  "caption": "Constellation",
  "description": "The quadrature amplitude modulation (QAM) used by the mux. If you're not sure of the value leave as AUTO.",
  "default": "AUTO",  <=== "QAM/AUTO", NOT JUST "AUTO"
  "enum": [
    {
      "key": "QAM/AUTO",
      "val": "QAM/AUTO"
    },
    {
      "key": "QAM/16",
      "val": "QAM/16"
    },
    {
      "key": "QAM/32",
      "val": "QAM/32"
    },
    {
      "key": "QAM/64",
      "val": "QAM/64"
    },
    {
      "key": "QAM/128",
      "val": "QAM/128"
    },
    {
      "key": "QAM/256",
      "val": "QAM/256"
    }
  ]
}
```

There may also be a few that my script missed.  This won't stop the new UI from working, but the user will have to select a valid value from the drop-down.

oliversluke:
Thanks for the detailed sweep! I've folded all eight cases into this PR via a force-push, and an audit caught one more that the script missed — `dvb_mux_dtmb_class.constellation` had the same shape, so it's covered too. The other DVB tables (modetab / guardtab / hiertab + the bw/fec helpers) all canonicalise their AUTO consistently, so I believe nine is the complete set.

One nuance worth flagging: for the delsys side, the legacy and canonical strings parse to the same `DVB_SYS_*` constant, so persisted configs reload unchanged. For the QAM cases `"AUTO"` and `"QAM/AUTO"` parse to different constants (`DVB_MOD_AUTO` vs `DVB_MOD_QAM_AUTO`). Existing muxes saved with the legacy value continue to load and tune fine (both mean "let the driver decide"), but the loaded value isn't an enum member, so any UI that doesn't synthesise a current-value option may need a one-off re-pick. PR body has the full migration note.

PR #2112, opened 2026-05-14
idnode: strict gt/lt + new ge/le/ne filter comparators, API v20

oliversluke:
## Summary

The idnode numeric filter (the `comparison` field on
`type:"numeric"` filter entries, used by every `/api/.../grid`
endpoint) offered only `gt` / `lt` / `eq`, and `gt` / `lt` matched
*inclusively* — `gt` behaved as `>=`, `lt` as `<=` — contradicting
their names. The anomaly is documented:
https://docs.tvheadend.org/documentation/development/json-api/api-description/common-parameters

This PR tightens `gt` / `lt` to strict inequality and adds the three
missing operators:

| `comparison` | meaning |
|---|---|
| `eq` | `==` |
| `gt` | `>`  (now strict) |
| `lt` | `<`  (now strict) |
| `ge` | `>=` (new) |
| `le` | `<=` (new) |
| `ne` | `!=` (new) |

The operators apply to numeric (integer and double) filters.

## API version

`TVH_API_VERSION` is bumped 19 → 20 so clients can detect the
corrected/expanded operator set via `api/serverinfo`.

## Behaviour change

Making `gt` / `lt` strict is a breaking change for any client that
relied on the old inclusive behaviour; such callers should move to
`ge` / `le`, which preserve the previous semantics. There is no
per-request API-version negotiation, so the change applies to every
client of an upgraded server — the version bump is the detection
signal.

## Changes

- `src/idnode.h` — `IC_GE` / `IC_LE` / `IC_NE` added to the filter
  comparator enum, which is now the named `idnode_filter_comp_t`.
- `src/idnode.c` — `gt` / `lt` made strict; the comparison decision is
  a single shared helper (`idnode_filter_cmp_fails()` on a three-way
  compare) used by the string, integer, and double branches, replacing
  three copies of the operator switch.
- `src/api/api_idnode.c` — `ge` / `le` / `ne` wire strings.
- `src/api.h` — `TVH_API_VERSION` 19 → 20.
- `src/webui/static-vue/` — the Vue UI adopts the new operator set
  (see "Web UIs" below): numeric filter picker, wire mapping,
  summary chips + funnel tooltip, persisted-filter migration,
  comparator-vocabulary documentation.

## Web UIs

**The new Vue UI ships the new features in the same commit**, so
server and UI can never be inconsistent:

- The numeric column filter (any grid column with a numeric funnel,
  e.g. Channels → Number) now offers the full truthful operator set —
  `=`, `≠`, `<`, `≤`, `>`, `≥`, and `Between` — mapping 1:1 onto the
  server comparators. `Between` translates to `ge:min` + `le:max`, so
  the displayed range is inclusive at both bounds.
- Filter summary chips (View Options → Filters) and the funnel-icon
  hover render the matching glyphs (`> 5`, `≤ 10`, `5 – 10`). The
  hover previously stringified the filter model — fixed here.
- **Saved filters keep their meaning:** persisted per-grid filters are
  now versioned; filters saved before this change stored `gt`/`lt`
  under ≥/≤ labels (matching the then-inclusive server) and migrate
  to `ge`/`le` on load.
- The EPG grid endpoint (`epg/events/grid`) has its own `eq`/`gt`/`lt`
  filter vocabulary (`api_epg.c`, matched inclusively) and is
  deliberately untouched — the client senders that talk to it stay on
  `gt`/`lt`, and both vocabularies are documented at the type and call
  sites so they don't get conflated later.

**The legacy ExtJS UI needs no change** — a nice side effect: its
numeric filter menu has always displayed strict `>` / `<` / `=` icons
while the server matched inclusively. With this PR the displayed
operators become truthful for the first time.

## Test plan

- [x] Builds clean under `-Werror` (`-Wswitch` enforces that every
      `idnode_filter()` switch handles the new enum members).
- [x] `api/serverinfo` reports `api_version: 20`.
- [x] Exercised a numeric grid filter (`channel/grid` on `number`)
      with all six operators — `gt`/`lt` strict, `ge`/`le`
      inclusive, `ne` excludes only the equal row.
- [x] Vue UI: 2653 vitest cases green (new coverage for the operator
      picker, wire mapping, Between as ge+le, chip/tooltip glyphs, and
      the legacy-filter migration); browser-verified on Channels
      (strict vs inclusive differ by exactly the boundary row) and the
      EPG time-window presets (unchanged).

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -25,7 +25,8 @@
 #include "redblack.h"
 #include "access.h"
 
-#define TVH_API_VERSION 19
+/* 20: idnode numeric filter gained the ge / le / ne comparators */
+#define TVH_API_VERSION 20
 
 /*
  * Command hook
--- a/src/api/api_idnode.c
+++ b/src/api/api_idnode.c
@@ -55,7 +55,10 @@ api_idnode_flist_conf( htsmsg_t *args, const char *name )
 static struct strtab filtcmptab[] = {
   { "gt", IC_GT },
   { "lt", IC_LT },
-  { "eq", IC_EQ }
+  { "eq", IC_EQ },
+  { "ge", IC_GE },
+  { "le", IC_LE },
+  { "ne", IC_NE }
 };
 
 static void
--- a/src/idnode.c
+++ b/src/idnode.c
@@ -867,30 +867,48 @@ idnode_filter_init
   }
 }
 
+/* Shared ordered-comparison outcome: `cmp` is the three-way result
+ * (<0 / 0 / >0) of comparing the row value against the filter value;
+ * returns 1 when the row FAILS the filter. */
+static int
+idnode_filter_cmp_fails ( int cmp, idnode_filter_comp_t comp )
+{
+  switch (comp) {
+    case IC_EQ: return cmp != 0;
+    case IC_LT: return cmp >= 0;
+    case IC_GT: return cmp <= 0;
+    case IC_GE: return cmp < 0;
+    case IC_LE: return cmp > 0;
+    case IC_NE: return cmp == 0;
+    case IC_IN:
+    case IC_RE: return 0; // Note: invalid for ordered comparisons
+  }
+  return 0;
+}
+
 int
 idnode_filter
   ( idnode_t *in, idnode_filter_t *filter, const char *lang )
 {
   idnode_filter_ele_t *f;
-  
+
   LIST_FOREACH(f, filter, link) {
     if (!f->checked)
       idnode_filter_init(in, filter);
     if (f->type == IF_STR) {
       const char *str;
       char *strdisp;
-      int r = 1;
+      int r;
       str = strdisp = idnode_get_display(in, idnode_find_prop(in, f->key), lang);
       if (!str)
         if (!(str = idnode_get_str(in, f->key)))
           return 1;
-      switch(f->comp) {
-        case IC_IN: r = strstr(str, f->u.s) == NULL; break;
-        case IC_EQ: r = strcmp(str, f->u.s) != 0; break;
-        case IC_LT: r = strcmp(str, f->u.s) > 0; break;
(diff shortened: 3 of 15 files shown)


Flole998:
I think we can merge this as-is, I'd like to hear the opinion from others though :)

DeltaMikeCharlie:
This logical anomaly has already been documented:

https://docs.tvheadend.org/documentation/development/json-api/api-description/common-parameters

`The "comparison" field is only used with numeric data. A "gt" comparison actually matches on greater-or-equal, and "lt" on less-or-equal.`

Should we introduce `le` and `ge` to maintain existing functionality?  Or leave the existing anomalous functionality as it is and introduce new accurate operators?  Perhaps introduce `<`, `>`, `<=` and '>=` that operate correctly and leave the others for backwards compatibility.

dave-p:
> Perhaps introduce <, >, <= and '>=` that operate correctly and leave the others for backwards compatibility.

This non-breaking change is perhaps the best option.

DeltaMikeCharlie:
The API version number should probably also be incremented.

https://docs.tvheadend.org/documentation/development/json-api/api-description/config#serverinfo

https://github.com/tvheadend/tvheadend/blob/ae37453bcc51ccb4b500f3b2e08b56763c189575/src/api.h#L28

Flole998:
/gemini What do you think about this change?

dave-p:
Looks good to me.

oliversluke:
Heads-up on the latest push: the PR now also ships the client side — the Vue UI's numeric column filter offers the full operator set (`= ≠ < ≤ > ≥ Between`) mapped 1:1 onto the new comparators, with saved filters migrating so they keep the meaning they were created with. Description updated with the details. The legacy ExtJS UI intentionally gets no change: its filter menu has always shown strict `>`/`<` icons, so the strict server semantics simply make the legacy display truthful.

PR #2120, opened 2026-05-22
epggrab: per-service EIT processing policy with global default

oliversluke:
## Update (2026-05-23) — reworked to per-service enum + global default

Reworked per the discussion above. Two settings instead of one:

- **Per-service "EIT processing"** (replaces the legacy "Ignore EPG (EIT)" toggle on each DVB service)
- **Global "EIT processing (default)"** (under EPG Grabber → OTA settings)

Per-service values:

| Value | Behaviour |
|---|---|
| **Default** | Defer to the global default |
| **None** | Ignore EIT for this service (= old `Ignore EPG (EIT) = true`) |
| **Actual only** | Accept only actual-TS sub-tables (0x4e, 0x50–0x5f) |
| **Other only** | Accept only other-TS sub-tables (0x4f, 0x60–0x6f) |
| **Either** | Accept both, no precedence — today's behaviour |
| **Adaptive** | The original v1 behaviour: accept both, but once a service's own actual-TS schedule has been seen, drop further other-TS for it |

Global default exposes the same enum minus Default. New installs default to **Either** — no behaviour change from today, no surprise migrations.

Legacy migration on first load: `dvb_ignore_eit=true` → **None** (same observable behaviour), `dvb_ignore_eit=false` (the vast majority) → **Default** (defers to the global, also same observable behaviour while the global stays Either). The old key disappears from the config on next save.

The motivation, validation, prior report, peer-project comparison, and standards basis sections below are unchanged. The "Opt-in / compatibility", "The change", and "Testing" sections below are updated to reflect this shape.

---

## The problem

DVB EIT (Event Information Table) is broadcast in two forms, distinguished by `table_id`:

- **actual-TS** — `0x4E` (present/following) and `0x50–0x5F` (schedule): the EPG of the transport stream that carries the table. A service describing *itself*.
- **other-TS** — `0x4F` (present/following) and `0x60–0x6F` (schedule): EPG for services on a *different* transport stream — a second-hand description, carried so a receiver tuned to one mux can show EPG for others.

tvheadend's OTA grabber tunes every multiplex, so on a typical installation it ingests both the actual-TS EIT (from a service's own transponder) and any other-TS EIT for that same service (from neighbouring transponders). `_eit_callback()` resolves both to the same service and stores whatever was parsed last — **last-writer-wins, with no precedence between the two sub-tables.**

When the two disagree, the EPG **intermittently flips** between the service's correct, detailed, show-by-show schedule and whatever coarse description a neighbouring transponder happens to carry. A common failure mode: a service's own transponder carries a full schedule, while an unrelated platform sharing the same orbital position carries a single multi-hour channel-level placeholder for that service via other-TS — and the two overwrite each other on every grab cycle.

## Validation

This was reproduced and traced on a **live European DVB-S2 installation receiving Astra 19.2°E**:

- A free-to-air channel's EPG was observed flipping between its detailed schedule and a single **24-hour placeholder event**.
- Instrumenting the EIT handler showed the placeholder arriving as an **other-TS present/following event (`table_id 0x4F`)**, 24 hours long, with descriptors tagged language `spa` and rating country `ESP` — broadcast by a Spanish-language DTH platform that shares the 19.2°E position.
- The placeholder is **cross-carried by several transponders of that platform** — not a single rogue mux. The affected channel's own transponder, meanwhile, carries a correct detailed actual-TS schedule for it.
- With the option in this PR enabled, the other-TS placeholder is dropped as soon as the channel's own actual-TS schedule has been seen, and the detailed EPG stops being overwritten.

## Prior report

This exact situation was reported previously on the tvheadend issue tracker — **Bug #6249**, full URL:

> https://old.tvheadend.org/issues/6249

(EIT recorded off a Spanish pay-TV platform's transponders carrying another network's free-to-air channel, with a coarse Spanish-language placeholder.) That report was closed as *Invalid*, on the basis that tvheadend correctly follows the DVB specification for service identification (services keyed on original_network_id / transport_stream_id / service_id).

This PR does **not** dispute that. Service identification is unchanged. The change here is narrower and separate: an *optional* precedence between two spec-distinct EIT sub-tables describing the *same* service — a question the specification leaves to the receiver (see "Standards basis" below) — and it is opt-in, so it changes nothing unless an operator chooses to enable it.

## How other DVB projects handle this

Every comparable open-source DVB project already treats other-TS EIT as a distinct, lower-trust category rather than as interchangeable with actual-TS:

- **VDR** — drops EIT `0x4F` (present/following other-TS) entirely, and for schedule tables suppresses an incoming `0x60–0x6F` section once actual-TS schedule data has been seen for that schedule (`cSchedule::OnActualTp()` in [`eit.c`](https://github.com/vdr-projects/vdr/blob/master/eit.c)).
- **Enigma2** — tags every cached event with its source and ranks actual-TS sources above their other-TS counterparts; a lower-priority other-TS write is rejected when a higher-priority actual-TS event is already stored (`eEPGCache::sectionRead()` in [`epgcache.cpp`](https://github.com/openatv/enigma2/blob/master/lib/dvb/epgcache.cpp)).
- **MythTV** — splits the two by `table_id`, anchoring actual-TS to the tuned multiplex while resolving other-TS via the full onid/tsid/sid triple (`EITHelper::AddEIT()` in [`eithelper.cpp`](https://github.com/MythTV/mythtv/blob/master/mythtv/libs/libmythtv/eithelper.cpp)).

tvheadend is currently the outlier: it resolves other-TS to any enabled mux's service and lets it overwrite unconditionally.

## Why this PR mirrors VDR's approach

Of the three models, this PR adopts **VDR's** — an adaptive, per-service rule that keys on *data actually received*:

- **VDR's model is adaptive, not configuration-based.** It suppresses other-TS for a service only once that service's *own* actual-TS schedule has genuinely been observed. A service that never broadcasts its own schedule — e.g. one whose EPG is legitimately distributed by a dedicated EPG multiplex via other-TS — is never affected. The rule is self-correcting and safe for every network topology, with no per-network configuration to get wrong.
- **Enigma2's model** — a source-priority tag persisted on every cached event — would, in tvheadend, require a new field on the core `epg_broadcast_t`, a change to the on-disk EPG database format, and a project-wide priority ordering across *all* grabbers (EIT, OpenTV, XMLTV, …). That is a far larger and more invasive change than the problem requires, and is deliberately out of scope.
- **MythTV's model** is the weakest precedent (no per-event arbitration) but confirms the principle.

This PR implements the VDR model, contained entirely within the EIT grabber:

- a runtime, per-service flag (`mpegts_service_t::s_dvb_eit_actual_seen`) is set the first time that service's actual-TS *schedule* EIT (`0x50–0x5F`) is parsed;
- once set, other-TS EIT (`0x4F` and `0x60–0x6F`) for that service is dropped in `_eit_callback()` before it can be stored.

No change to the EPG core and no change to the on-disk EPG database format. The flag is runtime-only and relearned each session.

## Opt-in / compatibility

Existing behaviour is preserved by default:

- **Fresh installs**: global default = **Either**. Identical to today — both EIT sub-tables accepted, no precedence between them.
- **Existing installs with services on `Ignore EPG (EIT) = true`**: migrate to per-service **None** on first load. Same observable behaviour as before (EIT remains ignored for those services).
- **Existing installs with services on `Ignore EPG (EIT) = false`** (the vast majority): migrate to per-service **Default**. With the global at Either, behaviour is unchanged. Operators can then opt into Adaptive (or any other policy) via a single global change without touching individual services.

The legacy `dvb_ignore_eit` key is read on first load and silently dropped from the config on next save (no matching prop on the class).

## Standards basis — why this is not a specification violation

The relevant DVB documents — **ETSI EN 300 468** (Specification for Service Information) and **ETSI TS 101 211** (Guidelines on implementation and usage of SI) — are *transmission* specifications: they define how broadcasters encode SI and how a receiver decodes the bitstream.

1. **The standard itself treats actual-TS and other-TS EIT as distinct.** EN 300 468 allocates them *separate `table_id` values* (`0x4E`/`0x50–0x5F` vs `0x4F`/`0x60–0x6F`). They form independent sub-tables, versioned independently. The distinction — "this transport stream describing itself" vs "describing another" — is encoded in the standard by design.
2. **No clause governs receiver conflict-resolution between them.** Neither EN 300 468 nor TS 101 211 contains a normative requirement specifying how a receiver must build its EPG when an actual-TS sub-table and an other-TS sub-table describe the same event differently. EPG-merge / conflict-resolution policy is outside the scope of these documents — it is receiver implementation behaviour. TS 101 211 makes EIT present/following *actual* mandatory while EIT schedule and EIT other-TS are optional, tiering the two by guaranteed availability, but it still mandates no precedence.

Consequently there is **no clause for this change to violate.** Preferring a service's own actual-TS description over a second-hand other-TS one, for a service the receiver demonstrably receives directly, is a permitted implementation choice in an area the specification deliberately leaves open. If anything, tvheadend's *current* last-writer-wins behaviour is the one that discards a distinction the standard explicitly encodes (the separate `table_id` allocation); this change makes tvheadend honour that distinction rather than flatten it — and only when the operator opts in.

## The change

A single commit:

- `src/epggrab.h` — adds the `EIT_PROCESSING_*` enum values (Default, None, Actual only, Other only, Either, Adaptive) and renames the global config field to `eit_processing_default`.
- `src/epggrab.c` — replaces the global `eit_actual_priority` boolean with the `eit_processing_default` PT_INT enum (5 values, no Default) and its `.list` callback. Fresh-install default is `EIT_PROCESSING_EITHER`.
- `src/input/mpegts.h` — renames the per-service field `s_dvb_ignore_eit` to `s_dvb_eit_processing` (now an int enum). Keeps the runtime `s_dvb_eit_actual_seen` flag, used by the Adaptive policy.
- `src/input/mpegts/mpegts_service.c` — replaces the per-service "Ignore EPG (EIT)" boolean prop with the "EIT processing" PT_INT enum (6 values, includes Default). Adds an `ic_load` hook that migrates the legacy `dvb_ignore_eit` key once on first load.
- `src/epggrab/module/eit.c` — resolves the effective policy per call (per-service unless Default, in which case the global default applies) and switches on it: drop everything for None; drop other-TS for Actual only; drop actual-TS for Other only; the original adaptive logic for Adaptive; no filter for Either.

~150 lines (incl. comments + property tables); no EPG-core or database-format changes.

## Testing

There is no automated test suite for this area; verification is by build and runtime observation.

- Builds cleanly under `-Werror -Wmissing-prototypes`.
- Verified on the same DVB-S2 / 19.2°E installation that motivated the original PR. The **Adaptive** policy reproduces v1's observed behaviour: the 24-hour other-TS placeholder is dropped once the affected channel's own actual-TS schedule is seen, and the channel retains its detailed schedule. The **Other only** policy, applied to a single service, was used as the inverse test — that service's EPG converges to the placeholder while every other service (on Default → global Either) remains on its detailed schedule, confirming the per-service override path.
- A reviewer can verify by setting **EIT processing (default)** to Adaptive (under EPG Grabber → OTA settings), running an OTA EPG grab with `--trace tbl-eit --stderr`, and observing `actual-TS schedule seen` lines for each service as their own schedule arrives, followed by `skip other-TS tid 0x... actual-TS seen` lines when a neighbour's other-TS for them is dropped.

### DIFF
--- a/docs/property/eit_processing.md
+++ b/docs/property/eit_processing.md
@@ -1,0 +1,51 @@
+
+
+This setting controls which Event Information Table (EIT) sub-tables
+are accepted when building the EPG for this service.
+
+DVB EIT is broadcast in two forms, distinguished by `table_id`:
+
+- **Actual transport stream EIT** (`0x4e` / `0x50–0x5f`) — broadcast
+  by the service's *own* multiplex. Describes the service itself.
+- **Other transport stream EIT** (`0x4f` / `0x60–0x6f`) — broadcast
+  by *another* multiplex describing this one. Often coarser; mostly
+  used so a receiver tuned to one multiplex can still show some EPG
+  for services on neighbouring multiplexes.
+
+Without a precedence rule, the two sources can overwrite each other
+on every grab cycle, making the EPG flip between a detailed
+schedule (from actual-TS) and a coarser placeholder (from other-TS).
+
+| Value | Behaviour |
+| --- | --- |
+| **Default** | Defer to the global *EIT processing (default)* setting under EPG Grabber → OTA. Use this for most services. |
+| **None** | Ignore all EIT for this service. Equivalent to the legacy *Ignore EPG (EIT)* toggle. |
+| **Actual transport stream only** | Accept only actual-TS sub-tables (`0x4e`, `0x50–0x5f`). |
+| **Other transport stream only** | Accept only other-TS sub-tables (`0x4f`, `0x60–0x6f`). |
+| **Either** | Accept both, no precedence. Today's default behaviour — what tvheadend has always done. |
+| **Adaptive** | Accept both, but once this service's own actual-TS schedule has been seen this session, drop further other-TS for it. A neighbour's coarse description cannot overwrite the service's detailed schedule, while services whose actual-TS is never received keep using other-TS (so a dedicated EPG multiplex still works). |
+
+The Adaptive policy is self-correcting per service: it suppresses
+other-TS only once detailed actual-TS data has actually been
+received. Inspired by VDR's adaptive EIT handling.
+
+**One operational caveat for Adaptive.** The "actual-TS observed"
+state is per session and resets when tvheadend restarts. For a
+short window at startup, other-TS may briefly overwrite the
+detailed EPG until the service's own actual-TS schedule is re-
+observed in the new session. DVR entries whose link to an EPG
+event was made before the restart are reconnected by the existing
+fuzzy-match fallback in `dvr_entry_fuzzy_match` once actual-TS
+data returns. The one remaining edge is a recording already in
+progress at restart that happens to receive a stop-time-extension
+update during this swap window — the extension lands on the
+placeholder rather than the resumed entry, so the recording stops
+at its originally-scheduled time. Across normal operation (no
+restart), Adaptive self-heals after the first actual-TS arrival
+and the placeholder is then permanently filtered for that service.
+
+Most installations should leave this at *Default* and pick the
+policy globally. Per-service overrides are useful for diagnosing or
+working around broadcaster-specific quirks — a service whose own
+actual-TS EIT is intermittent, or one where you trust the
+neighbour's other-TS more than the service's own.
--- a/docs/property/eit_processing_default.md
+++ b/docs/property/eit_processing_default.md
@@ -1,0 +1,46 @@
+
+
+This setting is the default EIT (Event Information Table)
+processing policy applied to every DVB service whose own *EIT
+processing* is left at *Default*.
+
+DVB EIT is broadcast in two forms, distinguished by `table_id`:
+
+- **Actual transport stream EIT** (`0x4e` / `0x50–0x5f`) — broadcast
+  by the service's *own* multiplex. Describes the service itself.
+- **Other transport stream EIT** (`0x4f` / `0x60–0x6f`) — broadcast
+  by *another* multiplex describing this one. Often coarser.
+
+Without a precedence rule, the two sources can overwrite each other
+on every grab cycle, making the EPG flip between a detailed
+schedule (from actual-TS) and a coarser placeholder (from other-TS).
+
+| Value | Behaviour |
+| --- | --- |
+| **None** | Ignore all EIT for services set to Default. |
+| **Actual transport stream only** | Accept only actual-TS sub-tables (`0x4e`, `0x50–0x5f`). |
+| **Other transport stream only** | Accept only other-TS sub-tables (`0x4f`, `0x60–0x6f`). |
+| **Either** | Accept both, no precedence. Today's default behaviour — what tvheadend has always done. |
+| **Adaptive** | Accept both, but once a service's own actual-TS schedule has been seen this session, drop further other-TS for it. A neighbouring multiplex's coarse description cannot overwrite the service's detailed schedule. Services whose actual-TS is never received keep using other-TS, so a dedicated EPG multiplex still works. |
+
+Fresh installations default to **Either**, which preserves the
+historical behaviour. Switch to **Adaptive** to opt into the
+actual-TS-precedence heuristic without configuring services
+individually.
+
+**One operational caveat for Adaptive.** The "actual-TS observed"
+state is per session and resets when tvheadend restarts. For a
+short window at startup, other-TS may briefly overwrite the
+detailed EPG until each service's own actual-TS schedule is re-
+observed in the new session. DVR entries whose link to an EPG
+event was made before the restart are reconnected by the existing
+fuzzy-match fallback in `dvr_entry_fuzzy_match` once actual-TS
+data returns. The one remaining edge is a recording already in
+progress at restart that happens to receive a stop-time-extension
+update during this swap window — the extension lands on the
+placeholder rather than the resumed entry, so the recording stops
+at its originally-scheduled time. Across normal operation (no
+restart), Adaptive self-heals after the first actual-TS arrival.
+
+Any service can override this global default via its own *EIT
+processing* setting.
--- a/src/epggrab.c
+++ b/src/epggrab.c
@@ -333,9 +333,25 @@ epggrab_class_ota_genre_translation_notify(void *self, const char *lang)
   epggrab_ota_set_genre_translation();
 }
 
+static htsmsg_t *
+epggrab_class_eit_processing_default_list(void *o, const char *lang)
+{
+  /* Per-service EIT_PROCESSING_DEFAULT defers here, so the global
+   * choice list omits Default itself. */
+  static const struct strtab tab[] = {
+    { N_("None"),                         EIT_PROCESSING_NONE        },
+    { N_("Actual transport stream only"), EIT_PROCESSING_ACTUAL_ONLY },
+    { N_("Other transport stream only"),  EIT_PROCESSING_OTHER_ONLY  },
+    { N_("Either"),                       EIT_PROCESSING_EITHER      },
+    { N_("Adaptive"),                     EIT_PROCESSING_ADAPTIVE    },
+  };
+  return strtab2htsmsg(tab, 1, lang);
+}
+
 CLASS_DOC(epgconf)
 PROP_DOC(cron)
 PROP_DOC(ota_genre_translation)
+PROP_DOC(eit_processing_default)
 
 const idclass_t epggrab_class = {
   .ic_snode      = &epggrab_conf.idnode,
@@ -467,6 +483,20 @@ const idclass_t epggrab_class = {
       .group  = 3,
     },
     {
+      .type   = PT_INT,
+      .id     = "eit_processing_default",
+      .name   = N_("EIT processing (default)"),
+      .desc   = N_("Default EIT processing policy applied to "
+                   "services whose own \"EIT processing\" is set "
+                   "to Default. See Help for the full policy "
+                   "descriptions."),
+      .doc    = prop_doc_eit_processing_default,
+      .off    = offsetof(epggrab_conf_t, eit_processing_default),
+      .opts   = PO_ADVANCED | PO_DOC_NLIST,
+      .list   = epggrab_class_eit_processing_default_list,
+      .group  = 3,
+    },
+    {
       .type   = PT_STR,
       .id     = "ota_cron",
       .name   = N_("Over-the-air Cron multi-line"),
@@ -587,6 +617,7 @@ void epggrab_init ( void )
   epggrab_conf.epgdb_periodicsave = 0;
   epggrab_conf.epgdb_saveafterimport = 0;
   epggrab_conf.epgdb_processparentallabels = 0;
+  epggrab_conf.eit_processing_default = EIT_PROCESSING_EITHER;
 
   epggrab_cron_multi              = NULL;
 
--- a/src/epggrab.h
+++ b/src/epggrab.h
@@ -317,6 +317,22 @@ struct epggrab_module_ota_scraper
 /*
  *
  */
+/*
+ * EIT processing policy. Per-service value selects which EIT sub-tables
+ * are accepted for that service; the per-service Default value defers
+ * to the global default (which therefore omits Default from its choice
+ * list). Adaptive accepts both actual- and other-TS but, once the
+ * service's own actual-TS schedule has been received, drops further
+ * other-TS for it so a neighbouring multiplex's coarse description
+ * cannot overwrite the service's detailed schedule.
+ */
+#define EIT_PROCESSING_DEFAULT       0   /* per-service only: use global */
+#define EIT_PROCESSING_NONE          1
+#define EIT_PROCESSING_ACTUAL_ONLY   2
+#define EIT_PROCESSING_OTHER_ONLY    3
+#define EIT_PROCESSING_EITHER        4
+#define EIT_PROCESSING_ADAPTIVE      5
+
 typedef struct epggrab_conf {
   idnode_t              idnode;
   char                 *cron;
@@ -330,6 +346,7 @@ typedef struct epggrab_conf {
   char                 *ota_genre_translation;
   uint32_t              ota_timeout;
   uint32_t              ota_initial;
+  uint32_t              eit_processing_default;
   uint32_t              int_initial;
 } epggrab_conf_t;
 
--- a/src/epggrab/module/eit.c
+++ b/src/epggrab/module/eit.c
@@ -1285,8 +1285,54 @@ svc_ok:
   if (!LIST_FIRST(&svc->s_channels))
     goto done;
 
-  if (svc->s_dvb_ignore_eit)
+  /* Apply the effective EIT processing policy for this service:
+   * the per-service value, falling back to the global default when
+   * the service is set to Default. Tableid ranges in this callback
+   * (see the bounds check earlier in this function) are
+   *   0x4e               actual-TS present/following
+   *   0x50-0x5f          actual-TS schedule
+   *   0x4f, 0x60-0x6f    other-TS present/following + schedule. */
+  int processing = svc->s_dvb_eit_processing;
+  if (processing == EIT_PROCESSING_DEFAULT)
+    processing = epggrab_conf.eit_processing_default;
+
+  switch (processing) {
+  case EIT_PROCESSING_NONE:
     goto done;
+  case EIT_PROCESSING_ACTUAL_ONLY:
+    if (tableid == 0x4f || tableid >= 0x60)
+      goto done;
+    break;
+  case EIT_PROCESSING_OTHER_ONLY:
+    if (tableid == 0x4e || (tableid >= 0x50 && tableid < 0x60))
+      goto done;
+    break;
+  case EIT_PROCESSING_ADAPTIVE:
+    /* Track actual-TS schedule arrival per service. Once seen,
+     * drop further other-TS for it so a neighbouring multiplex's
+     * coarse description cannot overwrite the service's own
+     * detailed schedule. Services whose actual-TS is never
+     * received keep using other-TS, so a dedicated EPG multiplex
+     * still works. P/F (0x4e / 0x4f) is only two events per
+     * service and is not used as the "actual-TS seen" trigger. */
+    if (tableid >= 0x50 && tableid < 0x60) {
+      if (!svc->s_dvb_eit_actual_seen) {
+        svc->s_dvb_eit_actual_seen = 1;
+        tvhtrace(LS_TBL_EIT, "%s: %s actual-TS schedule seen",
+                 mt->mt_name, svc->s_nicename);
+      }
+    } else if ((tableid == 0x4f || tableid >= 0x60) &&
+               svc->s_dvb_eit_actual_seen) {
+      tvhtrace(LS_TBL_EIT,
+               "%s: skip other-TS tid 0x%02X for %s, actual-TS seen",
+               mt->mt_name, tableid, svc->s_nicename);
+      goto done;
+    }
+    break;
+  case EIT_PROCESSING_EITHER:
+  default:
+    break;
+  }
 
   /* Queue events */
   len -= 11;
--- a/src/input/mpegts.h
+++ b/src/input/mpegts.h
@@ -578,7 +578,7 @@ struct mpegts_service
   char    *s_dvb_provider;
   char    *s_dvb_cridauth;
   uint16_t s_dvb_servicetype;
-  int      s_dvb_ignore_eit;
+  int      s_dvb_eit_processing;            //EIT processing policy (EIT_PROCESSING_* in epggrab.h); Default defers to the global setting
   int      s_dvb_subtitle_processing;       //Various options for replacing/augmenting the desc from the sub-title
   int      s_dvb_ignore_matching_subtitle;  //Ignore the sub-title if same as title
   char    *s_dvb_charset;
@@ -593,6 +593,11 @@ struct mpegts_service
    */
 
   int      s_dvb_eit_enable;
+  /* Runtime flag: this service's own actual-TS schedule EIT
+   * (table_id 0x50-0x5f) has been received this session. Used by
+   * the EIT_PROCESSING_ADAPTIVE policy to start dropping other-TS
+   * once detailed actual-TS data is available. Not persisted. */
+  int      s_dvb_eit_actual_seen;
   uint64_t s_dvb_opentv_chnum;
   uint16_t s_dvb_opentv_id;
   uint16_t s_atsc_source_id;
--- a/src/input/mpegts/mpegts_service.c
+++ b/src/input/mpegts/mpegts_service.c
@@ -94,7 +94,42 @@ mpegts_service_subtitle_procesing ( void *o, const char *lang )
    return strtab2htsmsg(tab, 1, lang);
 }
 
+static htsmsg_t *
+mpegts_service_eit_processing_list ( void *o, const char *lang )
+{
+  static const struct strtab tab[] = {
+    { N_("Default (use global setting)"), EIT_PROCESSING_DEFAULT     },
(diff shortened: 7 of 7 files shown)


DeltaMikeCharlie:
@oliversluke - Having this setting off by default is a wise precaution.

**Discussion**

I think that making this setting global to the whole grabber could be dangerous in some situations.  For example:  If an installation has a mix of some multiplexes carrying 'other' data and 'actual' data with some 'other' data being more suitable than the 'actual' data.

Moving the setting to the multiplex would allow for more granular control.

As the EPG is ultimately managed at the channel level and a channel can be sourced from multiple services, perhaps at the channel level, there could be a setting that controls the exact EPG source for the channel as can already be configured for XMLTV EPG sources.

Another option could be to implement this feature as an `EIT_HACK_`.  This would allow the creation of a new grabber specifically for this purpose.  In the multiplex settings, that new grabber could be selected.

**Conclusion**

Each service also already has an `Ignore EPG (EIT):` setting.

Perhaps an "Ignore 'Other' EPG" flag should be applied at the service level.  This would maintain configuration consistency and provide enough granularity to manage the EPG source relayed to the channel(s).  The trade off would be that each individual service would need to be configured rather than a single global or multiplex setting.

oliversluke:
Thanks for the detailed feedback — and agreed, default-off felt like the right call.

One thing worth spelling out about how the mechanism works, since it speaks directly to the concern about a global setting: the option enables an *adaptive, per-service* rule. tvheadend suppresses other-TS for a service only once it has actually received that service's own actual-TS schedule (table_id 0x50–0x5f). A service whose transport stream never carries its own schedule — e.g. one served by a dedicated EPG multiplex — never trips the flag, and its other-TS is left untouched. So although the *setting* is global, the *behaviour* is already decided per service, and it can't strand a service.

That said, your edge case is real: a service that does carry its own actual-TS, but where a neighbour's other-TS is the better data — the global switch has no per-entity override for that.

If we add such an override, I think your conclusion — a per-service flag, consistent with the existing `Ignore EPG (EIT)` — is the right axis. The other options are a less natural fit:

- **Per-multiplex** — the conflict is per-service, and in this case the placeholder is cross-carried by several of the neighbouring platform's transponders, so it would mean configuring many muxes for what is really one decision.
- **A per-channel EPG-source selector** — arguably the cleanest long-term home (the channel is where EPG is consumed), but a sizeable feature in its own right.
- **An `EIT_HACK_` / dedicated module** — also lands at multiplex granularity, with the same fit issue.

So the open question is really just timing — add the per-service override now, or as a follow-up. A few things bear on that:

- Every case reported so far — this one and the earlier #6249 — is fully resolved by the global switch; the reverse case (actual-TS present, other-TS better) hasn't appeared in a report yet. And a service in that reverse case is already subject to the same last-writer-wins flipping today, so enabling the switch would make its EPG deterministic rather than intermittent.
- The setting is opt-in and off by default, so it only affects someone who deliberately enables it, and if it ever misjudged a service for them, switching it back off is the immediate remedy.
- For comparison, none of VDR, Enigma2 or MythTV exposes a user setting to steer EIT actual-vs-other-TS precedence — all three resolve it automatically in their EIT-parsing code. (MythTV and Enigma2 do offer a blanket per-channel/per-service EIT on/off, but neither distinguishes actual-TS from other-TS.)

Two reasonable paths:

1. Merge the focused version as-is (global, opt-in, default-off), and add the per-service override as a follow-up if reports of the reverse case come in.
2. Add the per-service opt-out in this PR now — a flag alongside `Ignore EPG (EIT)`, e.g. "Keep other-TS EPG for this service," with the global toggle staying as the master enable and the adaptive behaviour intact.

Happy to go either way — which would you prefer?

DeltaMikeCharlie:
> Happy to go either way — which would you prefer?
>
Overnight, I thought of an alternate solution:  For the Service Class, replace the existing `Ignore EPG (EIT)` setting with a new `EIT Processing Method` (or whatever) setting.

The proposed new setting can have 5 possible values:

0 - 'None' - Ignore all EIT for this service.  This is equivalent to the old `true`.
1 - 'Actual Only' - Only process EIT from the 'Actual' table.
2 - 'Other Only' - Only process EIT from the 'Other' table.
3 - 'Either' - Use the EIT from either table.  This is the default and equivalent to the old `false`.
15 - 'Not Set' - During load time, look for the old setting and use that, otherwise, use the default.

To allow for transparent conversion of existing configurations, 'Not Set' is a transient state between initialising the property and loading the service configuration file and should never be saved.  If no value is present in the configuration file, the old setting or default value is used and the configuration file is updated.  If a value is present in the configuration file, that value is used and the old setting is ignored.

The ENUM values were selected to allow bitwise comparison when deciding which table data type to accept.  If a value of `15` does manage to make its to downstream functions, it will result in default functionality.

oliversluke:
@DeltaMikeCharlie Great proposal — turning `Ignore EPG (EIT)` into a per-service enum is exactly the right shape. I've put a polish on it to keep both your per-service granularity and a one-switch usability path; happy to land it as your model with the addition below.

**High-level idea**

Two coupled changes:

1. **Per-service** — replace the existing `Ignore EPG (EIT)` boolean with a richer **EIT processing** enum (your proposal, kept).
2. **Global** — add a sibling **EIT processing default** enum at the EPG-grabber level. Per-service entries set to "use the default" inherit from it.

One config knob (the enum), used at two levels, with a "use the default" value on the per-service side that wires the two together. Standard inherit-from-parent pattern.

**The enum**

(Used at both levels; the global level offers the same values minus `Default` — that would be circular.)

| Value | Meaning |
|---|---|
| `Default` | Use the global EIT processing setting. *Per-service only.* |
| `None` | Ignore all EIT for this service. Equivalent to the old `Ignore EPG (EIT) == true`. |
| `Actual Only` | Use only the service's own actual-TS EIT (table id `0x4e`, `0x50-0x5f`). |
| `Other Only` | Use only other-TS EIT (table id `0x4f`, `0x60-0x6f`). |
| `Either` | Use whichever arrives — today's last-writer-wins behaviour. |
| `Adaptive` | Use other-TS until the service's own actual-TS schedule (`0x50-0x5f`) is seen, then drop other-TS for that service. (This is what the current PR's runtime flag already implements — just exposed as a first-class option.) |

`Adaptive` at the global level means "apply Adaptive to every service whose per-service value is `Default`." Same logic; different scope.

**For new installations**

- Out of the box: global = `Either`, every service = `Default`. Behaviour identical to today.
- To fix the bug everywhere: flip global to `Adaptive`. No per-service config needed.
- To override one service: set its enum explicitly.

**For existing installations (migration on first load)**

A clean one-shot translation — no transient sentinel value needed:

| Old `Ignore EPG (EIT)` | New per-service value |
|---|---|
| `true` | `None` |
| `false` *(or absent)* | `Default` |

The new global setting initialises to `Either`, so the install behaves identically to before. Everything new (`Adaptive`, the explicit per-service values) is opt-in.

**Net effect**

- No upgrade surprise — default-off semantics preserved.
- Common case (hit the bug, want it fixed) → one global flip to `Adaptive` → every affected service self-resolves.
- Edge cases (this service should keep other-TS, use only one source, ignore all EIT) → that service's enum, explicit and discoverable.
- One config knob replaces the old boolean cleanly; no parallel fields to maintain.

Sound good? If so I'll rework the PR to this shape.

DeltaMikeCharlie:
> Sound good? If so I'll rework the PR to this shape.
>
That all sounds fine.

It would be nice to be able to apply these rules at the multiplex level to eliminate the need to edit multiple services on a mux if required.  However, I see this as being a relatively rare configuration operation so updating multiple services may be a pain, but it should only need to be done very infrequently.  Also, filtering services by mux with multi-select and multi-update in the WebUI could simplify the operation.

I wonder if there could be a way to automatically evaluate the quality of the competing EPG tables?  Perhaps the number of entries over a given time period or the relative length or presence of description strings could be used as a proxy for quality.  This could get overly complicated, but an interesting thought experiment.

Also, please keep in mind that what I write here are only my personal opinions.

oliversluke:
Hey @DeltaMikeCharlie — reworked per our discussion above and force-pushed.

What landed:

- **Per-service "EIT processing"** (replaces "Ignore EPG (EIT)"): Default / None / Actual only / Other only / Either / Adaptive.
- **Global "EIT processing (default)"**: same enum minus Default. Fresh installs get Either.
- One-shot migration from the old `Ignore EPG (EIT)` boolean: true → None, false → Default. Old key drops on next save.
- The original v1 adaptive heuristic is now the **Adaptive** value of the same enum, so operators can pick it per-service or globally.

Verified on the original DVB-S2 / 19.2°E setup: Adaptive global reproduces v1's behaviour (placeholder dropped once the affected channel's own actual-TS schedule is seen). Other-only as a per-service override on a single channel confirmed the override path works — that channel's EPG converges to the placeholder while every other service on Default → Either remains on its detailed schedule.

One subjective call to flag: **Other only** drops 0x4e (actual P/F) as well as the actual-TS schedule, by strict reading of "no actual-TS at all". One-line removal if you'd rather let actual P/F through as a fallback.

From your personal point of view — does this look good to go?

DeltaMikeCharlie:
> From your personal point of view — does this look good to go?
>
I am sorry, I am unable to test/validate this because my broadcasters do not use the 'Other' table.

I can't see where you test and convert the old flag.  I will look again tomorrow morning, perhaps I missed it.

Consider summarising the `.desc` text and then moving the detailed text into the appropriate help text document.  You also have some useful comments in the code, these could also be added to the help text.

I'm not totally sure, but the help text files will look something like this:

docs/class/epggrab_mod_ota.md
docs/class/mpegts_service.md

These files may need to be created.

oliversluke:
Thanks @DeltaMikeCharlie!

The migration is at `src/input/mpegts/mpegts_service.c:111-129` (`mpegts_service_class_load`): if the new `dvb_eit_processing` key is absent and the legacy `dvb_ignore_eit` is present, map `true → None` and `false → Default`. Wired via `.ic_load`; the old key drops from the config on next save.

No migration for `eit_actual_priority` — that field only ever existed on this PR branch, so nobody upgrading from `master` has it set. The new global defaults to `Either`, matching pre-PR behaviour exactly.

Help text: shipped (`cd4d3907b`). Full per-value descriptions now live in `docs/property/eit_processing.md` and `docs/property/eit_processing_default.md`, wired via `PROP_DOC` — same pattern as `cron` / `ota_genre_translation` in `epggrab.c`. The in-form `.desc` strings are now short pointers to the Help icon. (Per-property help goes under `docs/property/` rather than `docs/class/`; the latter is for page-level docs like `mpegts_service.md`.)

DeltaMikeCharlie:
> The migration is at `src/input/mpegts/mpegts_service.c:111-129` (`mpegts_service_class_load`):
>
Thanks for that.  It was late. 🥱

The help text/`.desc` look better.  Thanks.

It just occurred to me that in 'Adaptive' mode, links between DVR entries and EPG events may become broken in some situations.  This is because adaptive mode behaviour appears to be reset every session, ie, between TVH restarts.

For example:  If the EPG entries are currently sourced from 'Actual' and you schedule an event to be recorded via the EPG (`api/dvr/entry/create_by_event`), there is a link formed between the EPG event and the DVR entry.  This means that when the EPG event changes, the DVR entry is automatically adjusted.

After a reboot, the EPG events may be sourced from 'Other', at least temporarily until 'Actual' data arrives.  This could result in the original EPG event being deleted and replaced by the temporary 'Other' event, thus breaking the link to the DVR entry.  Once the 'Actual' data recreates the EPG, the link to the DVR will not be renewed and any future scheduling changes will not be propagated to the DVR entry.

You could test this hypothesis artificially on your system.  Set a service to 'Actual', do an EPG grab and then create a DVR entry via the EPG screen.  Next, change that service to 'Other', restart TVH and do another grab.  Check to see if the DVR entry retains its link to an EPG event.

https://github.com/tvheadend/tvheadend/blob/f37b7b2cbcb704f4ff133a57312e69a200648828/src/dvr/dvr.h#L268-L273

oliversluke:
Thanks — that's a sharp catch. I traced it through and it turns out the picture is a bit better than it first looks, because there's an existing safety net in `dvr_db.c` I hadn't credited. Let me walk through the same shape of scenario you described:

**Setup, before restart**
- A channel (call it Channel A) is served by its own multiplex carrying detailed actual-TS EIT — 30+ events per day, real titles, precise times.
- A neighbouring multiplex carries other-TS for Channel A as two 24h placeholder blocks — "today" and "tomorrow", title "Placeholder".
- Adaptive is settled: `s_dvb_eit_actual_seen` is set, other-TS dropped, EPG shows the 30+ detailed events.
- I schedule a DVR entry for "Show Z" tomorrow at 21:00–22:00 — linked to that detailed event by `dvb_eid`.

**Restart, step by step**

1. **EPG db loads from disk.** Detailed events are restored with their original `dvb_eid`s. The DVR entry's `de_dvb_eid` still matches → link is intact at this moment.

2. **Tuners come up. OTA scan starts.** Suppose the neighbouring multiplex finishes first (scan order is hardware-dependent — there's no actual-TS-priority sequencing today). Its other-TS for Channel A arrives. The runtime `s_dvb_eit_actual_seen` was reset on restart → other-TS is allowed → the two 24h placeholder events get queued.

3. **EPG merge replaces the detailed events with the placeholders.** Different `dvb_eid`s, overlapping slots, last-write-wins. `dvr_event_removed` fires for each detailed event. The DVR entry's `de_bcast` is cleared and `de_dvb_eid` reset to 0 (`dvr_entry_assign_broadcast`).

4. **During this window the user sees:** Channel A's EPG as two giant "Placeholder" blocks, and the DVR entry temporarily without an EPG link.

5. **Eventually Channel A's own multiplex is scanned.** Detailed actual-TS events arrive. For the 21:00 "Show Z":
   - `dvr_event_updated` walks the channel's DVR entries
   - Finds the orphaned one (`de_bcast == NULL`)
   - `dvr_entry_fuzzy_match` runs with the new event's EID
   - Duration ±20%: 60min vs 60min → ✓
   - EID match: `de_dvb_eid == 0`, fast-path fails
   - **Fallback (the safety net):** title "Show Z" matches stored title, start time 21:00 within `dvr_update_window` (60s default), episode-number compatible → **relink**
   - DVR entry now points to the new actual-TS event with its new `dvb_eid`.

6. **Settled state:** EPG back to detailed, DVR linked, `s_dvb_eit_actual_seen` set again, other-TS dropped going forward.

**Why placeholders can't fool the safety net:** a 24h "Placeholder" event has 24× the DVR's recorded duration. The ±20% gate at the top of `dvr_entry_fuzzy_match` rejects it outright — a placeholder can never satisfy that check. The relink only succeeds against an event whose duration is close to what the DVR remembers, which is the detailed actual-TS event itself.

**The residual edge:** if the broadcaster has shifted the slot (Show Z moved from 21:00 to 21:30) or the title text drifted between sessions, fuzzy_match's fallback can't bridge it. But that's general EPG-volatility — it would also break a manually-scheduled-then-rescheduled entry, and it doesn't depend on Adaptive.

**The only Adaptive-specific bite** is a recording IN PROGRESS at restart that receives a "running late, extend stop time" update DURING the swap window. The update lands on the placeholder, the DVR is unlinked, the extension is lost, recording stops at the originally-scheduled time. Cost: a few minutes of tail missed in a rare timing alignment.

**Compared to today (Either mode):** the same swap-window dynamic exists any time other-TS arrives after actual-TS during normal operation — not just at boot. Adaptive doesn't add a new failure mode; it makes the window deterministic at startup. On the upside, Adaptive **self-heals within one OTA cycle**: after the first actual-TS arrival, the placeholder is permanently filtered for that service. Either mode keeps flipping forever — the placeholder can re-overwrite at any moment.

**Two paths from here, depending on how surgical you'd like:**

**Option 1 — leave as-is, document the trade-off.** The fuzzy-match safety net handles the placeholder case for any DVR entry whose programme metadata hasn't drifted. I'd add a paragraph to `docs/property/eit_processing.md` describing the startup window and the in-progress-extension edge so operators know what they're opting into. Net behaviour is strictly better than today (the self-healing alone). Cheapest ship; the residual edge is minimal.

**Option 2 — persist `s_dvb_eit_actual_seen` per service.** Store the "actual-TS observed" flag in the service config. After the first session that observed actual-TS for a given service, the flag survives across restarts → other-TS is dropped from boot → startup window collapses to zero. The only follow-up question is where to surface a "reset adaptive learning" affordance for the case where a broadcaster genuinely stops carrying actual-TS (multiplex regroup) and a service needs to relearn. Two candidate placements:
- A button on the EPG Grabber configuration page, next to the new "EIT processing (default)" enum, that resets the flag across every service in one go.
- A per-service "reset" affordance alongside the per-service EIT processing dropdown.

If I'm honest about my preference, I lean slightly toward Option 1 — the safety net carries the realistic load and avoiding one more setting + a reset affordance feels right for the size of the residual edge. Genuinely fine with Option 2 too if you'd rather close the window definitively — what's your preference?

DeltaMikeCharlie:
> Two paths from here, depending on how surgical you'd like:
>
I think that Option 2 is the most thorough approach, and that the `actual_seen` flag could just be exposed as an 'Expert' boolean for each service via the UI that the user could manually reset when required.  However....
>
> what's your preference?
>
Remembering that I have only personal opinions to offer.

Option 1 would seem to be suitable for the _vast majority of use cases_ and Option 2 could be retro-fitted at a future date if required.

**Extra Bonus:**  I also have an idea about enhancing the fuzzy match that you described:  I think that the process could be augmented by adding an `orig_eit_event_id` helper property to the DVR entry class that is set when the DVR entry is first created.

Just before a fuzzy match is performed (the EPG event ID and DVR event ID link don't match) evaluate the `orig_eit_event_id`.  If it matches, no need to go fuzzy, you have a match.  If no match, but there is a subsequent successful fuzzy match, then set the `orig_eit_event_id` property to the matched EIT event ID.

This way the preferred EPG event ID is effectively cached by the DVR entry object so that it can be reunited after the preferred EPG data returns.  If both 'Actual' and 'Other' achieve a fuzzy match, the link will just flip flop between the 2 sources.

Also, the EPG also stores an XMLTV reference that should be cached in the DVR object at achieve a similar result.

oliversluke:
@DeltaMikeCharlie Thanks — going with Option 1. The startup-window caveat paragraph is now in `docs/property/eit_processing.md` and `docs/property/eit_processing_default.md`. If the persisted-flag retrofit (Option 2) turns out to matter in practice, it's a small additive change that can land on top without disturbing anything in this PR.

Long-term, if precedence questions keep surfacing across grabbers (not just EIT actual-vs-other), the more thorough answer is the Enigma2 model — source-tagged events at the `epg_broadcast_t` level with a project-wide priority ordering. That's a much bigger change than this PR (touches the EPG core + the on-disk database format) and I'd rather not pre-empt it; if Option 1 + a future Option 2 prove insufficient at scale, that's when the source-tag architecture earns its complexity.

On the fuzzy-match cache idea — that's a nice strengthening of `dvr_entry_fuzzy_match`. The fast-path `orig_eit_event_id` check would short-circuit the common "source flipped, original came back" case before fuzzy heuristics even run, and the cache update on fuzzy success means each session leaves the next one in a better state. Caching the XMLTV ref the same way gives equivalent resilience for XMLTV-sourced entries.

I'd like to spin it up as a separate small PR after this one merges rather than folding it in here — it's a DVR-module change that's independently valuable (helps every grabber, not just EIT) and keeping this PR scoped to the EIT-precedence policy keeps the diff reviewable.

**From my side this PR is good to merge.**

DeltaMikeCharlie:
> Thanks — going with Option 1.
>
Cool.
>
> From my side this PR is good to merge.
>
It looks good.  Thanks for the effort.
>
> source-tagged events at the `epg_broadcast_t` level.
>
Yes, I was thinking about something similar today, but it would be WAY out of scope for this PR.  Maintaining multiple EPG sources and then allowing the user to change which EPG source they are viewing.

Flole998:
Sorry, I tried to merge this from my phone but messed up while rebasing.... @oliversluke Could you just do a force-push of your local branch again please?

oliversluke:
@Flole998 Done — rebased onto current master and force-pushed, so it's back to a single commit with the stray merge commit gone. Shows as mergeable on my side; CI should re-run. Thanks for picking it up!

— Oliver

PR #2123, opened 2026-05-25
dvr: trim dangling partial UTF-8 in subscription title

### LINKED ISSUE #2121: [BUG]: WebSocket sends invalid JSON when DVR has recording with long unicode title
### Tvheadend Version

v4.3 (stable)

### Build Number

4.3.2662~gf37b7b2cb

### Platform

Ubuntu 24.04 (Noble)

### Architecture

Intel x86_64

### Category

PVR/DVR

### Description

Condition:
- DVR has running recordings with long unicode title (ex: 転生したらスライムだった件　ＦＲＩＤＡＹ　ＡＮＩＭＥ　ＮＩＧＨＴ🈑🈓)

Error details:
- `snprintf` at `dvr_rec.c (dvr_rec_subscribe)` cuts string mid of unicode character.
- `/comet/ws` response with invalid JSON string. Finally, WebUI is not accessible until recording ends.

Proposal modification:
```diff
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -95,6 +95,12 @@ dvr_rec_subscribe(dvr_entry_t *de)
   weight = prio2weight[pri];
 
   snprintf(buf, sizeof(buf), "DVR: %s", lang_str_get(de->de_title, NULL));
+  /* A long title (e.g. Japanese, 3 bytes/char) overflows buf and snprintf
+   * truncates it mid-UTF-8-character, leaving a dangling lead byte. That
+   * invalid string becomes the subscription title and is streamed in the
+   * comet status feed, breaking the web UI's websocket. Trim the partial
+   * trailing character so the subscription title stays valid UTF-8. */
+  utf8_validate_inplace(buf);
 
   if (de->de_owner && de->de_owner[0] != '\0') {
     aa = access_get_by_username(de->de_owner);
```

### Logfile

_No response_

### Reproduction

_No response_

oliversluke:
## Summary

`dvr_rec_subscribe` (`src/dvr/dvr_rec.c`) formats the subscription
title into a fixed 100-byte buffer via `snprintf`. `snprintf`
truncates byte-wise, so a long multibyte title — CJK characters
(3 bytes/char), wide ASCII, or emoji (4 bytes/char) — can leave a
dangling partial UTF-8 sequence at the end of the buffer.

The truncated title flows into the comet/ws JSON feed; an invalid
byte there causes the browser's WebSocket parser to drop the
frame, and the web UI stays silent until the recording ends.

## Fix

Call `utf8_validate_inplace(buf)` right after the `snprintf`. The
helper is already used twice in the same file (lines 356 + 359 on
the `dvr_sub_title` path) for the same purpose, so the pattern
matches existing code.

The buffer stays at 100 bytes — the title is just snapped to a
clean UTF-8 codepoint boundary instead of mid-sequence. Growing
the buffer was considered but rejected: arbitrary-length titles
exist, a larger fixed buffer only shifts the problem out, and a
dynamic allocation is a bigger change for the same end result.

A broader hardening (validate UTF-8 at the JSON encoder for every
string property) was raised in the issue thread but is a separate
concern with much wider scope — not folded into this PR.

## Test plan

- [ ] Build clean (`./configure && make`)
- [ ] Schedule a recording with a long CJK / emoji title (~32+ wide chars)
- [ ] Confirm the web UI's comet/ws feed stays connected during the recording
- [ ] Confirm normal-length ASCII recordings still show their title verbatim

Fixes: #2121

### DIFF
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -73,7 +73,9 @@ dvr_spawn_fetch_artwork(dvr_entry_t *de)
 int
 dvr_rec_subscribe(dvr_entry_t *de)
 {
-  char buf[100];
+  const char *title;
+  char *buf;
+  size_t buflen;
   int weight;
   profile_t *pro;
   profile_chain_t *prch = NULL;
@@ -94,7 +96,14 @@ dvr_rec_subscribe(dvr_entry_t *de)
     pri = DVR_PRIO_NORMAL;
   weight = prio2weight[pri];
 
-  snprintf(buf, sizeof(buf), "DVR: %s", lang_str_get(de->de_title, NULL));
+  /* Size the buffer to the actual title so long multibyte titles
+   * (CJK, emoji) aren't truncated mid-UTF-8 — a partial codepoint
+   * would otherwise break the comet/ws JSON feed. */
+  title = lang_str_get(de->de_title, NULL);
+  if (title == NULL) title = "";
+  buflen = strlen(title) + sizeof("DVR: ");
+  buf = alloca(buflen);
+  snprintf(buf, buflen, "DVR: %s", title);
 
   if (de->de_owner && de->de_owner[0] != '\0') {
     aa = access_get_by_username(de->de_owner);


Flole998:
A dynamic allocation does not have the same end result: In this case the title is cut, with dynamic allocation the full title is returned. Also with dynamic allocations the memory footprint is (slightly) reduced as the buffer is not larger than it has to be. The change is rather simple, tvh_alloca should be called with the correct length and the type should be changed to a char*, it's 1 line changed and 1 line added, so it's not a big restructuring or anything like that.

I am aware that there are more cases where assumptions on some arbitrary maximum are made, those should also be addressed at some point.

oliversluke:
Switched to a buffer sized via `alloca()` so the full title is preserved instead of truncated. Couldn't find a `tvh_alloca` symbol in the tree (or git history) — used bare `alloca()` per the existing pattern (`cron.c:167`, `http.c:313`, `string_list.c:92`, etc.). Force-pushed onto the same commit; PR stays at one commit.

You're right that other places in the tree make similar fixed-buffer assumptions — happy to take those on as separate PRs.

DeltaMikeCharlie:
What about using this issue to create a TVH helper function using `wchar_t` functions that respect character boundaries?

Functions like `mbsrtowcs` to respectfully convert the `char` to `wchar`, `swprintf` to produce the output string and `wcrtomb` to convert back to bytes?

There may be better specific functions, but is the methodology worth considering?

PR #2124, opened 2026-05-26
epggrab: bound queued EPG input data

RF1705:
```markdown
This adds a per-grabber limit for queued EPG input data.

During OTA EIT grabbing, MPEG-TS table callbacks can enqueue EPG sections into
`epggrab_queue_data()` faster than the epggrab data worker can process them.
That queue was previously unbounded. On constrained systems this can amplify
memory pressure and lock contention, and may contribute to repeated warnings
such as:

```text
mpegts: too much queued table input data (over 2MB), discarding new
```

The MPEG-TS table input queue already has a hard limit, but the downstream EPG
data queue did not. This patch tracks queued bytes per `epggrab_module_t` and
drops new EPG data once the per-module queue exceeds 8 MB, with a rate-limited
warning.

Dropping new EPG sections under overload is preferable to allowing the EPG
backlog to grow without bound and potentially stall the daemon or WebUI.

This change does not touch DVB tuning, descrambling, OSCam, mux defaults, or
delivery-system handling.

Testing:
- `git diff --check` passes.
- Full local build could not be completed in my environment because
  `pkg-config` is not installed, so `./configure` stops before generating the
  build tree.
```

**Optional note**
```markdown
This is intended as a defensive backpressure fix for reports where OTA
EPG/EIT/mux scanning coincides with repeated table queue warnings and an
unresponsive WebUI.
```

### DIFF
--- a/src/epggrab.c
+++ b/src/epggrab.c
@@ -51,6 +51,19 @@ static TAILQ_HEAD(, epggrab_module) epggrab_data_modules;
 
 static memoryinfo_t epggrab_data_memoryinfo = { .my_name = "EPG grabber data queue" };
 
+#define EPGGRAB_DATA_QUEUE_LIMIT (8 * 1024 * 1024)
+
+static void
+epggrab_queue_size_sub ( epggrab_module_t *mod, const epggrab_queued_data_t *eq )
+{
+  size_t len = sizeof(*eq) + eq->eq_len;
+
+  if (mod->data_queue_size >= len)
+    mod->data_queue_size -= len;
+  else
+    mod->data_queue_size = 0;
+}
+
 /* Config */
 epggrab_module_list_t  epggrab_modules;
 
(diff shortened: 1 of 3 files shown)


Flole998:
Just to double check: You didn't even build this, not to mention test if this actually works?

RF1705:
Correct. I could only run `git diff --check` locally before opening the PR.

The local build environment I used is missing `pkg-config`, so `./configure` stops before generating the build tree. I mentioned that in the PR description rather than pretending this had been build-tested.

I have now also addressed the automated review feedback:
- move the rate-limit state update under `epggrab_data_mutex`
- keep the actual warning outside the mutex
- avoid `memoryinfo_alloc()` on entries that are dropped
- add a defensive queue-size subtraction helper
- explicitly initialize the new queue accounting fields in `epggrab_module_create()`

I understand if you prefer this to be treated as an issue/proposed direction until it has been built and runtime-tested on a Tvheadend setup. The intent is a defensive backpressure limit for the downstream epggrab data queue, not a claimed complete fix for every cause of the MPEG-TS table queue warning.

RF1705:
I’m going to close this PR for now because it was not build-tested or runtime-tested.

The underlying concern remains: during OTA EIT grabbing, `epggrab_queue_data()` can accumulate data behind the MPEG-TS table parser path, while the upstream MPEG-TS table queue already has a hard 2 MB limit. On constrained systems this may contribute to repeated:

mpegts: too much queued table input data (over 2MB), discarding new

This PR was intended as a defensive backpressure proposal, but it needs proper build and runtime validation before it should be considered for merge.

PR #2136, opened 2026-06-13
Fix GCC 16 -Werror build failures (satip, ratinglabels)

### LINKED ISSUE #2137: [BUG]: Build fails on GCC 16 due to -Werror (unused-but-set-variable, format-truncation)
### Tvheadend Version
v4.3 (stable)

### Build Number
v4.3-2681-gbb09b4fae

### Platform
Other.. (Fedora 44)

### Architecture
ARMv8 (aarch64)

### Category
Packaging

### Description
Building current master (bb09b4fae) on Fedora 44 with GCC 16.1.1 fails. Two
long-standing constructs are now rejected under the project's default `-Werror`:

1. `src/ratinglabels.c:615` — a local variable that is set but never read:

```
src/ratinglabels.c:615:7: error: variable 'a' set but not used [-Werror=unused-but-set-variable=]
  615 |   int a=1;
      |       ^
```

2. `src/input/mpegts/satip/satip_rtsp.c:364` — `_w[16]` is too small to hold
   `"&tvhweight=%d"` for the full range of `int`:

```
src/input/mpegts/satip/satip_rtsp.c:364:42: error: '%d' directive output may be truncated writing between 1 and 10 bytes into a region of size 5 [-Werror=format-truncation=]
```

Both are dormant (the satip line dates to 2019, the ratinglabels line to 2023);
GCC 16's stricter diagnostics newly promote them to errors.

This isn't caught in CI because the `build-rpm-native` (Fedora) job is marked
`continue-on-error: true`, so its failures don't fail the pipeline.

Fix: #2136

### Reproduction
1. `../configure && make` on Fedora 44 (GCC 16.1.1)
2. Build aborts at `src/ratinglabels.o` / `satip_rtsp.o`.

pimzand:
Current master fails to build on Fedora 44 / GCC 16.1.1 because of two
diagnostics that GCC 16 newly treats as errors under the default `-Werror`.

**satip_rtsp.c** — the fixed 16-byte `_w` buffer cannot hold `"&tvhweight=%d"`
for the full range of `int weight` (worst case 23 bytes incl. NUL), tripping
`-Werror=format-truncation`. Enlarged to 24 bytes.

**ratinglabels.c** — `ratinglabel_class_enabled_notify()` only incremented a
local variable that was never read; it had no observable effect. Removed the
dead body (`-Werror=unused-but-set-variable`), leaving the registered notify
callback as a no-op — no behavior change.

Both are old, dormant lines, not regressions.

Fixes #2137

### DIFF
--- a/src/input/mpegts/satip/satip_rtsp.c
+++ b/src/input/mpegts/satip/satip_rtsp.c
@@ -340,11 +340,12 @@ satip_rtsp_play( http_client_t *hc, const char *pids,
                  int max_pids_len, int weight )
 {
   char *stream = NULL;
-  char _stream[32], _w[16];
+  char _stream[32], _w[24];
   const char *p[8], *add[8], *del[8];
   int pcnt, addcnt, delcnt;
   int i, r = 0, index = 0;
-  char buf[max_pids_len + 32];
+  /* room for "delpids=%s&addpids=%s%s": del+add up to max_pids_len, plus _w */
+  char buf[(max_pids_len < 32 ? 32 : max_pids_len) + 64];
 
   if (max_pids_len < 32)
     max_pids_len = 32;
--- a/src/ratinglabels.c
+++ b/src/ratinglabels.c
@@ -612,6 +612,8 @@ ratinglabel_class_get_icon ( void *obj )
 static void
 ratinglabel_class_enabled_notify ( void *obj, const char *lang )
 {
+  /* Intentionally empty: toggling "enabled" needs no side effect; the
+     property write is sufficient. Kept as a registered no-op hook point. */
 }
 
 CLASS_DOC(ratinglabel)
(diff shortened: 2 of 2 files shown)


pimzand:
Addressed the review feedback on `ratinglabel_class_enabled_notify`: kept it as a registered no-op and added a comment explaining why the body is intentionally empty (toggling `enabled` has no side effect). Left as a separate commit rather than a force-push.

pimzand:
Also, my apologies for how messy this PR became. I was under the impression I should always squash everything into one commit, and at the same time I kept re-running the automated review and reacting to each round of suggestions. Combined, that produced a confusing trail of rewrites and review comments that made the PR much harder to follow than it should have been.

Going forward I'll keep changes as plain follow-up commits so the history stays easy to read. And whenever you're happy with it, I'm glad to do a final squash into a single clean commit at your request — or you can simply squash-merge it yourself.

Flole998:
Its not messy, its a single commit per change just like it should be :) Thanks!

PR #2161, opened 2026-07-08
webui: per-route cache-control for the Vue UI's hashed assets

pimzand:
-

### DIFF
--- a/src/webui/webui.c
+++ b/src/webui/webui.c
@@ -244,12 +244,17 @@ to reauthenticate."));
 
 /**
  * Static download of a file from the filesystem
+ *
+ * maxage is the default Cache-Control max-age; the per-extension rules
+ * below may raise it but never lower it, so a caller that knows its
+ * content is immutable (see page_vue_asset in vue.c) keeps the longer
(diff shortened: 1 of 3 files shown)


src/webui/webui.c
```diff
@@ -244,20 +244,24 @@ to reauthenticate."));
...
  * Static download of a file from the filesystem
+ *
+ * maxage is the default Cache-Control max-age; the per-extension rules
+ * below may raise it but never lower it, so a caller that knows its
+ * content is immutable (see page_vue_asset in vue.c) keeps the longer
+ * lifetime it asked for.
  */
 int
-page_static_file(http_connection_t *hc, const char *_remain, void *opaque)
+page_static_file2(http_connection_t *hc, const char *_remain,
```
Flole998: Lets stop using function names with numbers at the end, they are super hard to read/understand. Give it a "maxage" suffix or something like that so that it's clear from the name what the function does

pimzand: Fair point - renamed to page_static_file_maxage() in 149ce5e4c.

PR #2181, opened 2026-07-15
build: add nodejs and npm to dockerfile

Trujulu:
Installs Node.js and npm to support the Vue.js build process. Previously, the Vue.js assets were not being included in the image because the JavaScript runtime was missing.

### DIFF
--- a/Containerfile.alpine
+++ b/Containerfile.alpine
@@ -34,6 +34,8 @@ RUN apk add --no-cache \
         'uriparser-dev' \
         'wget' \
         'zlib-dev' \
+        'nodejs' \
+        'npm' \
     && \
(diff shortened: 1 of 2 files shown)


Trujulu:
It seems the Vue interface is incompatible with 32-bit Linux, as the native Rollup build does not support the ia32 architecture:
`#32 694.8 Error: Your current platform "linux" and architecture "ia32" combination is not yet supported by the native Rollup build. Please use the WASM build "@rollup/wasm-node" instead.`

Docker Alpine build works fine on my local 64-bit machine. I'm wondering if it's possible to change the architecture of the GitHub builder or if there's a workaround to cross-compile the assets."

Flole998:
The interface is pre-built and it should just be copied over.

PR #2200, opened 2026-07-24
wizard: flag the setup wizard password fields as passwords

oliversluke:
### What

Marks the setup wizard's two password fields (`admin_password`, `password` on the login step) with `PO_PASSWORD`, so web UIs render them as masked password inputs instead of clear text.

### Why

The login step declares both props as plain `PT_STR` with no `.opts = PO_PASSWORD`. `PO_PASSWORD` is what `prop_serialize()` translates into the `password: true` JSON-metadata field that the web UIs key off for input-type masking — without it, both the classic and the new UI displayed the wizard's password values in plain text as you type. Other password-bearing consumers (`access.c`, `cclient.c`) already declare the flag; the wizard was the odd one out.

### How

- `src/wizard.c`: `.opts = PO_PASSWORD,` on both prop declarations (the two username fields stay plain, as intended).
- The new UI had a scoped client-side overlay that patched `password: true` onto exactly these two field ids to compensate — with the server metadata now correct, that workaround (and its unit-test pin) is removed. Net: +2 C lines, −83 workaround lines.

The wizard's save path is untouched — the `*`-placeholder round-trip logic in `wizard.c` behaves exactly as before.

### Testing

- Built and ran the wizard: the login step's Admin password / Password fields now render as `<input type="password">` (with the new UI's show/hide toggle); usernames stay plain; saving the step sets the passwords correctly.
- Verifiable in the classic UI too — same metadata drives its form renderer.
- Vue unit suite green (workaround tests removed with the workaround); `vue-tsc` + ESLint pass; full `make` clean.

### DIFF
--- a/src/webui/static-vue/src/views/wizard/WizardStepGeneric.vue
+++ b/src/webui/static-vue/src/views/wizard/WizardStepGeneric.vue
@@ -56,14 +56,6 @@ import { useWizardStore, type WizardStepName } from '@/stores/wizard'
 import { addExternalLinkAttrs, renderMarkdown, rewriteStaticUrls } from '@/utils/markdown'
 import type { IdnodeProp } from '@/types/idnode'
 
-/*
- * Wizard password-field workaround scope — see handleLoaded
- * below. Module-scope Set for O(1) lookup; matched in the
- * loaded-emit patch loop. Deletable once
- * src/wizard.c:434, 452 add `.opts = PO_PASSWORD,`.
- */
-const WIZARD_PASSWORD_IDS = new Set(['admin_password', 'password'])
-
 interface Props {
   /* Canonical step name. Drives the load/save endpoints and the
    * prev/next route resolution. Route components pass this as a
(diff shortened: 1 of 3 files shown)


DeltaMikeCharlie:
> so web UIs render them as masked password inputs instead of clear text.

Is there a 'show password' button so that you can verify the password that you have just entered?

oliversluke:
Good question — it points at the one trade-off in this PR. In the new UI: yes, every password field carries the standard per-field eye toggle (show/hide), and that's exactly what this fix switches on for the wizard — masking plus the toggle.

For the classic UI the honest answer is: before this PR the wizard showed the passwords in clear text (that's the bug), and after it they're masked with no reveal. The classic UI's existing "Show passwords" affordance is a grid-toolbar button, which the wizard's editor window doesn't include — so the wizard becomes a password surface without a reveal there.

I see two ways forward:

1. **Merge as-is.** The classic UI is in maintenance mode and new UI investment is where the effort goes; the wizard is a run-once flow, and a mistyped admin password is recoverable (superuser account, or re-running the wizard).
2. **Extend this PR** to retrofit a "show password" button to **all** places the classic UI renders password fields — the wizard window and the grid row editors alike — so the classic UI gets the same per-field verifiability the new UI has.

Opinions welcome — happy to go either way.

chewitt:
IMHO the more daylight we place between old vs new GUI the better, and the sooner we justify dropping Classic from the codebase. I'd say no effort on Classic is required.

PR #2205, opened 2026-07-29
mpegts: preserve stream mapping across PID switches

LiBe-net:
## Summary

- restore the active stream map after timeshift seeks
- keep pass-through recording PIDs stable when a service changes its source PIDs

## Dependency

Depends on #2204, which ensures that teletext streams have unique component indexes. The stable PID mapping introduced here uses those indexes when matching replacement streams.

## Problem

Some services change their video, audio, or teletext PIDs while remaining on the same service.

Replacing the active elementary streams also changed the stream map exposed to pass-through recordings and timeshift playback. Players could consequently lose tracks or fail to seek reliably.

## Implementation

The pass muxer now maintains an output stream map and maps replacement source streams to the established output PIDs.

This keeps the output PMT, PCR PID, and elementary stream PIDs consistent while the underlying service changes its source streams.

Timeshift restores its stream map after a seek.

No external API changes are required.

## Validation

- `git diff --check origin/master...HEAD`
- successful Docker build
- pass-through recording keeps its output stream PIDs stable across source PID changes
- video and audio continuity counters remain valid across the stream replacement

### DIFF
--- a/src/muxer/muxer_pass.c
+++ b/src/muxer/muxer_pass.c
@@ -868,25 +868,25 @@ pass_muxer_rewrite_enabled(const pass_muxer_t *pm)
          pm->pm_rewrite_sdt || pm->pm_rewrite_nit || pm->pm_rewrite_eit;
 }
 
-static uint8_t *
+static int
 pass_muxer_output_buffer(pass_muxer_t *pm, const uint8_t *src, size_t len)
 {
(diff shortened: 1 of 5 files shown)


LiBe-net:
will be tested in https://github.com/tvheadend/tvheadend/pull/2206

Issue #1973, opened 2025-10-25
[BUG]: Black/Crash when switching from national to regional stream (ORF - Austria)


mariospendier:
### Tvheadend Version

v4.3 (stable)

### Build Number

4.3-2500~g7de8bf482

### Platform

Ubuntu 24.04 (Noble)

### Architecture

Intel x86_64

### Category

Muxes

### Description

When you stream ORF2 (ORF2K HD, ORF2St HD, ORF2W HD, ...) channels, every day at 19:00 there is a switch to a regional part of program. After the switch, the channel "pauses" last frame. If you stop and play the channel again, channel stays dark.

Only resolution is to restart TVHeadend. After TVH restart, channel works again (national stream) until you stream to regional stream.

### Logfile

TVH and Oscam Logfile
Switch from national to regional was 18:57:17

[...]

PR #2206, opened 2026-07-30
mpegts: combine ORF2 stream index and PID mapping fixes

LiBe-net:
## Summary

Combine the fixes from #2204 and #2205 in a single branch so testers can use
one build containing both changes.

- keep teletext component indexes unique when streams move to a different
  parent PID
- preserve pass-through recording stream/PID mappings across source PID
  changes
- restore the active stream map after timeshift seeks

## Included fixes

- #2204 — `mpegts: keep teletext stream indexes unique`
- #2205 — `mpegts: preserve stream mapping across PID switches`

The commits from both PRs are preserved individually in this branch. This PR
is intended to provide combined CI artifacts for testing; the individual PRs
remain the focused review units for each fix.

## Problem

During ORF2 regional-to-national transitions, elementary streams can be
replaced while the service remains active. Duplicate internal component
indexes can disrupt playback, while changing output PIDs and losing the stream
map after a timeshift seek can disrupt recordings and seeking.

Testing both fixes together is important because the stable PID mapping uses
the unique component indexes when matching replacement streams.

## Validation

- `git diff --check origin/master...HEAD`
- both source PR branches were independently built and tested as documented in
  #2204 and #2205

Related forum discussion:
https://tvheadend.org/d/9260-orf-2-stream-crash-on-tvh-server

### DIFF
--- a/src/muxer/muxer_pass.c
+++ b/src/muxer/muxer_pass.c
@@ -30,6 +30,8 @@
 #include "muxer_pass.h"
 #include "spawn.h"
 
+#define PASS_PID_COUNT 8192
+
 typedef struct pass_muxer {
   muxer_t;
 
@@ -46,6 +48,12 @@ typedef struct pass_muxer {
 
   /* Streaming components */
   streaming_start_t *pm_ss;
+  uint16_t          *pm_pid_map;
+  int8_t            *pm_pid_cc;
+  uint8_t           *pm_pid_rewrite_cc;
+  uint8_t           *pm_pid_buf;
+  size_t             pm_pid_buf_size;
+  uint8_t            pm_pid_active;
 
   /* TS muxing */
   uint8_t  pm_rewrite_sdt;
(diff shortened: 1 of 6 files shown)


Flole998:
You have multiple commits which are basically fixing things that you have changed in previous commits in this PR. This whole PR should be 3 commits based on your description:

- keep teletext component indexes unique when streams move to a different parent PID
- preserve pass-through recording stream/PID mappings across source PID changes
- restore the active stream map after timeshift seeks

If its easier for you I would also be fine with a single commit for "Handle PID changes properly" with the detailed description in the commit message itself.

Flole998:
Please get rid of the last merge commit, either leave the other commits as they are or rebase them onto latest master. either way is fine.

LiBe-net:
I reverted this pull request back to draft.

If the PID changes (only occurs when switching ORF2 National to Regional or back): With this pull request, live TV and recordings will continue to work, including the time-shift function when playing back recordings. However, time-shifted playback of live TV will still not work after the PID change. I will do some further testing.

LiBe-net:
So timeshift now also works as expected. As this commit also contains the fix for live-tv: https://github.com/tvheadend/tvheadend/pull/2204 
@Flole998: if you prefer to merge this all in one pull request, you can close [2204](https://github.com/tvheadend/tvheadend/pull/2204). 
It solves this: https://github.com/tvheadend/tvheadend/issues/1973 and the related discussion: https://tvheadend.org/d/9260-orf-2-stream-crash-on-tvh-server/145

LiBe-net:
tested the last implemented suggestions from copilot yesterday: ORF2 still works after PID changes in all cases: live, timeshift and recording.

phaze75:
I'm still waiting for the builds to become available.

Flole998:
I am not quite sure if it's 100% correct (as there have been reports of small glitches when the switch occurs) but its significantly better than before, so once there is sufficient testing I am intending to merge this.

phaze75:
I just tested the actual fix(es). The regional  switch at the beginning causes a small glitch, while the switch at the end causes a stream rebuffering. I‘m no expert and I don’t know if switches without glitches are possible at all. With tvheadend I never experienced a glitch-free switch. Not even before the freeze of the ORF2 stream started to happen.

phaze75:
> so once there is sufficient testing I am intending to merge this.

Which timeline do you have in mind?

LiBe-net:
I've continued testing this PR extensively since the latest fixes were completed, especially the ORF2 regional/national PID changes.

So far it has been working reliably for me in all relevant scenarios:

* live TV across the PID change
* timeshift across the PID change, including seeking back and forth
* recordings across the PID change

I have repeated these tests multiple times and have not encountered any remaining functional issues.

The previous review comments have been addressed and the current CI checks are passing as well. Since this PR fixes a real issue that still affects ORF2 users, is there anything specific still blocking it from being merged?

If additional testing of a particular scenario is needed, please let me know what exactly should be tested.

src/muxer/muxer_pass.c
```diff
@@ -72,6 +80,230 @@ typedef struct pass_muxer {
...
+static int
+pass_muxer_component_match(const streaming_start_component_t *old,
+                           const streaming_start_component_t *new)
+{
+  int score;
+
+  if (old->es_type != new->es_type)
+    return -1;
+
+  score = old->es_index == new->es_index ? 8 : 0;
```
Flole998: Is there a specific reason why the index has a higher score when matching than the language? I would expect the language to have the highest score so that if a broadcaster shuffles around the languages during a regional switch the language is followed and the index is ignored (which would be the expected behaviour).

src/timeshift.c
```diff
@@ -327,6 +339,43 @@ timeshift_packet( timeshift_t *ts, streaming_message_t *sm )
   return 0;
 }
 
+static void
+timeshift_packet_rebase( timeshift_t *ts, streaming_message_t *sm )
```
Flole998, line 360: This whole logic doesn't seem to account for any rollover if I see it correctly? The 33 bit timestamp will just overflow at some point?

src/muxer/muxer_pass.c
LiBe-net: Good point. There wasn't a strict requirement for the index to have the highest score. I gave it more weight because the component index is kept stable/unique by the other part of this fix and worked reliably for the ORF2 PID changes I tested.

But I agree that if multiple streams of the same type are present and the broadcaster rearranges them, the language should be the stronger identity signal.

I'll adjust the scoring so that a matching non-empty language has a higher priority than the index, with the index remaining as a fallback when no useful language information is available.

src/timeshift.c
LiBe-net: Yes, good catch. This code currently assumes that the adjusted PTS can be compared as a continuously increasing value and does not explicitly handle the 33-bit wrap here.

The previous timeshift code already applied `start_pts` directly as well, but with the new rebase logic I'm now also using that value to detect a backwards jump, so relying on the previous behaviour isn't a good idea.

I'll change this to handle the PTS wrap explicitly, using the existing wrap-aware logic (`PTS_MASK` / `pts_diff()`) or an equivalent unwrapped timestamp calculation, and retest the PID-change/timeshift cases afterwards.

LiBe-net:
I also added a separate commit to address the SonarCloud failure:

esstream: avoid use-after-free updating nicename

This issue is unrelated to the ORF2/PID-change changes. SonarCloud detected a pre-existing potential use-after-free in elementary_set_update_nicename() because nicename could alias set->set_nicename.

The fix duplicates the new value before freeing the old one. I kept it as a separate commit so it remains clearly separated from the ORF2 changes.

LiBe-net:
I checked the remaining SonarCloud failure in detail. It points to the existing `elementary_set_clean()` / `elementary_set_stream_destroy()` cleanup code in `src/esstream.c`, not to code introduced by this PR.

Sonar itself marks the issue as having been introduced about 8 years ago. The PR only changes `elementary_stream_create_parent()` in this file, so this appears to be a pre-existing Sonar false positive unrelated to the ORF2 fixes.

I had added a separate `nicename` cleanup commit while investigating the Sonar failure, but since it turned out to be unrelated to the reported issue, I removed that commit again to keep this PR focused.

Flole998:
Don't worry about the sonar stuff too much, it's always unhappy with something ;)

LiBe-net:
implemented both suggestions and tested again today: live tv, timeshift and recording: all cases still works when pids are changing..

PR #2208, opened 2026-07-30
webui: honour dedup-skipped recordings in the new UI

### LINKED ISSUE #2207: [BUG]: Vue UI: Upcoming grid shows dedup-skipped entries as normal scheduled recordings
### Tvheadend Version

v4.3 (stable)

### Build Number

Local build based on upstream commit 15515d784 (4.3-2738~g15515d784); code references below are to current master (fd531a452), which does not differ in this area.

### Platform

Other.. (Fedora 44)

### Architecture

ARMv8 (aarch64)

### Category

WebGUI

### Description

`api/dvr/entry/grid_upcoming` includes dedup-skipped (duplicate) entries unless the client passes `duplicates=0`; the server default is to include them (`src/api/api_dvr.c:83`).

The two UIs handle this differently:

- The ExtJS Upcoming grid always requests `duplicates=0` (`src/webui/static/app/dvr.js:508`), so dedup-skipped entries never appear in the grid. In addition, the DVR details dialog shows a red "Will be skipped because it is a rerun of:" note for such entries.
- The Vue Upcoming view (`src/webui/static-vue/src/views/dvr/UpcomingView.vue`) sends no `duplicates` parameter and does not request the `duplicate` field. Dedup-skipped entries therefore appear as ordinary rows with status "Scheduled for recording", indistinguishable from entries that will actually be recorded.

As a result the two Upcoming views disagree on the row set and on the recording count, and the Vue view is misleading: an entry it shows as scheduled will in fact be skipped.

Expected behaviour: the Vue Upcoming view should either pass `duplicates=0` like the ExtJS grid, or request the `duplicate` field and render skipped entries visually distinct (greyed out or struck through, with a hint naming the original broadcast date), possibly behind a show/hide toggle.

### Logfile

_No response_

### Reproduction

1. Create an autorec rule with a duplicate detection setting (for example "All: Record if different episode number") matching a show that has a rerun in the EPG.
2. Wait for the scheduler to mark the rerun as a duplicate (the entry's `duplicate` field becomes non-zero).
3. Compare the Upcoming grid in the ExtJS UI with the Upcoming view in the Vue UI: the ExtJS grid hides the rerun; the Vue view shows it as "Scheduled for recording".

oliversluke:
### What

Fixes the new UI showing dedup-skipped rerun entries as ordinary scheduled recordings, and adds the classic details dialog's "Will be skipped — rerun of …" explanation to the new UI's EPG event drawer.

Fixes #2207.

### Why

`api/dvr/entry/grid_upcoming` includes dedup-skipped entries unless the client passes `duplicates=0` (`api_dvr.c`). The classic Upcoming grid has always passed it (`dvr.js`); the new Upcoming view did not — skipped reruns appeared as normal "Scheduled for recording" rows although they will not record, and the two UIs disagreed on the row set and count (exactly as reported).

### How

Two halves, mirroring how the classic UI handles the same situation:

1. **Upcoming view**: pass `duplicates: 0` through the grid's existing extra-params mechanism — skipped reruns disappear from the list and the count, matching the classic grid. A unit test pins the parameter so it can't silently regress.
2. **EPG event drawer**: a hidden rerun still needs an explanation *somewhere* (the classic UI provides it in its details dialog, `epg.js:71`). When the drawer opens on a scheduled entry it loads the DVR entry's read-only `duplicate` property via a targeted `idnode/load` — the schedule status cannot carry this, since a skipped rerun still reports plain `"scheduled"` (`dvr_db.c:704`) — and renders an amber note: *"Will be skipped because it is a rerun of: `<date>`"*. The note reuses the classic message ids, so all existing translations apply unchanged.

The richer alternative from the report — showing skipped entries in the grid greyed-out behind a show/hide toggle — is a genuine enhancement beyond what either UI does today and is deliberately left for a separate PR if there's appetite; this one is strictly classic-parity.

### Testing

- New unit test pinning `duplicates: 0` + endpoint on the Upcoming grid mount; full Vue suite green; `vue-tsc`, ESLint, SPDX header check, `npm run build` all pass.
- Verified against a live instance: the endpoint accepts `duplicates=0` in the UI's wire form, and the drawer's `idnode/load` fetch was validated against a real DVR entry (`params[]` shape, `duplicate` present as a number, `0` for a non-duplicate → no note; the C getter returns the original broadcast's start time when the dedup logic matches, so a real duplicate renders the note with the original's date).
- What I could not stage locally is a live dedup-skipped entry (needs an autorec + rerun in EPG) — @pimzand, since you have the scenario at hand and wrote such a precise report: could you verify this branch on your build? Expected: the rerun disappears from the new UI's Upcoming view (count matches classic), and opening the rerun's event in the EPG shows the amber "Will be skipped" note with the original broadcast date.

### DIFF
--- a/src/webui/static-vue/src/views/dvr/UpcomingView.vue
+++ b/src/webui/static-vue/src/views/dvr/UpcomingView.vue
@@ -21,6 +21,12 @@
  *                    start time is close enough for sorting purposes
  *                    and is what the grid endpoint returns directly.)
  *   - `sched_status` "scheduled" / "recording" / etc. Plain string.
+ *
+ * `duplicates: 0` mirrors the ExtJS grid (`dvr.js:508`): the server
+ * includes dedup-skipped reruns by default (`api_dvr.c`), and without
+ * the param they'd show as ordinary "Scheduled" rows although they
+ * will not record. The EPG event drawer explains the skip for such
+ * entries (its "rerun of" note), matching the classic details dialog.
  *   - `pri`          priority enum (server renders to a localized
  *                    string in the `pri` field for the grid view).
  *
@@ -449,6 +455,7 @@ function buildActions(selection: BaseRow[], clearSelection: () => void): ActionD
     help-page="class/dvrentry"
     :columns="cols"
     store-key="dvr-upcoming"
+    :extra-params="{ duplicates: 0 }"
     :default-sort="{ key: 'start_real', dir: 'ASC' }"
     :virtual-scroller-options="{ itemSize: 36, lazy: false }"
     count-label="recordings"
--- a/src/webui/static-vue/src/views/dvr/__tests__/UpcomingView.test.ts
+++ b/src/webui/static-vue/src/views/dvr/__tests__/UpcomingView.test.ts
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2026 Tvheadend contributors
+
+/* eslint-disable vue/one-component-per-file -- inline stub components
+ * for the grid and its heavy drawer/editor companions; each is a
+ * throwaway harness, not a real component. */
+
+/*
+ * UpcomingView — pins the `duplicates: 0` grid parameter (issue
+ * #2207). The server's grid_upcoming endpoint INCLUDES dedup-skipped
+ * rerun entries by default; without the param they render as ordinary
+ * "Scheduled for recording" rows although they will not record. The
+ * ExtJS Upcoming grid has always passed duplicates=0 (dvr.js) — this
+ * test keeps the Vue view from silently regressing to the default.
+ */
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { defineComponent, h } from 'vue'
+import { enableAutoUnmount, mount } from '@vue/test-utils'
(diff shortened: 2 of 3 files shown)


pimzand:
Verified on my production build, as asked. Full results are on the issue: https://github.com/tvheadend/tvheadend/issues/2207#issuecomment-5136199874 (a six-surface comparison of both UIs with #2208 installed).

tl;dr: both halves behave exactly as intended.

oliversluke:
Thanks @pimzand — that six-surface comparison is exactly the verification this needed, much appreciated.

Two follow-ups from it:

- You're right about the classic note's location: it's the **DVR** details dialog (`dvr.js:71`), not the EPG one — I had the wrong reference in a code comment here and have amended it (comment-only change, no behaviour delta).
- Your second observation (the EPG "will record" indicators overclaiming for skipped reruns, in both UIs, with nothing inviting the user to the drawer note) is a fair point and makes the greyed-out-behind-a-toggle idea for Upcoming genuinely useful rather than just cosmetic. Both are on the follow-up list — the Upcoming toggle as a client-only change, the EPG indicators needing a small additive server field (the schedule status string can't carry dup-ness), which both UIs could then consume. Will gladly take you up on the testing offer when those materialise.

Issue #2168, opened 2026-07-09
[FEATURE]: Smart autorecs: boolean expression matching as an opt-in mode on existing autorecs


pimzand:
### Description

**Motivation**

Autorec matching is a fixed, implicit AND over single-valued fields: one title regex, one channel *or* one channel tag, one content type, etc. (`dvr_autorec_cmp()`, `src/dvr/dvr_autorec.c:191`). Rules like these can't be expressed today:

- record on *any of* several channels or tags, but *not* on specific others
- OR-combine title patterns, or exclude a title pattern ("Cycling, but not `vrouwen`")
- match channel/tag *names* by pattern instead of a single fixed reference

The workaround is maintaining several near-duplicate autorecs, which drift apart and defeat dedup. This proposal adds "smart autorecs" (inspired by Navidrome's smart playlists): the same matching leaves, combined with explicit `all`/`any`/`not` operators, unlimited nesting.

**Proposal**

*One class, one new field, not a parallel subsystem.* `dvr_autorec_entry` gains a text property (say `expression`). When empty, behaviour is exactly today's. When set, `dvr_autorec_cmp()` delegates to an expression-tree evaluator and the legacy selection fields are ignored. Everything downstream (`de_autorec` back-pointers, purge-on-mismatch, dedup, priority/profile handling, HTSP autorec messages) is untouched because a smart autorec *is* an autorec.

Recording *settings* (name, directory, DVR config, priority, dedup, comment, …) stay ordinary fields; the expression only selects events.

*Format: JSON with comments.* Parsed by the existing `htsmsg_json` after a small comment-stripping prepass, so no new dependency. Validated on save; invalid expressions are rejected with a parse error, not stored.

```jsonc
{
  "all": [
    { "any": [
      // uuid wins when it resolves; name is the fallback
      { "channel": { "name": "NPO 1", "uuid": "8a7c…" } },
      { "tag":     { "name": "HD channels" } }
    ]},
    { "title": { "regex": "^Tour de France" } },
    { "not": { "title": { "regex": "vrouwen" } } }
  ]
}
```

Leaves mirror the existing predicates (title, channel, tag, content type, categories, duration/year/season ranges, star rating, broadcast type, weekday/time window, serieslink), plus name-pattern matching for channels/tags, which the flat model can't express. The expression is parsed once on load/save into a compiled tree (regexes precompiled, as `dae_title_regex` is today), so per-event cost stays comparable.

[...]

PR #2210, opened 2026-07-31
webui: reveal dedup-skipped reruns in the Upcoming view on demand

oliversluke:
> [!NOTE]
> #2208 has merged — this branch is now rebased onto master and carries the single commit described below.

### What

The follow-up discussed in #2207/#2208: with the Upcoming view now hiding dedup-skipped reruns (classic parity), this adds the opt-in way to *see* them — a **"Skipped reruns"** setting in the grid-settings popover's Filters section (`Hidden` by default / `Shown (dimmed)`).

When shown, skipped entries render **struck through and dimmed** — the classic UI's own duplicate treatment (`ext.css` `.x-epg-duplicate`) — with the status column reading **"Will be skipped"** (classic message id, so existing translations apply). Picking the non-default lights the Filters section's accent chip, and the choice persists per browser.

This makes Upcoming the first list in either UI that can show what the duplicate detection will *not* record — the classic grid hardcodes `duplicates=0` with no way back.

### How

- The setting rides the existing `GlobalFilterSpec` machinery (the same mechanism as the Services view's "Hide" select): the grid seeds the value into every fetch and refetches on change — no new fetch plumbing.
- Skipped rows are recognised by the entry's read-only `duplicate` field (present in every grid row; the schedule status itself still reads plain `"scheduled"` for a skipped rerun).
- New small piece of shared machinery: a per-row class hook threaded `IdnodeGrid → DataGrid` onto the desktop `<tr>` and the phone card root alike, so one resolver styles both layouts.
- The row styling is deliberately unscoped and anchored on the row class alone — `IdnodeGrid` is a multi-root template, so a fallthrough class on it never reaches the DOM, and scoped `[data-v]` anchors can't reach PrimeVue's rows two component levels down.

### Testing

- Unit tests pin: the filter spec defaults to hidden (ExtJS-parity fetch), the flip + per-browser persistence, the row classification, and the status-column override. Full Vue suite green; `vue-tsc`, ESLint, SPDX check, build all pass.
- Verified against a live instance with 20 real dedup-skipped entries (manufactured by scheduling under "record all" and then tightening the DVR profile's duplicate handling — the same "settings changed after entries existed" path that produces such entries in the wild): default view shows only the recording entries; the setting reveals the 20 skipped ones struck-through with correct "Will be skipped" statuses.
- @pimzand — you offered to test this follow-up; it composes with #2208 on this branch, so your existing setup should exercise both.

### RFC: the EPG side (your observation 2)

@pimzand's #2208 review noted that the EPG "will record" indicators (Table clock, Timeline/Magazine stripe — both UIs) overclaim for skipped reruns, and nothing invites the user to open the event and find the note. The schedule status string cannot carry dup-ness, so fixing that honestly needs a small **additive server field** on the EPG rows — e.g. `dvrDuplicate: <original broadcast's start>` next to `dvrUuid`/`dvrState` in the `api_epg` DVR join. The new UI would then render a distinct "skipped" state (dimmed stripe / distinct icon), and the classic UI could adopt the same field if wanted. Before building it: does that wire shape look right, or would you rather see dup-ness folded differently? Opinions welcome — that would be a third, server-touching PR.

### DIFF
--- a/src/webui/static-vue/src/components/DataGrid.vue
+++ b/src/webui/static-vue/src/components/DataGrid.vue
@@ -238,6 +238,12 @@ interface Props {
    * prop. */
   countLabel?: string
 
+  /* Optional per-row class resolver, applied to the desktop table's
+   * <tr> (PrimeVue's `rowClass`) AND the phone card root, so a
+   * consumer's row-state styling (e.g. DVR Upcoming's dedup-skipped
+   * dimming) covers both layouts from one function. */
+  rowClass?: (row: Row) => string | undefined
+
   /*
    * Active group-by field. When set, the table renders with
    * PrimeVue's subheader row-grouping mode — rows cluster by this
@@ -369,6 +375,7 @@ const props = withDefaults(defineProps<Props>(), {
   phoneItemSize: 88,
   sortLockedByGroup: false,
   countLabel: undefined,
+  rowClass: undefined,
   groupField: null,
   groupOrder: 'ASC',
   groupableFields: () => [],
@@ -1759,6 +1766,7 @@ defineExpose({
               `${bemPrefix}__card`,
               isRowSelected(row as Row) ? 'data-grid__card--selected' : null,
               isRowSelected(row as Row) ? `${bemPrefix}__card--selected` : null,
+              rowClass?.(row as Row) ?? null,
             ]"
             :style="{ height: `${phoneItemSize}px` }"
           >
@@ -1913,6 +1921,7 @@ defineExpose({
               `${bemPrefix}__card`,
               isRowSelected(item.row) ? 'data-grid__card--selected' : null,
               isRowSelected(item.row) ? `${bemPrefix}__card--selected` : null,
+              rowClass?.(item.row as Row) ?? null,
             ]"
           >
           <label
@@ -2018,6 +2027,7 @@ defineExpose({
         :value="entriesForTable"
         :loading="loading"
         :data-key="keyField"
+        :row-class="rowClass"
         :selection-mode="
           selectable === true ? 'multiple' : selectable === 'single' ? 'single' : undefined
         "
--- a/src/webui/static-vue/src/components/IdnodeGrid.vue
+++ b/src/webui/static-vue/src/components/IdnodeGrid.vue
@@ -205,6 +205,9 @@ interface Props {
    * → "748 services"). Falls back to "rows" when unsupplied.
    */
   countLabel?: string
+  /* Per-row class resolver, threaded to DataGrid (desktop <tr> and
+   * phone card root alike). See DataGrid's `rowClass` prop. */
+  rowClass?: (row: BaseRow) => string | undefined
   /*
    * Override DataGrid's default phone-mode card height (px)
    * when the phone path also virtualises (i.e.
@@ -430,6 +433,7 @@ const props = withDefaults(defineProps<Props>(), {
   notificationClass: undefined,
   filters: undefined,
   extraParams: undefined,
+  rowClass: undefined,
   virtualScrollerOptions: undefined,
   countLabel: undefined,
   phoneItemSize: undefined,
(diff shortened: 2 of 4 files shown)


pimzand:
Tested on a local build with this branch stacked on #2208, against real dedup-skipped entries. With the setting on its default the Upcoming row set matches the classic grid exactly; flipping Skipped reruns to Shown (dimmed) reveals the skipped entries struck through with "Will be skipped" statuses, the Filters chip lights up, and the choice survives a reload. The event drawer's rerun note agrees with the revealed rows. Works as described, nice that the classic conventions carry over.

The RFC on the EPG side deserves a considered answer rather than a quick nod; a fuller response follows shortly.

pimzand:
@oliversluke as promised, the fuller response on the EPG-side RFC, now that the other side of this story is in reviewable shape again.

First the context: [smart autorecs (PR #2186)](https://github.com/tvheadend/tvheadend/pull/2186) is out of draft and ready for review again. For anyone landing here without having followed the RFC in #2168: it adds an opt-in boolean expression mode to the existing autorec rules. Instead of the implicit AND over the rule's fields, a rule can carry one expression combining any-of / all-of groups and, crucially, negation, so it can say "this title, but not that channel, and not repeats titled like this". The primary aim is cutting the false positives a flat rule cannot express its way out of; a welcome side effect is that one smart rule can replace a whole stack of overlapping flat rules, with Duplicate handling trimming the broader match back to what should actually record. Since the draft round it has grown two things relevant here: converting a rule now leaves a disabled backup copy behind (so trying the feature is a safe, reversible step), and the match preview now shows the Duplicate handling verdicts, so a rule's preview no longer overstates what will record. Review and comments are very welcome, from maintainers and from you, it is a substantial change and could use more eyes.

The preview work is where the two efforts converge: the server runs the Duplicate handling logic over the preview candidates and flags the ones that would be skipped with `duplicate=1` (flagging rather than filtering, per the HTSP precedent), and the preview grid renders these exactly the way this PR does, muted line-through with the "Will be skipped because it is a rerun of" note as the tooltip. The preview answers "what would this rule skip" before saving, this PR answers "what will these existing entries do". Same verdict, same presentation, deliberately.

On the RFC itself: the wire shape looks right to me, and I have a direct interest in it. Precisely because a smart rule that replaces several flat rules matches a broader union of broadcasts and relies on Duplicate handling to trim it back, it produces more dedup-skipped entries than flat rules would, and today every EPG surface presents those as certain to record. An honest skipped state in the EPG (Table clock and status column included) makes that behaviour legible. On the shape: `dvrDuplicate: <original broadcast's start>` mirrors the existing read-only `duplicate` property on the DVR entry (which the drawer already loads in #2208), so the semantics are established; I would omit the field entirely when the entry will record rather than sending 0, and only compute it for entries in the scheduled state, which bounds the cost to one duplicate scan per visible scheduled entry, the same work the classic details dialog does on demand. One semantic note worth writing into the field's docs: the verdict is as-of-now. It is computed live from the entries' stored metadata, so it can flip between fetches as EPG updates land (a rerun becomes recognisable as a duplicate only once its episode data arrives), and the authoritative decision still falls at recording start. Clients should treat it as uncacheable display state, which is also the argument for computing it server-side rather than letting clients guess at dup-ness.

One coordination note before building it. The preview work in #2186 refactors the duplicate machinery in `dvr_db.c`: the master filter, mode resolution and per-mode field requirements move into helpers shared between `_dvr_duplicate_event` and the new preview scan. `_dvr_duplicate_event` is static, so the `api_epg` join will need a small exported accessor, and it would be a shame to grow one in parallel with that refactor. Suggest sequencing the server-touching PR after #2186, which can export the accessor for `api_epg` to consume. The Vue work is unaffected either way: #2208 is merged, and this PR has no file overlap with #2186.

pimzand:
One thing the long comment above should have opened with, so here it is on its own. @oliversluke, credit where it is due: the new interface is a huge step for the project, and the follow-up support since it landed sets a standard. Report an issue and a fix appears within days, the webroot redirect among many others, with the #2207 to #2208 to this PR arc only the latest example, each step with tests and classic-parity thinking. It also did something for me personally. Smart autorecs is an idea I had been sitting on for a long time without an interface to build it in; the flat autorec dialog had nowhere to put an expression editor, and the new UI finally did. The feature advertised above owes its existence to this work in a rather direct way.

PR #2213, opened 2026-08-01
ci: build aarch64 debs natively on arm64 runners

### LINKED ISSUE #2212: [BUG]: CI: aarch64 build jobs run emulated, adding a flaky QEMU setup step and hour-long build times
### Tvheadend Version

v4.3 (stable)

### Build Number

n/a (CI infrastructure; workflows as of master a502d063c)

### Platform

Other.. (Fedora 44)

### Architecture

ARMv8 (aarch64)

### Category

Other..

### Description

On the CI run for PR #2186, the job "Build on bookworm aarch64" failed after 19 seconds:

```
Unable to find image 'tonistiigi/binfmt:latest' locally
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": received unexpected HTTP status: 502 Bad Gateway
##[error]The process '/home/runner/work/_actions/uraimo/run-on-arch-action/v3/src/run-on-arch.sh' failed with exit code 125
```

Run: https://github.com/tvheadend/tvheadend/actions/runs/30647457143, job: https://github.com/tvheadend/tvheadend/actions/runs/30647457143/job/91213673067

The job died in QEMU emulation setup, before checking out or compiling anything. No change to the tree can cause or prevent this failure; it is a property of how the job is hosted.

The structural picture: the "Build on \<distro\> \<arch\>" matrix in `build-ci.yml` (and the same matrix in `build-cloudsmith.yml`) runs its 10 aarch64 legs under QEMU user-mode emulation on amd64 runners, via `uraimo/run-on-arch-action`. That hosting choice has two costs:

1. **An avoidable failure mode.** The action's setup unconditionally pulls and runs `tonistiigi/binfmt` from Docker Hub (`src/run-on-arch.sh`, `install_deps`), so every leg carries a third-party registry dependency that can kill the job before the build starts, as above.
2. **Emulation speed.** In the run linked above the emulated aarch64 legs took between 29 and 97 minutes; the three slowest jobs of the entire run were all aarch64. These legs set the wall-clock time of every CI run.

The same workflow already demonstrates the alternative. The almalinux jobs run a runner matrix that includes `ubuntu-24.04-arm`, GitHub's free native arm64 runners (`build-ci.yml` line 323), and in that same run the two native aarch64 almalinux builds finished in 1 and 2 minutes. The workloads are not identical, so that is not a clean ratio, but the gap is dominated by emulation.

Suggested minimal fix: select the runner by target architecture in the run-on-arch jobs of both workflows, so the aarch64 legs execute natively while everything else is untouched:

```diff
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
```

On an arm64 host, Docker runs the action's arm64v8-based distro containers natively, so the builds themselves stop being emulated. Two caveats, stated so the limits of the one-liner are clear:

- `run-on-arch-action` v3 still performs its binfmt setup unconditionally, so the Docker Hub pull (and its failure mode) remains until either the action learns to skip it on a matching host, or the aarch64 legs are given the almalinux treatment (a plain `container:` on the arm runner), which is a slightly larger rework and would remove the dependency entirely.
- The armv6/armv7 legs must stay where they are: GitHub's arm64 runners are Azure Cobalt 100 (Neoverse N2 family) without 32-bit AArch32 execution, so 32-bit containers cannot run natively there, and QEMU user-mode emulation is roughly the same speed on either host.

Deliberately out of scope: the "Build on Raspberry Pi" jobs use `pguyot/arm-runner-action` full-system emulation with pinned CPU models and faked cpuinfo, which looks intentional and is left alone. The `linux/arm64` buildx platform in `container-build.yaml` could also move to a native runner with a manifest merge, but that is a separable change.

The practical gain beyond the hour of wall clock: infrastructure failures on these legs are common enough that unrelated red already needs explaining away in PRs (see https://github.com/tvheadend/tvheadend/pull/2179#issuecomment-4995136080 for a different flavour, apt against EOL archives, which native runners would not fix). Removing the emulation-specific class of failure and shrinking the matrix tail makes a red leg mean something again.

Happy to submit the minimal fix as a PR if that helps. It has a convenient property: `build-ci.yml` is `pull_request`-triggered, and pull_request runs use the workflow file from the PR itself, so the PR's own checks would show the aarch64 legs building natively before anything is merged. The `build-cloudsmith.yml` half can only prove itself after merge, since that workflow triggers on pushes to master (and ignores `.github` paths, so the merge itself would not trigger it either; its first native run would come with the next code push).

### Logfile

https://github.com/tvheadend/tvheadend/actions/runs/30647457143/job/91213673067 (full log; the failure excerpt is quoted in the description)

### Reproduction

1. Trigger the build workflows on any branch or PR.
2. Observe each `aarch64` leg of the "Build on \<distro\> \<arch\>" matrix spending its first step installing QEMU binfmt handlers (a `docker run` of `tonistiigi/binfmt` pulled from Docker Hub) and then building under emulation for 29 to 97 minutes.
3. Intermittently, the Docker Hub pull fails and the leg dies before checkout, as in the linked run.

pimzand:
Fixes #2212

**What**

Move the aarch64 deb builds from QEMU emulation (uraimo/run-on-arch-action on x86 runners) to native arm64 runners (`ubuntu-24.04-arm`), in both `build-ci.yml` and `build-cloudsmith.yml`.

Rather than adding a new job, the existing `build-deb-native` job gains a `runner` matrix axis, the same pattern `build-el-rpm-native` already uses for its almalinux aarch64 legs. The ten aarch64 containers are `include:` entries carrying `runner: ubuntu-24.04-arm`, so they are added as standalone matrix combinations and every build step stays shared with the x86 legs; future changes to the build steps automatically cover all architectures. The only per-arch delta is the "Fix old debian apt" condition, rewritten with `endsWith()` so the `arm64v8/` images match alongside the plain and `i386/` ones.

The armv6/armv7 legs are untouched: run-on-arch-action stays for those, and `build-crosscompile` simply loses its aarch64 entries.

**Why**

As discussed in #2212: run-on-arch-action cannot run on arm64 hosts at all (its QEMU helper image is amd64-only, uraimo/run-on-arch-action#161), so the emulated aarch64 legs cannot simply be re-hosted on arm64 runners. Emulation also makes them the slowest part of every PR run (29 to 97 minutes per leg), and it adds a third-party failure mode: one aarch64 leg recently failed in 19 seconds because the binfmt helper pull got a 502 from Docker Hub before checkout even ran.

**Testing**

Dispatched on a fork with this matrix, all legs green:

- all ten aarch64 legs: 2.4 to 6.8 minutes each (against 29 to 97 minutes emulated), packages built with `ffmpeg_static yes` and uploaded as artifacts;
- `debian:stretch` and `i386/debian:stretch` control legs on `ubuntu-latest`, confirming the runner axis and the rewritten apt condition leave the x86 path unchanged;
- no sources.list changes were needed: archive.debian.org serves arm64 packages for stretch and buster, and ports.ubuntu.com still carries xenial arm64.

The `build-ci.yml` half demonstrates itself in this PR's own checks. The `build-cloudsmith.yml` half only runs on push to master, so it proves itself on the first post-merge push; its cloudsmith upload steps are carried over unchanged from the x86 legs.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -81,7 +81,9 @@ jobs:
     # Run steps on a matrix
     strategy:
       matrix:
-       arch: [ armv7, aarch64]
+       # aarch64 is not emulated here: it builds natively on arm64
+       # runners in build-deb-native below.
+       arch: [ armv7 ]
        distro: [ stretch, buster, bullseye, bookworm, ubuntu16.04, ubuntu18.04, ubuntu20.04, ubuntu22.04, ubuntu24.04 ]
        include:
          - arch: armv6
@@ -90,8 +92,6 @@ jobs:
            distro: buster
          - arch: armv6
            distro: bullseye
-         - arch: aarch64
-           distro: ubuntu_latest
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


pimzand:
Perversely, CI is failing on a leg that this PR did not touch (armv7).

pimzand:
There were ten "Waiting for status to be reported" entries, fallout from the job rename: the required status checks on master expect the old aarch64 job names, and no job reported under those names any more, so re-running CI could not clear them.

The force-push keeps the old job names for the native aarch64 legs, so the required checks should match again and branch protection needs no changes. The re-run also covers the unrelated armv7 leg that died pulling the binfmt image from Docker Hub.

pimzand:
A note on the earlier comment: it described check results on the previous head commit, and those vanished from the checks tab when the force-push replaced that commit. The run on the current head is [30704135615](https://github.com/tvheadend/tvheadend/actions/runs/30704135615).

That run has now completed with all 62 jobs green, and the timings show what this PR buys. Comparing the aarch64 deb legs against the same legs in the last CI run on master ([19310009049](https://github.com/tvheadend/tvheadend/actions/runs/19310009049)):

| aarch64 leg | QEMU (master) | native arm64 (this PR) |
|---|---:|---:|
| stretch | 45 min | 2 min |
| buster | 52 min | 2 min |
| bullseye | 67 min | 3 min |
| bookworm | 78 min | 4 min |
| ubuntu16.04 | 85 min | 6 min |
| ubuntu18.04 | 48 min | 2 min |
| ubuntu20.04 | 55 min | 2 min |
| ubuntu22.04 | 74 min | 3 min |
| ubuntu24.04 | n/a * | 3 min |
| ubuntu_latest | 83 min | 4 min |

\* the master run predates the ubuntu24.04 aarch64 leg.

In total the ten aarch64 deb builds went from roughly 9.8 hours of runner time to 31 minutes, with the slowest leg dropping from 85 to 6 minutes. The aarch64 builds are no longer anywhere near the critical path: the overall run time is now set by the remaining armv6/armv7 QEMU legs and the Raspberry Pi builds.

pimzand:
One follow-up I would like to tackle, assuming this PR lands: the fedora job still builds x86_64 only, while the almalinux job next to it already has a runner axis and builds aarch64 natively. The two jobs run nearly identical dnf/configure steps, so the idea is to move those steps into a shared script for the rpm based distros and give fedora the same runner axis. That would add native fedora aarch64 builds at the same low cost as the legs in this PR.

Fedora aarch64 is the platform I run and develop tvheadend on, so I have a personal interest in seeing breakage there caught by CI before it reaches my own builds.

PR #2219, opened 2026-08-04
ci: share the rpm build steps and add native Fedora aarch64 legs

pimzand:
#2213 straightened out the aarch64 builds for the deb-based distros: the build steps became shared across architectures through a runner matrix axis, and QEMU emulation was replaced with native arm64 runners. This is the follow-up announced there, giving the rpm-based distros similar treatment.

The fedora and almalinux jobs in build-ci.yml and build-cloudsmith.yml carried four near-identical copies of the dnf and configure steps, so any change to the rpm build had to be made four times. This folds them into one shared script, `support/ci-rpm-build.sh`, which detects Fedora vs the Enterprise Linux clones via `/etc/os-release` (EL additionally enables CRB and uses the `free/el/` RPM Fusion path). The jobs now only install git and zstd before checkout; everything else runs from the checked-out script.

With the build steps shared, the fedora job gets the same runner axis the almalinux job already has, adding native aarch64 legs (ubuntu-24.04-arm) for the recent releases: 43, 44 and rawhide. The matrix include entries add combinations rather than expanding the cross-product, so the existing x86_64 coverage is unchanged.

Behaviour notes:

- The fedora job names gain an arch suffix ("Build on native fedora:44 x86_64"), matching the almalinux naming. None of the renamed jobs are required status checks.
- The fedora artifact names gain the same suffix to stay unique across arches, and the fedora cache keys gain the runner name, costing each fedora leg a one-time cold static-deps rebuild.
- zstd is installed next to git because the full dependency install, which previously ran before the staticlib cache step and pulled zstd in transitively, now runs from the script after it. actions/cache includes the compression method in the cache version, so the restore needs zstd present just as the save does; most of the base images do not ship it.
- The parallel-build fallback to a single-threaded make, previously fedora-only, now also applies to the almalinux legs.
- A duplicated gcc-c++ in the dnf line and a duplicated --disable-libfdkaac_static in the configure line were dropped.

Separately, this also adds Fedora 43 and 44 to the bug report template's version dropdown, which currently stops at 42. That fix is independent of CI: both releases are out and in use, and affected users cannot currently select their version when filing a bug.

Verified with a full run of the CI workflow on my fork: https://github.com/pimzand/tvheadend/actions/runs/30863599610. All 16 rpm legs pass (9 + 3 fedora, 2 + 2 almalinux), the inner rpmbuild `%configure` receives the Vue flags via `tvh_extra_configure` and reports `vue_cache yes`, the staticlib cache restores on the almalinux and older Fedora legs, and all 16 rpm artifacts upload under unique names.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -45,6 +45,8 @@ body:
         - Fedora 40
         - Fedora 41
         - Fedora 42
+        - Fedora 43
+        - Fedora 44
         - Fedora Rawhide
         - FreeBSD (all)
         - Raspberry Pi OS (Buster)
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -293,23 +293,28 @@ jobs:
 
   build-rpm-native:
     needs: vue
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     continue-on-error: true
-    name: Build on native fedora:${{ matrix.releasever }}
+    name: Build on native fedora:${{ matrix.releasever }} ${{ matrix.runner == 'ubuntu-latest' && 'x86_64' || 'aarch64' }}
 
     strategy:
       matrix:
         releasever: ["37", "38", "39", "40", "41", "42", "43", "44", "rawhide"]
+        runner: ["ubuntu-latest"]
+        # the include entries add combinations (their runner value differs
+        # from the base axis), so only recent releases build on aarch64
+        include:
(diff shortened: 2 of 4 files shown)


chewitt (approved):
This looks like a nice simplification. Great work 👍🏻

chewitt:
@pimzand can you rebase this, it looks like it should be merged to further simplify CI tasks

Flole998:
Rebased

pimzand:
Rebased again onto current master. The rebase from 9 September was overtaken by the translation batch that landed a minute later, which is why the rebase check stayed red.

PR #2225, opened 2026-08-18
transcoding: fix the buffersink format options on FFmpeg 8 and 9

heitbaum:
The transcoder cannot tell its buffersink what the encoder wants, so the sink is
left unconstrained and the negotiated format is whatever the decoder produced.
Two independent causes, one commit each.

**1. `transcoding: init the filter sink after setting its options`**

The buffersink format options are not runtime options, and since libavutil 60
(FFmpeg 8.0) `av_opt_set*()` refuses them once the object carries
`AV_CLASS_STATE_INITIALIZED`:

```
Option 'pix_fmts' is not a runtime option and so cannot be set after the object has been initialized
```

`avfilter_graph_create_filter()` initializes the filter it creates, and
`tvh_context_open_filters()` applies the sink options afterwards, so every one of
them fails with `AVERROR(EINVAL)`:

```
transcode: 0004: 02:AAC-LATM: [mp2 => aac]: filters: failed to set option: 'channel_layouts'
```

libavfilter only gained the state flag the check reads in FFmpeg 8.0
(`avfilter_class.state_flags_offset`), and libavutil 59 logged the rejection but
carried on, which is why this was not seen on FFmpeg 7.x — see #1764.

Allocating the sink with `avfilter_graph_alloc_filter()` leaves it
uninitialized; `avfilter_init_dict()` is called once all the options have been
applied. That is the documented order for options that must be set before init,
and it works on every libavfilter the tree supports.

**2. `transcoding: set the sink formats through the array options`**

libavfilter 12 (FFmpeg 9.0) removed the int-list/string format options of
buffersink and abuffersink:

```
filters: failed to set option: 'pix_fmts' with error: 'AVERROR_OPTION_NOT_FOUND'
filters: failed to set option: 'ch_layouts' with error: 'AVERROR_OPTION_NOT_FOUND'
```

The array-type replacements — `pixel_formats`, `sample_formats`, `samplerates`,
`channel_layouts` — arrived in libavfilter 10.6.100 (FFmpeg 7.1) and are set with
`av_opt_set_array()`, so `AV_OPT_SET_ARRAY` is added as a third option kind,
carrying the element type where the other kinds carry a size.

This also repairs the audio sink from 10.6.100 onwards, where the old names still
exist: `asink_query_formats()` only consults the deprecated `sample_fmts` /
`sample_rates` lists when no new-style list is set at all, and `common_init()`
parses the deprecated `ch_layouts` string into the new `channel_layouts` array.
Setting the layout as a string therefore made libavfilter ignore the sample
format and the sample rate entirely.

Passing the layout as an `AVChannelLayout` also drops the describe/parse round
trip through a 64 byte string.

### Testing

Built `master` against FFmpeg 8.0.1 (lavfi 11.4.100) and a locally built FFmpeg
9.0.1 (lavfi 12.1.101). A graph assembled the way `tvh_context_open_filters()`
assembles it, asking for nv12 from yuv420p input and s16/44100 stereo from
fltp/48000 input through `aresample`:

| | lavfi 11.4.100 | lavfi 12.1.101 |
|---|---|---|
| before | video nv12 ok, audio came back **fltp/48000** | options refused, graph never opened |
| after | video nv12, audio **s16/44100** | video nv12, audio **s16/44100** |

Frames were pushed through and pulled back out in each case, and
`av_buffersink_set_frame_size()` still applies.

### DIFF
--- a/src/transcoding/transcode/audio.c
+++ b/src/transcoding/transcode/audio.c
@@ -287,7 +287,22 @@ tvh_audio_context_open_filters(TVHContext *self, AVDictionary **opts)
         return -1;
     }
 
-#if LIBAVCODEC_VERSION_MAJOR > 59
+#if TVH_BUFFERSINK_ARRAY_OPTS
+    int ret = tvh_context_open_filters(self,
+        "abuffer", source_args,                           // source
+        filters,                                          // filters
+        "abuffersink",                                    // sink
+        "channel_layouts", AV_OPT_SET_ARRAY,              // sink option: channel_layout
+        AV_OPT_TYPE_CHLAYOUT,
+        &self->oavctx->ch_layout,
+        "sample_formats",  AV_OPT_SET_ARRAY,              // sink option: sample_fmt
+        AV_OPT_TYPE_SAMPLE_FMT,
+        &self->oavctx->sample_fmt,
+        "samplerates",     AV_OPT_SET_ARRAY,              // sink option: sample_rate
+        AV_OPT_TYPE_INT,
+        &self->oavctx->sample_rate,
+        NULL);                                            // _IMPORTANT!_
+#elif LIBAVCODEC_VERSION_MAJOR > 59
     char ch_layout[64];
     av_channel_layout_describe(&self->oavctx->ch_layout, ch_layout, sizeof(ch_layout));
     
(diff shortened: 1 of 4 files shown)


uknunknown:
Can you check this PR https://github.com/tvheadend/tvheadend/pull/2117 and confirm if is fixing your problem?

PR #2226, opened 2026-08-21
intlconv: keep UTF-8 when iconv does not implement //TRANSLIT

BKPepe:
`intlconv_charset_id()` appends the GNU `//TRANSLIT` and `//IGNORE`
suffixes unconditionally. Implementations that do not know them reject
the whole charset name instead of ignoring the suffix — on musl,
`iconv_open()` fails with `EINVAL` — so every conversion fails and the
startup self-test aborts tvheadend.

This change:

- probes whether `//TRANSLIT` and `//IGNORE` are supported;
- keeps the original UTF-8 text when `//TRANSLIT` is unavailable;
- omits `//IGNORE` when it is unavailable;
- warns instead of aborting in the self-test when transliteration is
  unavailable, while still aborting when it is available but misbehaves;
- preserves the existing behaviour on glibc and GNU libiconv.

This matters most for recording file names: keeping the original UTF-8
is preferable to a lossy ASCII conversion, which on musl turns every
non-ASCII character into `*` and would collide for titles differing only
in accented characters.

## Testing

| | before | after |
|---|---|---|
| startup on musl | `iconv() routine is not working properly ((null)), aborting!` | `[WARNING] iconv() does not implement //TRANSLIT, non-ASCII text will be kept as UTF-8` |
| `intlconv_charset_id("ASCII", 1, 1)` | `"ASCII//TRANSLIT//IGNORE"`, cannot be opened | `NULL` |
| recording file name | `NULL`, `cleanup_filename()` fails | `Žluťoučký kůň úpěl` |

Built and run on Alpine with musl 1.2.6; the last two rows are from a
unit test linked against the real `intlconv.o` from both builds. Also
checked on OpenWrt/musl on a CZ.NIC Turris 1.x, where musl rejects both
`ASCII//TRANSLIT` and `ASCII//IGNORE` with `EINVAL`.

### DIFF
--- a/src/dvr/dvr_config.c
+++ b/src/dvr/dvr_config.c
@@ -135,6 +135,33 @@ dvr_config_find_by_list(htsmsg_t *uuids, const char *name)
  *
  */
 static int
+dvr_charset_is_ascii(const char *s)
+{
+  return s && (strcasecmp(s, "ASCII") == 0 ||
+               strcasecmp(s, "US-ASCII") == 0 ||
+               strcasecmp(s, "ANSI_X3.4-1968") == 0);
+}
+
+/*
+ * The charset a new configuration starts with.  Converting to ASCII
+ * without transliteration replaces every non-ASCII character with the
+ * same substitute, which makes unreadable and colliding file names, so
+ * fall back to UTF-8 there.  A charset the user picks is left alone.
+ */
+static const char *
+dvr_default_charset(void)
+{
+  const char *s = intlconv_filesystem_charset();
+
+  if (dvr_charset_is_ascii(s) && !intlconv_translit_supported()) {
+    tvhinfo(LS_DVR, "iconv() cannot transliterate, keeping UTF-8 in file "
+                    "names rather than converting to %s", s);
+    return NULL;
+  }
+  return s;
+}
+
+static int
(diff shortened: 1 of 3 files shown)


Flole998:
Am I understanding your changes correctly: Effectively you are now disabling that check, and later on silently fail during filename creation in case the implementation is not working as expected?

BKPepe:
@Flole998 sorry for the slow reply. You were looking at 8d5240f, which did skip the check. The PR was reworked since then.

`intlconv_test()` still runs at every start and aborts on any failure or unexpected result. The only new accepted output is `*lu*ou*k*K**`, which is musl's form of the `?lu?ou?k?K??` that glibc returns in the C locale.

The actual change is that `//TRANSLIT` and `//IGNORE` are omitted when `iconv_open()` rejects them, instead of causing all conversions to fail.

Filename conversion errors are still handled by the existing fallback, and the second commit adds the missing check for directory components, which could otherwise result in a `(null)` directory.

On musl, without `//TRANSLIT`, characters not representable in ASCII become `*`, so new DVR profiles use UTF-8 when the filesystem charset is ASCII.

The recording-thread self-join is a separate existing issue, fixed in #2262. The new directory check is one more way to reach it, so that PR should go in first.

PR #2230, opened 2026-08-27
epg & eit: Small fixes and updates

DonKult:
Hi,

as the dates on those commits might indicate, I still have a somewhat long-running personal fork with many changes I should publish and upstream… soon™. Fast forward a couple years and they still waited for a final brush up while new ones pill on top of them. :flushed:

Any ways, this set is certainly random, but splitting them up into individual branches seemed too procrastinationable as most commits in this repo are larger than this branch diff, so I skipped on that for now, but happy to do that now (as its public now) if need be.

btw: The ETSI standard v1.20.0 is not finalized yet if I read it right, but the last relevant changes for the code were made in 1.18.1 (most of it in 1.17.1 and before) so I cross-read the latest version instead of leaving it at 1.17.1 (which I had before today). Except the million bugs I probably introduced in the last minute changes, all this should be reasonably well tested for literal years now. No AI involved so far (but that seems no concern here; I don't mind being reviewed by one if it helps).

### DIFF
--- a/src/epg.c
+++ b/src/epg.c
@@ -2200,6 +2200,7 @@ int epg_genre_list_add ( epg_genre_list_t *list, epg_genre_t *genre )
     LIST_FOREACH(g2, list, link) {
       /* Already exists */
       if (g2->code == genre->code) return 0;
+      if ((g2->code & 0xF0) == genre->code) return 0;
     }
     while (g1) {
       /* Update a major only entry */
--- a/src/epg.c
+++ b/src/epg.c
@@ -1924,8 +1924,8 @@ epg_broadcast_t *epg_broadcast_deserialize
 // that relates more to broadcast content than what I call a "genre"
 // these will be handled elsewhere as broadcast metadata
 
-// Reference (Sept 2016):
-// http://www.etsi.org/deliver/etsi_en/300400_300499/300468/01.11.01_60/en_300468v011101p.pdf
+// Reference (June 2026):
+// https://www.etsi.org/deliver/etsi_en/300400_300499/300468/01.20.00_20/en_300468v012000a.pdf
 
 #define C_ (const char *[])
 static const char **_epg_genre_names[16][16] = {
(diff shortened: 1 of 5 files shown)


DeltaMikeCharlie:
>If we already have an entry in the genre list saying 04|3 (Soccer) we
>don't need to add 04|0 (Sport) as well. It is still okay to add 04|1
>(Special event like World Cup).
>
>Note that major-only genres like 04|0 are overridden if we receive
>specific entries later on in the following while-loop, so this commit
>just ensures we get the same list regardless of receiving order.

@DonKult - Thanks for your contribution.

What real-world issues are you attempting to resolve with this?  Are there any HTTP or HTSP clients that you know of that exhibit undesirable behaviour because of this?

My personal view is that removing multiple instances of the same genre code (as per your other commit) is prudent.  However, removing group genres when a sub-genre is also present may cause unforeseen issues.

I'd consider controlling this behaviour by adding either 1) a user accessible option [default=off] or 2) a developer accessible 'hack' (via the JSON EPG Grabber configuration file) to the EIT EPG Grabber module.  (Option 1) If this issue is restricted to a small number of users, they could then enable the alternate behaviour.  (Option 2) If a whole country or region encounters this issue, then a new EPG Grabber module can be added that enables the hack.

Others may disagree with my view.

DonKult:
> What real-world issues are you attempting to resolve with this? Are there any HTTP or HTSP clients that you know of that exhibit undesirable behaviour because of this?

The (classic) web UI shows the genre(s) in its details dialog and that looks rather silly if that ends up being e.g. "Sports, Sports/Soccer, Sports/Soccer, Sports/Soccer, Sports/Soccer", but no, nothing really breaks. The clients could perform deduplication here of course, but I don't see the benefit in carrying this duplication (sometimes, depending on input order) in the back end.

> My personal view is that removing multiple instances of the same genre code (as per your other commit) is prudent. However, removing group genres when a sub-genre is also present may cause unforeseen issues.

This is already the case in the existing code, as I tried to mention in the commit message, but it expects that the generic code is sent before the specific one. So if you got "04|0 04|3" than it would be rewritten to "04|3", but if you got "04|3 04|0" instead it was saved as "04|0 04|3" (generic before specific, the list is sorted). See the unchanged code prefixed with the comment "Update a major only entry". I don't think this order-dependency is desirable, especially if multiple event descriptions are merged into one in "random" discovery order (e.g. if you have multiple services providing the same channel)

DeltaMikeCharlie:
Understood, thanks.

PR #2233, opened 2026-09-03
mpegts: don't move an untuned mux to the active scan queue

chewitt:
## Problem

A DVB network scan can stall permanently, leaving some muxes never scanned. The
setup wizard's progress bar sticks below 100% (80% in my case) and the scan never
resumes — not after the 120s safety re-arm, not ever. Nothing is logged.

## Cause

`mpegts_mux_set_tsid()` calls `mpegts_network_scan_mux_reactivate()` every time a
mux's TSID changes. The TSID is frequently learned from the NIT of a *different*
mux, so this also fires for muxes that are merely sitting in the scan queue and
have never been tuned.

`mpegts_network_scan_mux_reactivate()` then takes the mux out of `mn_scan_pend`
and puts it into `mn_scan_active` with no subscription, no `mm_active` and no
`mm_scan_timeout` armed. `mpegts_network_scan_timer_cb()` only walks
`mn_scan_pend` and `mn_scan_ipend`, so the mux is never tried again; and since
nothing is left to call `mpegts_mux_scan_done()` for it, it never returns to
`MM_SCAN_STATE_IDLE`. It sits in ACTIVE for good. As the wizard progress is
derived from the count of muxes not in the idle state, the scan looks frozen.

The unconditional reactivation goes back to cec33bc6e (issue #4942). Note that
`reactivate()` early-returns when the mux is already ACTIVE, so its only live
effect today is on muxes that are *not* being scanned — precisely the case that
breaks.

## Reproducer

uk-Oxford from dtv-scan-tables, DVB-T, reproducible on every run.

The NIT reports 634.167/657.833/601.833 MHz, further than the 2 kHz tolerance in
`dvb_network_find_mux()` from the configured 634/658/602, so new mux objects are
created for those (TSID supplied at construction, so no `set_tsid()` call). But it
reports 538000000 and 554000000 exactly, so the queued 538MHz and 554MHz muxes are
matched and updated through `mpegts_mux_set_onid()`/`mpegts_mux_set_tsid()` while
674MHz is still being scanned. Both are orphaned. Eight of the ten muxes complete
and the scan stops at 80%.

## Changes

1. **mpegts: don't move an untuned mux to the active scan queue** — the fix. Only
   reactivate a mux which is really tuned, and arm its scan timeout so the scan
   can complete.
2. **mpegts: trace the silent scan-queue skips** — the early returns in
   `mpegts_network_scan_do_mux()` and `mpegts_network_scan_mux_reactivate()` say
   nothing, so a mux dropping out of the scan leaves no trace at all. Diagnosing
   this needed reconstructing the queue order from mux UUIDs to work out which
   entry the scanner had passed over.
3. **mpegts: rescue muxes stuck in the active scan queue** — defensive. Sweep
   `mn_scan_active` for muxes with no `mm_active` and no armed `mm_scan_timeout`
   and requeue them, so a bug of this shape degrades into a warning and a retry
   instead of a scan that hangs in silence. Adds `mtimer_armed()`.

Patches 2 and 3 are independent of the fix and can be dropped if unwanted.

## Testing

Before: 8/10 muxes scanned, stuck at 80% indefinitely.

After: all ten muxes scan in ~80s, 216 services found, OTA EIT grab follows
normally. The new trace shows the guard declining exactly the two affected muxes:

```
12:28:58 mpegts: 538MHz in DVB-T Network - scan not reactivated, mux is not tuned
12:29:03 mpegts: 554MHz in DVB-T Network - scan not reactivated, mux is not tuned
```

The watchdog in patch 3 did not fire, which confirms patch 1 covers the whole root
cause. Its requeue path is therefore not exercised by this reproducer.

Tested on LibreELEC (aarch64, Amlogic) with a Sony CXD2837ER DVB-T/T2/C tuner.

### DIFF
--- a/src/input/mpegts/mpegts_mux.c
+++ b/src/input/mpegts/mpegts_mux.c
@@ -89,21 +89,21 @@ mpegts_mux_instance_create0
 }
 
 static void
+mpegts_mux_scan_timeout_arm ( mpegts_mux_t *mm, mpegts_input_t *mi )
+{
+  mtimer_arm_rel(&mm->mm_scan_timeout, mpegts_mux_scan_timeout, mm,
+                 sec2mono(mpegts_input_grace(mi, mm)));
+}
+
+static void
 mpegts_mux_scan_active
   ( mpegts_mux_t *mm, const char *buf, mpegts_input_t *mi )
 {
-  int t;
-
   /* Setup scan */
   if (mm->mm_scan_state == MM_SCAN_STATE_PEND ||
       mm->mm_scan_state == MM_SCAN_STATE_IPEND) {
     mpegts_network_scan_mux_active(mm);
-
-    /* Get timeout */
-    t = mpegts_input_grace(mi, mm);
-  
-    /* Setup timeout */
-    mtimer_arm_rel(&mm->mm_scan_timeout, mpegts_mux_scan_timeout, mm, sec2mono(t));
+    mpegts_mux_scan_timeout_arm(mm, mi);
   }
 }
 
@@ -1356,7 +1356,8 @@ mpegts_mux_set_tsid ( mpegts_mux_t *mm, uint32_t tsid, int force )
   mm->mm_tsid = tsid;
   tvhtrace(LS_MPEGTS, "%s - set tsid %04X (%d)", mm->mm_nicename, tsid, tsid);
   idnode_changed(&mm->mm_id);
-  mpegts_network_scan_mux_reactivate(mm);
+  if (mpegts_network_scan_mux_reactivate(mm))
+    mpegts_mux_scan_timeout_arm(mm, mm->mm_active->mmi_input);
   return 1;
 }
 
--- a/src/input/mpegts/mpegts_network_scan.c
+++ b/src/input/mpegts/mpegts_network_scan.c
@@ -92,7 +92,10 @@ mpegts_network_scan_do_mux ( mpegts_mux_queue_t *q, mpegts_mux_t *mm )
          state == MM_SCAN_STATE_ACTIVE);
 
   /* Don't try to subscribe already tuned muxes */
-  if (mm->mm_active) return 0;
+  if (mm->mm_active) {
+    tvhtrace(LS_MPEGTS, "%s - scan skipped, mux is already tuned", mm->mm_nicename);
+    return 0;
+  }
 
   /* Attempt to tune */
   r = mpegts_mux_subscribe(mm, NULL, "scan", mm->mm_scan_weight,
@@ -292,8 +295,10 @@ mpegts_network_scan_mux_reactivate ( mpegts_mux_t *mm )
    * is orphaned - it is taken out of the scan queues, but nothing is able
    * to finish the scan for it, so it never returns to the idle state.
    */
-  if (mm->mm_active == NULL)
+  if (mm->mm_active == NULL) {
+    tvhtrace(LS_MPEGTS, "%s - scan not reactivated, mux is not tuned", mm->mm_nicename);
     return 0;
+  }
   mpegts_network_scan_queue_del0(mm);
   mm->mm_scan_init  = 0;
   mm->mm_scan_state = MM_SCAN_STATE_ACTIVE;
--- a/src/input/mpegts/mpegts_network_scan.c
+++ b/src/input/mpegts/mpegts_network_scan.c
@@ -282,16 +282,23 @@ mpegts_network_scan_mux_active ( mpegts_mux_t *mm )
 }
 
 /* Mux has been reactivated */
-void
+int
 mpegts_network_scan_mux_reactivate ( mpegts_mux_t *mm )
 {
   mpegts_network_t *mn = mm->mm_network;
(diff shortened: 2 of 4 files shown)


dave-p:
Confirmed working on DVB-S Astra 2 28.2E. Nice work Claude!

Flole998:
I think it was too early to merge this, looking at the changes it doesnt look right, the whole reactivate stuff seems weird

chewitt:
There were three commits, it's possible the second/third are not required and can be reverted without removing the fix from the first one. Let me know if that's required.

Flole998:
The second commit is fine, about the other 2 I am not sure. The whole logic seems flawed right now, hoping that the mux is tuned while its reactivated and then putting it in the active state.... Someone needs to either go through how the scan (normally) works, or ask AI to sum it up and to explain how the new changes work in that context (and I believe thats the point where AI will notice that further adjustments are needed).

chewitt:

@Flole998 — walkthrough as asked, then one question that's yours to answer.

## How the scan works

`mm_scan_state` and queue membership are kept in lockstep by `queue_del0()`: IDLE → in no queue, PEND → `mn_scan_pend`, IPEND → `mn_scan_ipend`, ACTIVE → `mn_scan_active`.

`mpegts_network_scan_timer_cb()` walks **only** pend and ipend. Nothing ever re-tunes an entry in `mn_scan_active`. The only two routes out of ACTIVE are:

1. `mm_scan_timeout` → `mpegts_mux_scan_timeout()` → `mpegts_mux_scan_done()`
2. `mpegts_table_fastswitch()` → `mpegts_mux_scan_done()`, once every `MT_QUICKREQ` table is complete — which requires the mux to be delivering data

So the invariant is: **a mux in `mn_scan_active` must be tuned, or on a timer, or it is stuck for good.** `mpegts_mux_scan_done()` is also the only thing that puts the mux back to IDLE, and wizard progress counts muxes not in IDLE — hence a scan that looks frozen.

## Where `reactivate()` actually bites

This is the part that reads as weird until you work out its live domain. Every tune goes through `mpegts_mux_instance_start()`, which calls `mpegts_mux_scan_active()` and promotes PEND/IPEND → ACTIVE at tune time. So when `mpegts_mux_set_tsid()` fires, a **tuned** mux is in one of:

- **ACTIVE** — `reactivate()` early-returns. No effect at all.
- **IDLE** — tuned by a streaming/EPG subscription, never queued for a scan.
- **PEND/IPEND** — finished a scan (`mux_done0` dropped the "scan" subscription, but a streaming subscription keeps the mux tuned) and was then requeued for idlescan.

In the last two the mux **is** receiving data, so re-running the scan after a TSID change is both meaningful and completable — `fastswitch` may even finish it early. That looks like exactly what #4942 wanted ("switch back to the scan state when new TSID is detected").

The case aee318393 removed is the fourth one: **not tuned and not scanning**. There the promotion can never complete, which is the reported stall. So to your wording — it doesn't hope the mux is tuned, it checks `mm_active`, and the armed timeout means even a mux that stops delivering still leaves ACTIVE.

## The two things worth double-checking in the merged code

- **`mtimer_armed()`** (commit 3): `mti_callback` is cleared *before* the callback runs, so it reads 0 for a timer in flight. Not a problem here — there is one mtimer thread and it holds `global_lock` for the whole of `scan_timer_cb`, so no `mm_scan_timeout` can be mid-dispatch or armed concurrently.
- **the watchdog's requeue weight**: `SUBSCRIPTION_PRIO_SCAN_INIT` is 5, below `SUBSCRIPTION_PRIO_MIN` (10), so a rescued mux can never preempt a user subscription.

The Copilot comment (add `mm_active->mmi_input` to the guard) isn't needed: `mpegts_input_grace()` is explicitly NULL-tolerant and falls back to a 5 s floor, and `mm_active` is already guaranteed non-NULL there.

## What does need fixing

Your instinct was right about one thing structurally. aee318393 split the invariant across two files — `reactivate()` performs the transition, the caller in `mpegts_mux.c` has to remember to arm the timeout from the return value. A second caller that ignored it would silently recreate the orphan.

#2247 moves the arming inside `reactivate()`, next to the state change and the queue insert, and restores the `void` return. Same guard, same muxes, same grace period, only the call site moves — so the hardware results here still stand.

## The question

`reactivate()` could also just be **deleted outright**, reverting to pre-#4942 behaviour: `set_tsid()` would stop touching the scan state entirely.

Given the enumeration above, its only live effect is pulling a tuned-but-not-scanning mux into the scan machine on a TSID change. That is arguably useful — a mux whose TSID changed has a different identity and its services want re-enumerating — but it does mean a mux tuned purely for streaming gets a `scan complete` / `scan timed out` log line and its `mm_scan_result` overwritten, which nobody asked for.

I'd keep it, because I can't reproduce or re-read #4942 and don't want to regress a fix I can't test. But you may remember what it was for. If you'd rather it went away, say so and I'll send that instead — #2247 would then be moot and commit 3's watchdog becomes the only thing standing between this bug shape and a silent stall.

dave-p:
Overtaken by events, but I've confirmed that just the first commit is sufficient to fix the problem of the setup wizard hanging during scan.

PR #2242, opened 2026-09-08
webui: make Auto the default theme

chewitt:
Auto answers with whatever the host is set to, which is what most users want and what none of them had to ask for until now: the default was Light, so every install started on a fixed palette and following the OS was an opt-in that had to be found in Configuration → General.

This inverts that. `config_boot()` defaults `theme_ui` to `"auto"` and `access_get_theme()`'s two nothing-configured fallbacks return it too, so a fresh install follows the host and only a user who wants to ignore that setting — pinning Light or Dark, or needing the high-contrast Access palette — has to pick a theme at all. Auto also leads `theme_get_ui_list()`, so the dropdown opens on the default rather than burying it third.

Nothing downstream changes. `"auto"` still never reaches the DOM: the access store resolves it to `light` or `dark` before setting `data-theme`, and the legacy interface has no `xtheme-auto.css` any more than it had an `xtheme-light.css`, so it keeps falling back to blue exactly as before.

### Existing installs are left alone

They already have `theme_ui` written to disk, usually `"light"` because that was the old default, and no migration rewrites it — the stored value cannot distinguish a deliberate choice of Light from the default nobody touched, and silently repainting a working interface is the worse of the two mistakes. Upgraders switch to Auto by hand if they want it.

### Known consequence

The pre-Comet flash is now the common path rather than a corner case. Before the first `accessUpdate` lands there is no `data-theme` yet and the document renders the `:root` Light palette, so every untouched install on a dark host sees a brief light flash on cold load. Deciding the palette before the websocket answers needs a pre-paint script or a persisted last-resolved theme; that is left for its own change and noted in the store comment.

### Testing

Builds clean. The theme/store/scale Vue test files pass (38 tests), and `type-check` and `lint` are clean.

Follow-up: `intl/tvheadend.pot`'s source-line references for the Light/Dark/Auto strings are stale by a line or two after the reorder (comments only, no string changes), to be picked up by the next template regeneration.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,7 +1,7 @@
 Source: tvheadend
 Section: video
 Priority: extra
-Maintainer: Tvheadend Project <admin@tvheadend.org>
+Maintainer: Tvheadend Project <debian@tvheadend.org>
 Build-Depends: debhelper (>= 7.0.50), pkg-config, gettext, libavahi-client-dev, libssl-dev | libssl1.0-dev, zlib1g-dev, wget, bzip2, git-core, liburiparser-dev, python | python3, python-requests | python3-requests, ca-certificates, cmake, libpcre2-dev | libpcre3-dev, libdvbcsa-dev
 Standards-Version: 3.7.3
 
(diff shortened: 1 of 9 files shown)


Flole998:
@chewitt Lets do debian@tvheadend.org here as E-Mail address

PR #2247, opened 2026-09-09
mpegts: arm the scan timeout where the mux becomes active

chewitt:
Follow-up to #2233, addressing @Flole998's review comment there. Structural only — no behavioural change.

## The invariant

`mm_scan_state` and queue membership are kept in lockstep by `queue_del0()`. `mpegts_network_scan_timer_cb()` walks **only** `mn_scan_pend` and `mn_scan_ipend`, so nothing ever re-tunes an entry in `mn_scan_active`, and the only two routes out of ACTIVE are:

1. `mm_scan_timeout` → `mpegts_mux_scan_timeout()` → `mpegts_mux_scan_done()`
2. `mpegts_table_fastswitch()` → `mpegts_mux_scan_done()`, which needs the mux to be delivering data

So: **a mux put into `mn_scan_active` must be tuned, or on a timer, or it is stuck for good.**

## What's wrong with how #2233 left it

aee318393 made `mpegts_network_scan_mux_reactivate()` return whether it promoted the mux and left the arming to its caller in `mpegts_mux_set_tsid()`:

```c
if (mpegts_network_scan_mux_reactivate(mm))
  mpegts_mux_scan_timeout_arm(mm, mm->mm_active->mmi_input);
```

That splits one invariant across two files. A second caller that ignored the return value would reintroduce exactly the orphaned mux aee318393 set out to prevent, and nothing would flag it.

## The change

Arm the timeout inside `reactivate()`, next to the state change and queue insert it belongs to, and give the function its `void` return back. `mpegts_mux_scan_timeout_arm()` leaves static scope for that, carrying a comment that states the invariant so the next transition into the active queue has something to be measured against.

`mpegts_network_scan_mux_active()` deliberately keeps arming from its caller in `mpegts_mux_scan_active()` — there the transition and the arm are adjacent lines of one function, where the pairing is hard to miss.

## Equivalence

The same guard promotes the same muxes and arms the same timeout with the same grace period, computed from the same `mm_active->mmi_input`. Only the call site moves, and there is no code between it and the return it replaces. So the uk-Oxford DVB-T result in #2233 and @dave-p's DVB-S confirmation both still apply.

## On the Copilot review comment in #2233

It suggested folding `mm_active->mmi_input != NULL` into the guard. Not needed: `mpegts_input_grace()` is explicitly NULL-tolerant (`if (mi && mi->mi_get_grace)`) and falls back to its 5 s lower bound, and `mm_active` itself is already guaranteed non-NULL on that path. Documented on the helper rather than guarded twice.

## Testing

Builds clean. No tuner on this machine, so the scan itself is not re-tested here — the argument for this PR is the equivalence above rather than a fresh hardware run.

### DIFF
--- a/src/input/mpegts.h
+++ b/src/input/mpegts.h
@@ -973,6 +973,7 @@ th_subscription_t *mpegts_mux_find_subscription_by_name(mpegts_mux_t *mm, const 
 void mpegts_mux_unsubscribe_linked(mpegts_input_t *mi, service_t *t);
 
 void mpegts_mux_scan_done ( mpegts_mux_t *mm, const char *buf, int res );
+void mpegts_mux_scan_timeout_arm ( mpegts_mux_t *mm, mpegts_input_t *mi );
 
 void mpegts_mux_bouquet_rescan ( const char *src, const char *extra );
 
--- a/src/input/mpegts/mpegts_mux.c
+++ b/src/input/mpegts/mpegts_mux.c
@@ -88,7 +88,18 @@ mpegts_mux_instance_create0
   return mmi;
 }
 
-static void
+/*
+ * Arm the scan timeout. Nothing else can take a mux out of
+ * MM_SCAN_STATE_ACTIVE on its own: mpegts_mux_scan_done() is reached
+ * either from here or from mpegts_table_fastswitch() once every
+ * MT_QUICKREQ table is complete, and the latter needs the mux to keep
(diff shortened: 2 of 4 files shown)


Flole998:
The idea back then was, that if a channel is being watched and the TSID of the mux changes the scan is triggered automatically. So reactivate should only trigger if the mux is already tuned (so it should be delivering data, but of course adding a timeout is also always a good idea).

PR #2248, opened 2026-09-10
tsfix: fix DTS normalisation across the 33 bit timestamp wrap

### LINKED ISSUE #1876: [BUG]: DTS = INT64_MIN causes stream failure during Timeshift
### Tvheadend Version

v4.3 (development)

### Build Number

HTS Tvheadend 4.3.0~pre+202507271709-0~built202507271817~gitf1c460feb~ubuntu22.04.1

### Platform

Ubuntu 22.04 (Jammy)

### Architecture

Intel x86_64

### Category

Timeshifting

### Description

**Description**
After extended uptime of TVHeadend (typically >18 hours), the Timeshift feature begins to produce transport stream packets with invalid DTS values of -9223372036854775808 (INT64_MIN). This results in repeated DTS discontinuity warnings in the logs and causes Kodi to freeze or terminate the stream.

**Observed Behavior**
Kodi playback halts without explicit error messages. The TVHeadend log shows entries like:

_[WARNING]:tsfix: transport stream MPEG2VIDEO, DTS discontinuity. DTS = -9223372036854775808, last = 4992815426
[WARNING]:tsfix: transport stream MPEG2AUDIO, DTS discontinuity. DTS = -9223372036854775808, last = 4992816866
[WARNING]:tsfix: transport stream AC3, DTS discontinuity. DTS = -9223372036854775808, last = 4992816866
[WARNING]:tsfix: transport stream TELETEXT, DTS discontinuity. DTS = -9223372036854775808, last = 4992817302
[WARNING]:tsfix: transport stream TEXTSUB, DTS discontinuity. DTS = -9223372036854775808, last = 4992610966_

The DTS value -9223372036854775808 corresponds to INT64_MIN, which appears to be either uninitialized or the result of an internal overflow.

**Expected Behavior**
TVHeadend should never emit DTS values of INT64_MIN. Timestamps should remain consistent and valid, even after prolonged uptime or heavy Timeshift usage.

**System Info**
TVHeadend Version: [4.3.0~pre+202507271709-0~built202507271817~gitf1c460feb]
Kodi Version: [20.4 Nexus]
Operating System: [Ubuntu 22.04 LTS]

**Timeshift Settings:**
Enabled: Yes
Max Time: 600 minutes
Max Size: 100000 MB
Storage: RAM + Disk

**Suspected Cause**
An internal DTS counter may overflow or reset incorrectly after prolonged operation. The use of INT64_MIN suggests an uninitialized or invalid timestamp being passed through the stream pipeline.

### Logfile

[WARNING]:tsfix: transport stream MPEG2VIDEO, DTS discontinuity. DTS = -9223372036854775808, last = 4992815426
[WARNING]:tsfix: transport stream MPEG2AUDIO, DTS discontinuity. DTS = -9223372036854775808, last = 4992816866
[WARNING]:tsfix: transport stream AC3, DTS discontinuity. DTS = -9223372036854775808, last = 4992816866
[WARNING]:tsfix: transport stream TELETEXT, DTS discontinuity. DTS = -9223372036854775808, last = 4992817302
[WARNING]:tsfix: transport stream TEXTSUB, DTS discontinuity. DTS = -9223372036854775808, last = 4992610966

### Reproduction

1. Start TVHeadend with Timeshift enabled

2. Let it run continuously for 18+ hours

3. Use Timeshift (pause, rewind, etc.)

4. Observe DTS discontinuity warnings and stream failure

chewitt:
`normalize_ts()` derives each packet's normalised DTS as an offset from a reference clock latched once when the subscription starts:

```c
dts = pts_diff(ref, pkt->pkt_dts);
```

`pts_diff()` returns `PTS_UNSET` (`INT64_MIN`) whenever `b < a`, unless its wrap heuristic `b < PTS_MASK/4 && a > PTS_MASK/2` fires — and that heuristic only recognises a 33 bit wrap when the *reference* sits in the top half of the range.

So once the input DTS wraps, which it must every 26.5 hours, any reference below `2^32` yields `PTS_UNSET`, and it is used unchecked as the normalised DTS. That is the value in the reported logs. It also poisons `tfs_last_dts_norm`, so every subsequent packet fails the negative timestamp test further down and is dropped for the remaining life of the subscription — which is why the stream stays dead until a retune.

The `tfs_dts_epoch` machinery in `normalize_ts()` already exists to compensate for exactly this wrap, but it expects the offset to step backwards by `2^33`. `pts_diff()` never produces that. The plain subtraction that preceded 433bcd2e5 did, so that commit left the epoch handling dead for this path.

## Change

Compute the offset modulo `2^33`. A sweep of ~2.9M `(ref, offset)` pairs confirms this returns **bit-identical values wherever `pts_diff()` returned anything at all**, and additionally hands the epoch code the backwards step it looks for. Since a timestamp preceding the reference now folds to the top of the range instead of going negative, the first-packet test becomes `dts > PTS_MASK / 2`.

`pts_diff()` itself is deliberately untouched. All eight other callers (`iptv.c`, `parsers/message.c`, `parsers/parsers.c`, `muxer_mkv.c`) compare timestamps that lie close together and every one checks the result against `PTS_UNSET` before use. `normalize_ts()` was the only caller using the result unchecked against a fixed reference across a full 26.5 hour span.

The `#5252` case that motivated 433bcd2e5 still works: a stream starting across the wrap gets continuous positive offsets, and packets preceding the reference are still dropped. The only first-packet decisions that differ require the reference clock itself to land within a fraction of a second of the wrap point, where dropping the pre-reference packet is the intended behaviour anyway.

## Verification

The arithmetic of `normalize_ts()` was replicated offline and swept across reference-clock values, simulating 40 hours of stream at 40ms packets:

```
reference clock       current pts_diff()   modular offset
           0             ok                     ok
  1000000000   broken after    23.43h           ok
  2200000000   broken after    19.72h           ok
  3597119165   broken after    15.41h           ok
  4294967296   broken after    19.88h           ok
  6000000000   broken after    14.62h           ok
  8000000000   broken after     8.45h           ok
  8589000000   broken after     6.63h           ok

trace for ref=3597119165:
  current: t= 15.41h DTS discontinuity. DTS = -9223372036854775808, last = 4992814800
  modular: t= 26.51h wrap detected, epoch -> 8589934592
```

The reported log line is `DTS = -9223372036854775808, last = 4992815426`; the simulated `last` differs only by the packet granularity used. @DeltaMikeCharlie's `last = 3461472442` is 10.7 hours, matching his "chase-played for the next 10+ hours".

Time to failure is `(2^33 - reference) / 90000`, i.e. **6.6 to 26.5 hours** depending on where the reference clock happens to land. That is why this reproduces reliably but never at a predictable time, and never in a short session.

Not actually timeshift-specific: `tsfix` sits upstream of timeshift in the pipeline, so timeshift is only what keeps a single subscription alive long enough. Long recordings are affected identically.

This is also a regression rather than a long-standing bug: 433bcd2e5 (2018-10-12) was cherry-picked to the 4.2 branch as 6590060c3 and shipped in v4.2.8; v4.2.7 and earlier are unaffected.

Builds clean. A live end-to-end confirmation needs 6.6+ hours of continuous streaming, so that has not been done — testers welcome.

Supersedes #1931, which was reverted in #2049. That approach patched the `PTS_UNSET` at the output and substituted `dts = 0`, which trades the stall for a corrupted timeline; this fixes the ambiguous input instead.

Fixes: #1876

### DIFF
--- a/src/plumbing/tsfix.c
+++ b/src/plumbing/tsfix.c
@@ -99,6 +99,18 @@ tsfix_ts_diff(int64_t ts1, int64_t ts2)
 }
 
 /**
+ * Offset of a timestamp from the 33 bit reference clock, folded into
+ * [0, PTS_MASK]. The wrap of the result is compensated by the
+ * tfs_dts_epoch handling in normalize_ts().
+ */
+static int64_t
+tsfix_ts_offset(int64_t ref, int64_t ts)
+{
+  return (ts - ref) & PTS_MASK;
+}
+
+
+/**
  *
  */
 static void
@@ -238,12 +250,13 @@ normalize_ts(tsfix_t *tf, tfstream_t *tfs, th_pkt_t *pkt, int backlog)
 
   /* Subtract the transport wide start offset */
   if (tf->dts_offset_apply)
-    dts = pts_diff(ref, pkt->pkt_dts + tf->dts_offset);
+    dts = tsfix_ts_offset(ref, pkt->pkt_dts + tf->dts_offset);
   else
-    dts = pts_diff(ref, pkt->pkt_dts);
+    dts = tsfix_ts_offset(ref, pkt->pkt_dts);
 
   if (tfs->tfs_last_dts_norm == PTS_UNSET) {
-    if (dts < 0 || pkt->pkt_err) {
+    /* Timestamps before the reference clock fold to the top of the range */
+    if (dts > PTS_MASK / 2 || pkt->pkt_err) {
       /* Early packet with negative time stamp, drop those */
       tsfix_packet_drop(tfs, pkt, "negative/error");
       return;


DeltaMikeCharlie:
@chewitt - I will test this PR when time allows, after I have rebuilt part of my test hardware.  This issue was last investigated about a year ago and my setup has changed.  Also, with a cycle time of over a day, this could take a while to catch a timestamp wrap and confirm that the bug has been fixed in TVH.

chewitt:
@DeltaMikeCharlie the bug has existed for years, so another few days isn't an issue :)

DeltaMikeCharlie:
I appear to have had 1 successful test run.

I ran 2 DVB-T systems (reference and patched) in parallel, tuned to the same channel, both time shifting via a Kodi client.

As expected, the reference system eventually encountered issues and the Kodi log shows a timeout/restart sequence.  The TVH log also shows a DTS discontinuity error 10 seconds prior and then a session reconnect in line with the timestamp in the Kodi log.

At the same timestamp on the patched system, Kodi showed no issues and appears to have remained connected.  After more than 27 hours of time shifting, the patched system has not reported any unexpected errors.

This is only 1 successful test by 1 user.  More testing, including by other users, is recommended prior to merging this PR.

I intend to reset and perform another test when time permits.

chewitt:
@DeltaMikeCharlie thanks for the feedback. Hopefully a few others can test and give more thumbs-up 👍🏻

userxy79:
@chewitt Thank you for your fast respond ;) What does Claude have to say to Copilots Output?

@DeltaMikeCharlie Thank you for testing, my second Testsystem with seperate TV-Card doesnt exist anymore, i only have my 24/7 Livesystem wich cannot be used for testing. So thank you ;)

DeltaMikeCharlie (approved):
I have successfully completed a second test run of two systems in parallel.  I used a different mux on the second test.  The 'DTS Discontinuity' error occurred on my reference system and did not occur on my patched system.  The patched system also time-shifted continuously for over 28 hours without error or any unexpected log messages.

All I can really say is that 'this fix works for me'.

We have been in a similar situation with #1931 where it worked on my test setup but other users encountered errors.  Perhaps my testing methodology is flawed, maybe some subtle configuration option that I have overlooked is having an unforeseen influence.

Personally, I would prefer to see more testing on a broader set of environments before this PR is merged.  However, lack of testers may make this impractical.  If this PR does not work for other users, perhaps it can be reversed, like with #1931/#2049.

userxy79:
@DeltaMikeCharlie Thank you again for Testing. The old https://github.com/tvheadend/tvheadend/pull/1931 had a complete different approach, Flole998 approach just resetet the DTS Counter to Zero when the Wraparound happend wich leaded to Stream Corruption, this https://github.com/tvheadend/tvheadend/pull/2248 approach trys to solve the Problem in an different way. 

Anyway thank you for Testing again, but as it happend before, or since i posted the Bug, none else will Test this any further.

Maybe @chewitt can let Copilots Answer https://github.com/tvheadend/tvheadend/pull/2248#pullrequestreview-5187494239 run trough Claude Code again to think over it and if it really is that Problem, wich i dont think it is but we never know, we can make this Fix live and if it throws Problems it gets reverted as you mentioned and we have to look over it again.

Flole998:
My approach was simply an older AI model at work ;) Yeah lets have claude think over the review of Copilot and based on what it thinks we can merge this then. The options to do it "properly" and 100% correct are kinda limited here

chewitt:

@userxy79 @DeltaMikeCharlie — I went back over Copilot's review. **The finding is correct**, so I've pushed a change for it (branch also rebased on current master).

### What Copilot found

The patch swaps `pts_diff(ref, dts)` for a modular offset `(dts - ref) & PTS_MASK`, and the first-packet test from `dts < 0` to `dts > PTS_MASK / 2`. Working through the cases, the new and old code agree everywhere **except** one band — offsets between `PTS_MASK/2` and `PTS_MASK`, i.e. 13.26h to 26.5h after the reference:

| case | old | new (before this push) |
|---|---|---|
| 1s before ref (early packet) | drop | drop |
| ref near top, DTS just wrapped | keep | keep |
| ref 0, DTS 2^32 (13.26h late) | keep | **drop** |

That band is not hypothetical. `tsfix_update_ref()` only substitutes a stream's own DTS for `tf_tsref` on the audio (`>3s`) and teletext (`>2s`) paths. A DVB subtitle packet within 6s of the audio's last DTS gets `tfs_local_ref = tf->tf_tsref` — the *session* reference — and video has no branch at all. So a subtitle track that stays silent for 13+ hours, or a component added by a PMT change, has a first-packet offset equal to the elapsed session time.

Worse, `tsfix_packet_drop()` doesn't set `tfs_last_dts_norm`, so the stream stays in the first-packet branch and keeps dropping until the offset passes 26.5h — roughly a 13 hour outage on that track.

### The change

I didn't take Copilot's suggested remedy of restoring the `pts_diff()` ordering check, because the old behaviour in that band was inconsistent: it depended on the absolute values rather than the modular relationship (`ref=0, dts=2^32` was accepted, and so was `ref=5e9, dts=1e9`, but via the narrow wrap heuristic). Reinstating it brings back what this PR set out to remove.

Instead the cut point moves to just below the reference, because the two sides aren't symmetric. Only the backlog can legitimately precede the reference, and `tsfix_input()` pulls it back by at most 160000 ticks (~1.78s) to cover that. Everything above is fair game for a late-starting stream:

```c
    if (dts > PTS_MASK - 2 * 90000 || pkt->pkt_err) {
```

Every drop the patch already made for genuinely early packets is preserved, and the 13.26h–26.5h band is handed back. The trade-off is that a packet more than two seconds *before* the reference is now read as a very late one rather than dropped — that is out-of-spec input either way, and the discontinuity logic handles what follows.

This is a corner case that neither test run would have hit, since @DeltaMikeCharlie's 27h and 28h sessions had all components live from the start. It doesn't change the behaviour those runs validated.

PR #2249, opened 2026-09-11
timeshift: record from the cache of a tuned channel

th0ma7:
### What

A recording started after its programme began — Kodi's record button on what is playing, typically — started at the moment it was requested, even when the channel had been watched for hours with timeshift on. It now starts at the programme's start, as far back as the channel's cache reaches. Kodi's timeshift is served from that same cache.

New timeshift option **Record from cache** (off by default, needs timeshift enabled). Three commits:

1. **Record from cache.** Each tuned channel keeps its last *Maximum period* of MPEG-TS, as the service hands it to its subscribers, before any per-subscription parsing. A recording whose start lies in the past replays it from `start − pre-padding`, then joins live. Appending to the cache and handing over both happen under `s_stream_mutex`, so the join has neither a gap nor a duplicate. Every profile works: packet profiles parse the replayed TS as they parse live TS.
2. **HTSP clients timeshift through the channel cache**, instead of each writing a buffer of its own: one cache per channel, however many clients and recordings. Same speed and skip requests and replies as the packet timeshift. It can be dropped without touching the first commit.
3. **A fix to the packet timeshift**, independent of the other two (below).

### Testing

Local bench: a real-time IPTV `pipe://` ffmpeg source, HTTP and HTSP clients, recordings created through the API.

- **Recording started 30 s late**, `pass` and Matroska: 49.8 s and 48.4 s instead of 19 s, video and audio continuous, 0 CC errors at the join. Also covered: capped by *Maximum period*, *RAM only*, 20 Mbit/s over four disk segments, two recordings at once, a recording stopped mid-replay.
- **Scripted HTSP client**: pause, resume, relative and absolute skip, 4×, −2× and back to live, compared with the packet timeshift on the same run. Skip replies and landing keyframes are identical, and DTS are continuous across pause/resume and back to live. Also covered: two clients on one channel (sharer), a pause longer than the cache holds, a disconnect while rewinding.
- Not yet run on DVB/ATSC hardware.

### The packet timeshift fix

On current master the packet timeshift aborts tvheadend when an HTSP client rewinds, then goes back to live. In `timeshift_reader()`, `SMT_SKIP_LIVE` frees the packet waiting to be delivered but keeps the pointer; the same message is then delivered, and freed a second time, further down the loop — `streaming_msg_free()` ends in `abort()` on the stale message type. Reproduced on `2e64ec02c` with a scripted client (speed −200, then `subscriptionLive`); with the fix the same scenario goes back to live:

```diff
                 /* Release */
                 if (sm)
                   streaming_msg_free(sm);
+                sm = NULL;
```

The packet timeshift stays in use whenever the new option is off — the default — and, with it on, for transcoding profiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -485,7 +485,8 @@ SRCS-TIMESHIFT = \
 	src/timeshift/timeshift_filemgr.c \
 	src/timeshift/timeshift_writer.c \
 	src/timeshift/timeshift_reader.c \
-	src/timeshift/timeshift_svcbuf.c
+	src/timeshift/timeshift_svcbuf.c \
+	src/timeshift/timeshift_svcts.c
 SRCS-${CONFIG_TIMESHIFT} += $(SRCS-TIMESHIFT)
 I18N-C += $(SRCS-TIMESHIFT)
 
--- a/Makefile
+++ b/Makefile
@@ -484,7 +484,8 @@ SRCS-TIMESHIFT = \
 	src/timeshift.c \
 	src/timeshift/timeshift_filemgr.c \
 	src/timeshift/timeshift_writer.c \
-	src/timeshift/timeshift_reader.c
+	src/timeshift/timeshift_reader.c \
+	src/timeshift/timeshift_svcbuf.c
 SRCS-${CONFIG_TIMESHIFT} += $(SRCS-TIMESHIFT)
 I18N-C += $(SRCS-TIMESHIFT)
 
--- a/docs/class/timeshift.md
+++ b/docs/class/timeshift.md
@@ -21,9 +21,13 @@ the moment it was requested. It works with every stream profile.
 
 The cache holds the channel's stream as received, in the *Storage path*
 (or in RAM with *RAM only*, bounded by *Maximum RAM size*), and is removed
-once nothing watches or records the channel any more. It is separate from
-a client's own timeshift buffer, so a Kodi client pausing live TV keeps
-using its own as before.
+once nothing watches or records the channel any more.
+
+HTSP clients (Kodi) pausing, rewinding or skipping through live TV then use
+that same cache instead of writing a buffer of their own: one cache per
+channel, however many clients and recordings share it. A client paused
+for longer than the cache holds goes on, when it resumes, from a little
+after the oldest data still there.
 
 ---
 
--- a/docs/class/timeshift.md
+++ b/docs/class/timeshift.md
@@ -10,6 +10,23 @@ This tab is used to configure timeshift properties.
 
 ---
 
+## Recording from cache
+
+With *Record from cache* enabled, every tuned channel keeps its own
+cache of the last *Maximum period*, from the moment it was tuned. A
+recording started after its programme began -- pressing record on what is
+playing, typically -- then begins at the programme's start (less the
+pre-recording padding), or as far back as the cache reaches, instead of
+the moment it was requested. It works with every stream profile.
+
+The cache holds the channel's stream as received, in the *Storage path*
+(or in RAM with *RAM only*, bounded by *Maximum RAM size*), and is removed
+once nothing watches or records the channel any more. It is separate from
+a client's own timeshift buffer, so a Kodi client pausing live TV keeps
+using its own as before.
+
+---
+
 ## Buttons
 
 <tvh_include>inc/buttons</tvh_include>
--- a/src/dvr/dvr_rec.c
+++ b/src/dvr/dvr_rec.c
@@ -177,6 +177,13 @@ dvr_rec_subscribe(dvr_entry_t *de)
     goto _return;
   }
 
+  /* Started after the programme began (Kodi's record button, typically):
+   * take what the channel's cache still holds, if it is cached -- for
+   * the first file only, a restarted recording already has that part.
+   * The subscription links later, from the reschedule timer. */
+  if (de->de_files == NULL && dvr_entry_get_start_time(de, 0) + 2 < gclk())
+    de->de_s->ths_backfill_from = dvr_entry_get_start_time(de, 0);
+
   de->de_chain = prch;
 
   atomic_set(&de->de_thread_shutdown, 0);
@@ -1370,6 +1377,7 @@ dvr_rec_start(dvr_entry_t *de, const streaming_start_t *ss)
   htsmsg_field_t *f;
   muxer_t *muxer;
   struct stat st;
+  time_t start;
   int i;
 
   if (!cfg) {
@@ -1526,11 +1534,19 @@ dvr_rec_start(dvr_entry_t *de, const streaming_start_t *ss)
 
   streaming_start_unref(ss_copy);
 
-  /* update the info field for a filename */
+  /* update the info field for a filename; backfilled from the channel
+   * cache, the file starts with what the cache held (once: a later file
+   * of the same recording starts when it does) */
   if ((f = htsmsg_field_last(de->de_files)) != NULL &&
       (e = htsmsg_field_get_map(f)) != NULL) {
     htsmsg_set_msg(e, "info", info);
-    htsmsg_set_s64(e, "start", gclk());
+    start = gclk();
+    if (de->de_s && de->de_s->ths_backfill_start &&
+        de->de_s->ths_backfill_start < start)
+      start = de->de_s->ths_backfill_start;
+    if (de->de_s)
+      de->de_s->ths_backfill_start = 0;
+    htsmsg_set_s64(e, "start", start);
   } else {
     htsmsg_destroy(info);
   }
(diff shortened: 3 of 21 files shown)


th0ma7:
@Flole998 this is something I always whished tvh could manage:  being able to reccord the entire movie I've been watching since the past hour from the beginning instead of it starting from now and discarding what's in cache from timeshift.

Well this PR adds exactly that with a "Reccord from Cache" option.

Would it be possible to allow this PR getting a sonarcube and copilot review so i can pursue testing?  Much thnx in advance.

th0ma7:
SonarCloud code review addressed.  A copilot review would be nice to have as well if feasible.  Thnx!

Flole998:
Ooops, out of quota again. But hey, progress, lasted almost 2 weeks this time (as I switched the review to "Lite")

LiBe-net:
This looks great — I'm also very interested in this feature.

I do have one concern regarding PID changes: with Record from cache enabled, HTSP timeshift uses the new channel-cache path instead of the current timeshift implementation. Could this cause problems when rewinding or seeking across a PID change?

PR #2206, which has not been merged yet, fix PID changes for live TV, recordings and timeshift. I'm wondering whether these two PRs are fully compatible, especially when the cached section spans a PID change.

th0ma7:
> This looks great — I'm also very interested in this feature.

@LiBe-net Glad to hear!  Seems I'm not the only one missing this feature.

> I do have one concern regarding PID changes: with Record from cache enabled, HTSP timeshift uses the new channel-cache path instead of the current timeshift implementation. Could this cause problems when rewinding or seeking across a PID change?

Good question -- And frankly this is something I don't have the expertise to answer.  Here is claude`s answer after passing almost the entirety of my daily credits... hopefully this will be more meaningful to you than it is to me?

Built a bench for it: a source whose elementary PIDs switch mid-stream (0x100/0x101 -> 0x200/0x201) on the same service id and PMT PID, with the PSI version bumped so tvheadend actually sees the change.

**Rewinding across the change, HTSP.** At the PID change tvheadend stops and restarts the subscription (`subscriptionStop {status: "Source reconfigured"}`, then `subscriptionStart`), so the client's time base restarts and *neither* path can rewind before the change -- that is not specific to this PR. What differs is what happens when a client tries: with *Record from cache* on, the seek lands at the start of the new timeline and playback continues (425 video packets after resume, no hole); with it off, on the current timeshift, the same seek returned no video packets at all. So for seeking, the cache path is no worse, and in this test better.

**Backfilling a recording across the change -- this one is a real defect, and thanks for pointing at the area.** The replay feeds raw TS back through the subscription's parser, which maps packets to streams *by current PID*
(`parser_find_es()`), and the stream table comes from the `SMT_START` of the moment the recording starts. So data cached before the change no longer matches:
 - pass profile: the old-PID packets are written, but the first PMT the muxer   emits announces no PID at all, so the backfilled head is unplayable. A live recording running across the change announces `0x100,0x101` and then
  `0x200,0x201` correctly.
 - matroska profile: the pre-change part is simply missing -- 39 s recorded out of 60.

**Suggested approach**: mark in the cache where the stream layout changes, and never replay across such a mark -- the backfill would start at the last change instead, and a client's oldest position would be clamped the same way. A recording then begins where it can be decoded, rather than with a head no player can read. The cost is that a recording started late, just after a PID change, gets little or no catch-up; the alternative -- replaying the old part with its old stream table -- means carrying the layout in the cache and re-announcing it downstream, which is a good deal more machinery for a case that lasts seconds.

> PR #2206, which has not been merged yet, fix PID changes for live TV, recordings and timeshift. I'm wondering whether these two PRs are fully compatible, especially when the cached section spans a PID change.

Regarding our two patches, again here's claude analysis: 
Your PR and this one merge cleanly (three files touched by both, in different areas), and your pass-muxer PID mapping would apply to the replayed part too once the replay is clamped.

Hope this answer doesn't muddy the water?  Let me know if I suggest approach makes sense to you in which case I'll look into.

Lastly, I've been using this patch since a little while already and its been working great: not that the functionality is something I need that often, but its super convenient to have ready when time comes.

LiBe-net:
This shared cache idea is really great. I've been experimenting with it a bit, and it opens the door to some interesting additional features.

For example, we could keep a subscription open for a while as long as tuners are available. Another possibility would be to allow recording a previous program from the cache alongside the currently airing program.

th0ma7:
> This shared cache idea is really great. I've been experimenting with it a bit, and it opens the door to some interesting additional features.

I'd like to say it was my idea but the AI did it...

> For example, we could keep a subscription open for a while as long as tuners are available. Another possibility would be to allow recording a previous program from the cache alongside the currently airing program.

I thought of the later.  Two things:  
1. if we keep caching it would block the system to go to sleep as always hitting the disks.  Although one could keep the caching active on a channel upon disconnection of a client until it hits its cache max, then stop allowing the system to go to sleep.  I wonder what other maintainers would think of this idea, and wetter it would deserve a on/off switch or being the default behavior.
2. Secondly, I'm not sure kodi (or other clients) allows going back in the (recent) past to hit reccord after the fact of an ended program.  I presume this issue is more complex than it looks like.

Lastly, any opinion on the proposed change in my previous comment, which i believe would also align with that other PR's behavior?

LiBe-net:
Hi,

my first priority is stability, so #2206 is a must-have for me, especially because of the PID/stream mapping fixes.

I tried combining this PR with #2206, using AI assistance for the merge, and added two additional improvements on top:

Retro recording: allow recording past programs as long as the programs is still available in the channel cache.
Keep tuned channels (mins): keep a channel subscription alive for a configurable amount of time (currently testing with 15 minutes), as long as a tuner is available. This keeps the channel cache alive when switching away from a channel.

The first tests look very promising so far. When zapping between channels, timeshift continues to work when switching back to a previously watched channel, and I can also start a recording for a program that is still available in the cache.

What I have not tested thoroughly yet is PID changes. That is probably the most important remaining part: I want to verify that PID/stream changes still work correctly in all relevant scenarios:

live TV
normal timeshift
regular recordings
recording from cache
retro recording

If all of these continue to work reliably together, this combination would cover pretty much the behavior I am looking for.

th0ma7:
Super interesting! Wouldn't you need a maximum cache hit for the case where someone (my kid) start zapping throughout the guide? Or instead if a zapped channel didn't lasted for enough time (e.g. the user changed to yey another channel within 1 min) then it doesn't timshift that program any longer?

LiBe-net:
Zapping would not be a problem, since unused subscriptions would be closed depending on their weight and adapter availability. If necessary, this could even happen before the 15-minute mark, as active subscriptions would have a higher weight than inactive ones.

That said, the likelihood of retaining an older timeshift would be lower. Having enough adapters available would help mitigate this, though.

Recording old programs from timeshift cache is more challenging, especially when other scheduled recordings are running parallel: I haven't managed to get that working reliably yet.

Flole998:
If you guys want me to test something on my 32 channel DVB-C tuner, that effectively receives all DVB-C muxes available simultaneously, just let me know :)

th0ma7:
> If you guys want me to test something on my 32 channel DVB-C tuner, that effectively receives all DVB-C muxes available simultaneously, just let me know :)

Wow that's serious gear :). Feel free to test this first patchset demonstrating the basic feature.

@libe-net would your work apply on top of this PR or some changes would be preferably needed here as well?  As otherwise this PR is fairly contained on its own.

LiBe-net:
For my testing, I had to make some adjustments to parts of this PR, as combining it with #2206 currently seems to break some of the fixes needed for stable PID changes.

So personally, I think we should get #2206 in first and then build the cool shared-timeshift features from this PR on top of it. That way, we can make sure the PID change handling stays intact while adding the new timeshift functionality.

@Flole998: A 32-channel DVB-C tuner is pretty crazy 🙂 Zapping around could potentially become a performance topic as well, since that many parallel timeshift buffers might start hitting other limits. It may be that we need some kind of limit here as well.

th0ma7:
> For my testing, I had to make some adjustments to parts of this PR, as combining it with #2206 currently seems to break some of the fixes needed for stable PID changes.
> 
> So personally, I think we should get #2206 in first and then build the cool shared-timeshift features from this PR on top of it. That way, we can make sure the PID change handling stays intact while adding the new timeshift functionality.

Totally good with that, feel free to add to this PR or provide me with patches i could apply on top as needed, unless there is a atrict dependency between the two.

> @Flole998: A 32-channel DVB-C tuner is pretty crazy 🙂 Zapping around could potentially become a performance topic as well, since that many parallel timeshift buffers might start hitting other limits. It may be that we need some kind of limit here as well.

Exactly the type of corner case i had in mind in my previous post!  I only have 6 tuners at home and thought this was already a lot of parallel reccording that could strain an energy efficient focused nas.  But Flole998's gear is next level...

PR #2251, opened 2026-09-14
Add export of XMLTV date element for program element

1stsetup:
On import program date element is parsed and internally saved as `ebc->copyright_year`. But on accessing the api endpoint `/xmltv/channels` it is not exported. This change add the date element to the exported programs.


src/webui/xmltv.c
```diff
@@ -215,6 +215,13 @@ http_xmltv_programme_one_long(const http_connection_t *hc,
       htsbuf_append_and_escape_xml(hq, ebc->image);
       htsbuf_append_str(hq, "\"/>\n");
   }
+  /* The XMLTV <date> element denotes the original production/release date.
```
Flole998: This comment is unnecessary, the code is not explanation-worthy as its effectively the same as above

1stsetup: I removed the comment.

chewitt: Please squash the changes to a single commit.

chewitt:
The git mistake you are probably making is working in the master branch instead of a topic branch that can be rebased against master. This might (or might not) help: https://wiki.libreelec.tv/development/git-tutorial

There is almost never a good reason to close and open a new PR; the original can always be fixed. If you have the two original commit hashes in history anywhere:

```
git checkout master
git checkout -b xmltv-date
git remote add upstream https://github.com/tvheadend/tvheadend.git
git fetch upstream master
git reset --hard upstream/master
git cherry-pick <hash-of-commit-1>
git cherry-pick <hash-of-commit-2>
git rebase -i HEAD~2
<fixup or squash commits into one>
git push origin xmltv-date
<edit this PR to change the source branch in your repo to xmltv-date>
```

NB: If you ever need git guidance; ask. Everyone remembers their first fumbling steps with git and most people will be happy to show you the way :)

Flole998:
Uhmm, it's actually a lot easier: If you have the branch already locally all you need to do is
`git rebase -i HEAD~2` and then keep the first commit and squash (or maybe fixup in this case) the second one into it, force-push, done (normal push like shown above won't work).

Or just soft-reset and create a new commit with both changes in the staging area.

AI is actually quite good with all of this, if you tell it how you messed up and what you want to achieve it usually gives you the right commands.

1stsetup:
I did that but the code got all newly formatted by vscode. Instead of three new lines there were multiple changes. This was easier.

PR #2253, opened 2026-09-17
linuxdvb/ddci: add multi-transponder CAM decryption support

VizDock:
## Summary

Implement multi-transponder decryption (MTD) for LinuxDVB DDCI devices, allowing a multi-service CAM to decrypt services from multiple DVB muxes concurrently through a single physical CI interface.

The implementation uses per-mux virtual transport-stream namespaces for CAM-facing PID and service identifiers while retaining one physical DDCI transport. Services from the same source mux share an MTD context.

## Problem

DDCI provides a single physical transport path to the CAM. When services from independent muxes are sent through that path, their PID and service-ID namespaces may overlap, and their PSI/SI and conditional-access signalling cannot simply be combined unchanged.

This prevents reliable multi-mux operation even when the CAM itself supports multiple simultaneous programs.

MTD provides an isolated virtual namespace for each active mux while multiplexing those namespaces through the physical DDCI interface.

## Implementation

The implementation includes:

- **Per-mux MTD contexts.** Services belonging to the same source mux share a context rather than consuming independent virtual namespaces.

- **PID virtualization.** CAM-visible PIDs are allocated within the context namespace while Tvheadend continues to use the original broadcast PIDs for service/component and CA-descriptor matching.

- **Service-ID virtualization.** CA-PMT program numbers are mapped independently from PID mappings.

- **CA-PMT mapping during serialization.** ES PIDs, CA descriptor PIDs and service IDs are translated while the CA-PMT is built, so ADD/UPDATE/DELETE operations use the same virtual namespace.

- **Mux-wide CAM PID ownership.** CAM-facing PID ownership is maintained across all active services sharing a mux context.

- **CAM return routing.** Packets returned by the CAM are resolved through the MTD namespace and translated back to their original PID before delivery to the appropriate mux.

- **CAT/EMM handling.** Each active context parses its source CAT and maps the required EMM PIDs. A single synthesized CAT is sent on the physical CAM transport containing the combined active CA descriptors, while EMM traffic from all active contexts can be forwarded.

- **PSI/SI isolation.** Source PSI/SI that would expose conflicting original mux identifiers is not forwarded directly into the combined CAM transport.

- **Initial EIT injection.** A small mapped present/following EIT is generated during service startup for CAMs that require service information during program activation.

- **Context-aware DDCI scheduling.** Native traffic and per-context queues share the physical CAM input without allowing one virtual context to monopolize it.

- **MTD namespace lifecycle management.** Released namespaces are quarantined briefly before reuse, and their PID allocator state is advanced across reuse to avoid immediately recycling the same CAM-visible identities.

- **DDCI EMM demux integration.** CAT-discovered EMM PIDs can reach the DDCI path even though they are not normal elementary-stream components.

- **Service-limit handling and DDCI lifecycle cleanup** required for repeated context creation/destruction and multi-service operation.

## Limits and configuration

The implementation provides up to **31 simultaneous MTD mux namespaces**.

This is a mux-context limit, not a 31-channel limit. Multiple services from the same DVB mux share one MTD namespace.

The number of services that may be assigned to the CAM is additionally controlled by Tvheadend's configured CAM **Service limit**. This should be configured according to the capabilities of the installed CAM.

The practical number of simultaneously decrypted services is hardware-dependent. CAMs and CI interfaces may impose independent limits on:

- simultaneous programs;
- simultaneous CA/ECM/EMM and elementary-stream PIDs;
- aggregate transport-stream throughput;
- CI interface throughput.

The 31-context namespace capacity therefore does not imply that a particular CAM can decrypt 31 services simultaneously.

Physical tuner availability is independent of the MTD namespace limit. Starting a service on an additional mux still requires an available input capable of receiving that mux.

## Testing

Testing was performed with a professional multi-service CI CAM intended for concurrent service decryption.

The final implementation was tested with:

- **7 simultaneous encrypted services**
- **4 independent DVB muxes because testing tuner was Digital Devices Max M4 with 4 tuners**
- multiple services sharing one mux context
- repeated service stop/start
- complete MTD context destruction and recreation
- reuse of released MTD namespaces
- mixed HTSP and HTTP clients
- physical tuner exhaustion followed by returning to an already active mux

A same-mux multi-service control test was also performed without MTD to verify that the CAM/CI hardware could decrypt multiple concurrent programs independently of the MTD implementation.

During the final fresh-build test, all seven encrypted services operated concurrently across four muxes. Services were repeatedly removed and re-added, and MTD contexts were destroyed and recreated without progressive degradation.

A request requiring an additional physical tuner correctly failed when all tuners were occupied. Returning immediately to a service on an already active mux succeeded without disturbing the existing MTD contexts.

No failed decrypt activation requiring a retry and no progressive degradation were observed during the final test.

## Build validation

The final change was applied to a clean Tvheadend tree and validated with:

- successful `git apply --check`
- successful build
- clean `git diff --check`
- runtime multi-mux/multi-service testing described above

The change is limited to the existing DVB-CAM, EN50221, LinuxDVB/DDCI and MPEG-TS demux paths.


VizDock:
@Flole998 Since you previously showed interest in multi-transponder DDCI support, I'd appreciate it if you could take a look at this PR when you have time. I hope I'm not pushy with this. The implementation has been tested with multiple encrypted services across four independent muxes, and the current Sonar Quality Gate passes.

PR #2257, opened 2026-09-22
container: build and bundle the Vue UI in the official images

chewitt:
Forum users report that the official Docker images do not have the new web GUI — `/gui` serves the **"Vue UI not available in this build"** stub instead of the Vue admin UI.

## Problem

`ENABLE_VUE_UI` is `(ENABLE_VUE_BUILD || ENABLE_VUE_CACHE)`. Neither `Containerfile.alpine` nor `Containerfile.debian` installs node/npm, and neither passes `--enable-vue_cache`, so the `vue_build:auto` probe finds no toolchain, prints a non-fatal warning, and `./configure` carries on with both options off. `webui.c` then registers the stub from `vue.c` in place of the real `/gui` routes and the image builds and publishes clean, which is why this went unnoticed.

The container images are the only packaging path that missed this plumbing — `build-ci.yml` and `build-cloudsmith.yml` already gate on the shared `webui-vue-build.yml` workflow and configure with `--enable-vue_cache --disable-vue_build`.

## Solution

Give `container-build.yaml` the same treatment: a `vue` job calling `./.github/workflows/webui-vue-build.yml`, a `needs: vue` gate, and a download of its dist artifact into `.vuecache` in the build context. The Containerfiles point `VUE_FILE_CACHE` at `/src/.vuecache` and configure with `--enable-vue_cache --disable-vue_build`, so `libcacheget` unpacks the staged tarball and the images bundle the exact same UI build as the deb and rpm packages. No Node toolchain enters the image, and there is only one workflow to keep current.

Builds outside CI — a plain `docker build` — find no artifact and fall back to the pcloud fetch, which is why Alpine gains `py3-requests` (Debian already had `python3-requests`).

Two supporting changes:

- `.dockerignore` gains `**/node_modules` and a stale `src/webui/static-vue/dist`, so a developer's checkout does not push hundreds of megabytes of host npm state through `COPY . /src/`.
- An explicit `--enable-vue_build` with no toolchain present now fails instead of quietly disabling itself, and the `auto` fallback says which UI the resulting binary will serve. That is the guard that would have made this visible when it was introduced.


Flole998:
Dont install node, use the pre-built UI from the other github actions step. That ensures all builds use the same webinterface and we only have to worry about a single workflow having the right versions and so on.

chewitt:

Good call — reworked to do exactly that. The Node stage is gone; `container-build.yaml` now gates on `webui-vue-build.yml` like `build-ci` and `build-cloudsmith`, downloads the dist artifact to `.vuecache`, and the Containerfiles consume it via `VUE_FILE_CACHE` with `--enable-vue_cache --disable-vue_build`. Same artifact in the images as in the deb/rpm packages, and one workflow to keep current.

The only addition is `py3-requests` on Alpine, so a plain `docker build` outside CI still resolves the UI through the pcloud fallback rather than silently shipping the stub.

PR #2258, opened 2026-09-22
dvr: fix use-after-free when purging obsolete autorec entries

MikrotikExe:
dvr_autorec_purge_obsolete_timers() iterated over dvrentries with a plain LIST_FOREACH while calling dvr_entry_destroy() on matching entries. dvr_entry_destroy() unlinks the entry and drops its last reference, so the next iteration read the link pointer from freed memory.

The bug is normally hidden because the freed block usually still holds the old pointer. Running with MALLOC_CONF=junk:true makes it fail immediately:

```
#0  dvr_autorec_entry_can_be_purged (de=0x5a5a5a5a5a5a5a5a) at src/dvr/dvr_autorec.c:106
106       if (!de->de_autorec)
#1  dvr_autorec_purge_obsolete_timers () at src/dvr/dvr_autorec.c:132
        de = 0x5a5a5a5a5a5a5a5a
        num_purged = 1
#2  dvr_init () at src/dvr/dvr_config.c:1633
#3  main () at src/main.c:1343
```

num_purged is 1: the first matching entry had just been destroyed, and the loop then dereferenced it.

Fetch the next entry before the current one can be destroyed. The same list is already iterated safely elsewhere in dvr_db.c (dvr_done).

Tested on two 4.3 installations with around fifty autorec rules. Before the patch, startup under junk:true faulted at the first purge in three runs out of three. After it, the same startup purges four entries and continues normally. Without junk:true neither build faults, which is why this has gone unnoticed.


Flole998:
Nice one!

PR #2259, opened 2026-09-22
webui: drop a stale comet backlog instead of accumulating it

MikrotikExe:
A comet mailbox grows without bound while a web UI page sits in a backgrounded browser tab. The ExtJS interface holds the websocket open, so `comet_flush()` never releases the mailbox (`cmb_refcount` stays above 1), and the queued messages are only ever handed over in `comet_message()`, which a tab with frozen timers never reaches. Meanwhile `subscription_status_callback()` and `mpegts_input_status_timer()` each queue one message per subscription and per input every second.

Measured on a 4.3 server with the drop instrumented:

```
comet: dropped 3702 messages queued for 60 seconds without being collected
```

That is 62 messages a second of accumulated htsmsg objects. systemd reported a peak RSS of 18.2G on that unit before the change; with it the resident size stays flat.

Nothing older than a minute is of any use to the UI: it wants the current state, not a replay. This drops what has piled up once the client has not collected anything for `MAILBOX_STALE_TIMEOUT` seconds, and reports it through `tvhlog_limit()`.

`cmb_last_taken` is reset only in `comet_message()`, which is the single place a mailbox is handed to a client, so an active client — long polling or websocket — never trips the timeout.

Separate from #2241 and #2258.


Flole998:
I assume you are using AI for assistance? Could you please ask it what cmb_last_used is for and how it differs from the newly introduced functionality? It might be possible that it was intended for something like this, but due to some bug its not working properly.

MikrotikExe:
> I assume you are using AI for assistance? Could you please ask it what cmb_last_used is for and how it differs from the newly introduced functionality? It might be possible that it was intended for something like this, but due to some bug its not working properly.

Yes, I use an AI assistant for the analysis. Everything below I checked against the code myself.

`cmb_last_used` governs the lifetime of the mailbox object, not the age of its contents, and the two cannot stand in for each other.

It is deliberately cleared to 0 while a client holds the mailbox — in `comet_find_mailbox()` and again in `comet_mailbox_ws_msg()` after every delivery — with the comment "Make sure we're not flushed out". It is set back to `mclk()` when the websocket handler exits and in `comet_message()`. `comet_flush()` then destroys a mailbox only when `cmb_refcount == 1 && cmb_last_used && cmb_last_used + sec2mono(60) < mclk()`.

With a websocket attached, `cmb_refcount` is 2 and `cmb_last_used` is 0, so `comet_flush()` correctly leaves that mailbox alone: the client is still there. What goes stale is the queued content of a mailbox that is legitimately alive, and `cmb_last_used` does not describe that at all.

One thing near it does look broken: `MAILBOX_UNUSED_TIMEOUT` (20) is defined and never used anywhere in the file — `comet_flush()` hardcodes `sec2mono(60)` instead.

And a correction to my own description above, which your question led me to. I wrote that `comet_mailbox_ws_msg()` only delivers when the client sends something first. That is wrong: `http_websocket_read()` returns 0 on `ETIMEDOUT`, so the loop in `comet_websocket()` calls `comet_mailbox_ws_msg()` roughly once a second whether or not the client sent anything, and that would drain the mailbox.

So a queue that grew for 60 seconds means the loop did not reach `comet_message()` at all in that time — `cmb_last_taken` is set only there, and only when messages were actually present. I have not established what stalls it. The candidates I can see are the blocking `http_websocket_send_json()` of the previous iteration, whose return value is discarded, and `tcp_read_timeout()` not returning within its timeout. I would rather say that than guess.

The measurement itself stands: 3702 messages queued for 60 seconds without being collected, repeatedly, on a server with a UI page sitting in a backgrounded tab.

So this patch bounds the symptom and does not explain the stall. If you would rather have the stall found first, I am happy to instrument it and come back instead of having this merged as it is.

Flole998:
Awesome, just wanted to double-check we are not adding duplicate code to fix something that should already be working in the existing code.

